PR #7, opened 2022-11-05
+midnight-commander.org

jhheider:
Adds midnight-commander (`mc`, et al.)

### DIFF
--- a/projects/midnight-commander.org/package.yml
+++ b/projects/midnight-commander.org/package.yml
@@ -1,0 +1,33 @@
+distributable:
+  url: http://ftp.midnight-commander.org/mc-{{ version }}.tar.xz
+  strip-components: 1
+
+versions:
+  github: MidnightCommander/mc/tags
+
+provides:
+  - bin/mc
+  - bin/mcdiff
+  - bin/mcedit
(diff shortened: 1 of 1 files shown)


projects/midnight-commander.org/package.yml
```diff
@@ -0,0 +1,42 @@
...
+  dependencies:
+    tea.xyz/gx/cc: '*'
+    tea.xyz/gx/make: '*'
+    gnu.org/autoconf: '>=2.64'
+    gnu.org/automake: '>=1.12'
+    gnu.org/libtool: '*'
+    freedesktop.org/pkg-config: '^0.29'
+  script: |
+    # Needed for the compiler to `#include <term.h>`
+    export CPATH=$CPATH:{{deps.invisible-island.net/ncurses.prefix}}/include/ncursesw
```
mxcl: could be in `env` YAML node. You can put $CPATH in there and it will be assigned after ours.

mxcl (approved):

projects/midnight-commander.org/package.yml
jhheider: Yum. Better.

PR #13, opened 2022-11-13
+sway

### LINKED ISSUE #12: +forc & +fuel-core
This package will enable users to run Sway smart contracts with the Fuel Toolchain. Sway is DSL based on Rust. I

I'm not sure what is the best way to build these packages, but I'll submit a WIP PR.

Installation instructions for reference:

https://fuellabs.github.io/sway/v0.31.1/introduction/installation.html

pedrocruzio:


projects/crates.io/forc/package.yml
```diff
@@ -0,0 +1,21 @@
...
+
+versions:
+  - 0.31.1
+
+build:
+  dependencies:
+    rust-lang.org/cargo: '*'
+  script:
+    cd sway
+    cargo build
```
jhheider: Options:
```yaml
  script: |
    cd sway
    cargo build
```
or
```yaml
  script:
    cd sway && \
    cargo build
```

jhheider:
Install commands for cargo are the somewhat unintuitive
```sh
cargo install --path=. --root={{prefix}}
```
(so you could probably skip the `cd sway` and do `cargo install --path=sway --root={{prefix}}`)

see [just.systems/package.yml](https://github.com/teaxyz/pantry.extra/blob/main/projects/just.systems/package.yml) for a working example.

jhheider:
(don't commit .DS_Store files to git ;). We've even jumped through hoops in git-scm.org to specifically exclude them globally.)

projects/forc/package.yml
```diff
@@ -0,0 +1,22 @@
...
+    rust-lang.org: ^1.47.0
+    rust-lang.org/cargo: ^0.65
+  script:
+    cargo install --path=sway --root={{prefix}}
+  test:
+    cargo test
+
+test:
+  script: |
+    cargo run --bin forc -- --help
```
jhheider: If it's installed, you should be able to run `forc --help` directly, I assume.

projects/forc/package.yml
```diff
@@ -0,0 +1,22 @@
...
+
+provides:
+  - bin/forc
+
+build:
+  dependencies:
+    rust-lang.org: ^1.47.0
+    rust-lang.org/cargo: ^0.65
+  script:
+    cargo install --path=sway --root={{prefix}}
```
jhheider: are you sure the `cargo.toml` is under `./sway`? usually it's just in `.` (and the build error suggests the former isn't correct).

PR #14, opened 2022-11-14
feat(projects): add htop.dev package

mfts:
I have issues building this locally. 

```
configure: error: in `/Users/mfts/.tea/htop.dev/src/v3.2.1':
configure: error: C compiler cannot create executables
See `config.log' for more details
error: Uncaught (in promise) Error: cmd failed: 77: /Users/mfts/.tea/htop.dev/src/build.sh
```

### DIFF
--- a/projects/htop.dev/package.yml
+++ b/projects/htop.dev/package.yml
@@ -1,11 +1,9 @@
 distributable:
-  # url: https://github.com/htop-dev/htop/archive/refs/tags/{{version}}.tar.gz
   url: https://github.com/htop-dev/htop/releases/download/{{version}}/htop-{{version}}.tar.xz
   strip-components: 1
 
 versions:
   github: htop-dev/htop/tags
-  strip: /v/
 
 dependencies:
   invisible-island.net/ncurses: 6
@@ -21,6 +19,7 @@ build:
     ./autogen.sh && \
     ./configure --prefix={{prefix}} && \
     make
+    make install
   test:
     make test
 
--- a/projects/htop.dev/package.yml
+++ b/projects/htop.dev/package.yml
@@ -19,10 +19,11 @@ build:
     invisible-island.net/ncurses: 6
   script: |
     ./autogen.sh && \
-    ./configure --prefix=/.tea/htop.dev/ && \
+    ./configure --prefix={{prefix}} && \
     make
   test:
     make test
+
 
 test:
   script: |
--- a/projects/htop.dev/package.yml
+++ b/projects/htop.dev/package.yml
@@ -14,9 +14,7 @@ build:
     tea.xyz/gx/make: '*'
     gnu.org/autoconf: '*'
     gnu.org/automake: '*'
-    invisible-island.net/ncurses: 6
   script: |
-    ./autogen.sh && \
     ./configure --prefix={{prefix}} && \
     make
     make install
--- a/projects/htop.dev/package.yml
+++ b/projects/htop.dev/package.yml
@@ -15,7 +15,7 @@ build:
     gnu.org/autoconf: '*'
     gnu.org/automake: '*'
   script: |
-    ./configure --prefix={{prefix}} && \
+    ./configure --prefix={{prefix}}
     make
     make install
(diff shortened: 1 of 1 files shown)


projects/htop.dev/package.yml
```diff
@@ -0,0 +1,29 @@
...
+
+build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+    gnu.org/autoconf: '*'
+    gnu.org/automake: '*'
+    invisible-island.net/ncurses: 6
+  script: |
+    ./autogen.sh && \
```
mxcl: they don’t provide a tarball with `./configure` already generated?

projects/htop.dev/package.yml
```diff
@@ -0,0 +1,29 @@
...
+    ./configure --prefix=/.tea/htop.dev/ && \
```
mxcl: `--prefix={{prefix}}` this is the wrong prefix

projects/htop.dev/package.yml
```diff
@@ -0,0 +1,29 @@
+distributable:
+  # url: https://github.com/htop-dev/htop/archive/refs/tags/{{version}}.tar.gz
+  url: https://github.com/htop-dev/htop/releases/download/{{version}}/htop-{{version}}.tar.xz
+  strip-components: 1
+
+versions:
+  github: htop-dev/htop/tags
+  strip: /v/
```
mxcl: I believe stripping the `v` is automatic, but maybe not.

projects/htop.dev/package.yml
mfts: In their ["compile from source"](https://github.com/htop-dev/htop#compile-from-source) it looks like `./configure` is provided.

projects/htop.dev/package.yml
mfts: What is the correct prefix? Their docs (https://github.com/htop-dev/htop#install) say the `make install` will install it to `/usr/local` by default if I don't change the prefix in configure... I assumed I want it installed in my `.tea` directory.

projects/htop.dev/package.yml
mxcl: Yeah so probs you don’t need to do the `autogen.sh`

projects/htop.dev/package.yml
mxcl: the correct prefix is `{{prefix}}`. the mustaches are expanded to `TEA_PREFIX/PKG.com/v1.2.3`

You can see we do the same in all package.ymls

projects/htop.dev/package.yml
mfts: my bad I misread your question. `configure.ac` is provided so we do need `autogen.sh`

mxcl: There's a `configure`, please remove the autogen and deps:
image

projects/htop.dev/package.yml
```diff
@@ -0,0 +1,29 @@
...
+build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+    gnu.org/autoconf: '*'
+    gnu.org/automake: '*'
+    invisible-island.net/ncurses: 6
+  script: |
+    ./autogen.sh && \
+    ./configure --prefix={{prefix}} && \
```
mxcl: Don’t `&& \`.

We prefer to have distinct invocations of each line for easier to read logs.

projects/htop.dev/package.yml
```diff
@@ -0,0 +1,29 @@
...
+dependencies:
+  invisible-island.net/ncurses: 6
+
+build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+    gnu.org/autoconf: '*'
+    gnu.org/automake: '*'
+    invisible-island.net/ncurses: 6
```
mxcl: no need to specify this in `build` deps since it is a runtime dep

PR #23, opened 2022-11-20
Switched to https

magnusviri:
Same as the commit to pantry.core


mxcl:
oof good catch, sadly Haskell is still not building… I will get to it eventually!

mxcl:
will merge after #2

PR #37, opened 2022-11-30
add boost

### LINKED ISSUE #36: +boost
To be able to package solidity, we need to package boost c++ library first.

https://github.com/boostorg/boost

pedrocruzio:
dependency for solidity


jonchang:
Boost uses git submodules, so you'll either need to do a `git clone --recursive`, or find a tarball that includes a snapshot of everything. I note that the official repository seems to be starting to provide these, but only for a beta release so far: https://github.com/boostorg/boost/releases/tag/boost-1.81.0.beta1 The boost website links to its artifactory repository here which should be the tarballs that you want: https://boostorg.jfrog.io/artifactory/main/release/1.80.0/source/

mxcl:
Needs a real test, cp brew: https://github.com/Homebrew/homebrew-core/blob/master/Formula/boost.rb#L86

PR #41, opened 2022-11-30
Add provides to ffmpeg

magnusviri:


mxcl:
This CI failure isn't your fault, it's an ongoing issue where the build script imports a bunch of stuff because it needs that stuff to decompress tarballs, and that gets into the builds.

ffmpeg probably does want to depend on bzip2 though, so you can just add that.

mxcl:
This will fix the dep injection: https://github.com/teaxyz/pantry.core/pull/50

PR #45, opened 2022-12-01
+yt-dlp

magnusviri:


magnusviri:
This creates the script python-venv.sh, which is the build script from mesonbuild.com & youtube-dl.org moved to a generic file. I don't know how to call it from pantry.core/scripts/brewkit, so I just put the script in the SRCROOT and am calling it from there.

magnusviri:
GHA failed with this message:

/opt/yt-dlp.org/src/build.sh: line 30: /opt/yt-dlp.org/src/v2022.11.11/props/python-venv.sh: Permission denied

I don't know why it couldn't execute, it should be 755. To fix it, I could change it from this

```
"$SRCROOT"/props/python-venv.sh yt-dlp yt-dlp.com
```

to

```
sh "$SRCROOT"/props/python-venv.sh yt-dlp yt-dlp.com
```

mxcl:
If you chmod it and then recommit it’ll fix it.

TO add to pantry-core you PR to pantry-core in the `scripts/brewkit` directory. If you check other package.ymls for how we call brewkit scripts you'll see you just call them. That's because we add the brewkit directory to the PATH.

If you want to PR that then we need to merge that before this will work (due to revlock). If you prefer I'll move the script after.

FYI the script should also write out the shim IMO.

magnusviri:
The test failed because it's revlocked (revision locked?) on https://github.com/teaxyz/pantry.core/pull/54.

mxcl:
Merged https://github.com/teaxyz/pantry.core/pull/54

PR #47, opened 2022-12-01
+catb.org/wumpus

magnusviri:
This isn't the final formula. There's a 1.8 version on GitLab, but it isn't building correctly. I submitted a PR on GitLab to get that fixed. When it's merged then this formula can point to that version.

Also, is there a gitlab version provider?

### DIFF
--- a/projects/catb.org/wumpus/package.yml
+++ b/projects/catb.org/wumpus/package.yml
@@ -1,0 +1,27 @@
+distributable:
+  url: http://www.catb.org/~esr/wumpus/wumpus-1.7.tar.gz
+  strip-components: 1
+
+# When https://gitlab.com/esr/wumpus/-/merge_requests/3 is merged, then switch to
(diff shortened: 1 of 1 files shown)


mxcl:
> Also, is there a gitlab version provider?

not yet, but ofc we want one.

projects/catb.org/wumpus/package.yml
```diff
@@ -0,0 +1,27 @@
...
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+
+  script: |
+    make
+    make prefix={{prefix}} install
+
+test:
+  script: echo no way to test this
```
mxcl, line 27: lol `true` with a comment is what we do, but this is fine.

Though there must be some way, even if it's just execute in bg then kill after a few seconds. But no need to go the extra mile today.

magnusviri: How would you tell if it worked if you execute it in the background? Check to see if the pid exists?

mxcl: you can't kill it if it crashes

PR #53, opened 2022-12-02
docs: add header to README

mfts:
Primarily for consistency


mxcl:
yeah, no. A contents for a tiny README is weird. I think we should switch the banner to Gypsy’s or get @tsmitty to do a revision that says `tea` and nothing else though.

PR #55, opened 2022-12-04
+the_silver_searcher

jhheider:
Add the_silver_searcher (`ag`) grep alternative.


jhheider:
Hm. linux-aarch64 errors. Given that this project hasn't been updated in 4 years and has 400 open issues, maybe I'll just update my workflows instead.

jonchang:
Just use ripgrep! It's already packaged so `tea -X rg ...`.

jhheider:
I should. I just got used to what I was used to. Does `ripgrep` recurse by default? Being able to just `ag foo` in a directory is nice.

jonchang:
Yep, it has a lot of the same features as ag, but not a drop-in replacement.

PR #60, opened 2022-12-05
+rigaux.org/hexedit

jhheider:
I'm not happy about the lack of a test. Any ideas, anyone?

### DIFF
--- a/projects/rigaux.org/hexedit/package.yml
+++ b/projects/rigaux.org/hexedit/package.yml
@@ -10,6 +10,7 @@ build:
     tea.xyz/gx/cc: c99
     tea.xyz/gx/make: '*'
     gnu.org/autoconf: '*'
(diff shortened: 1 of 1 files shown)


projects/rigaux.org/hexedit/package.yml
```diff
@@ -0,0 +1,36 @@
...
+dependencies:
+  invisible-island.net/ncurses: 6
+
+build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+    gnu.org/autoconf: '*'
+  script: |
+    ./autogen.sh
```
mxcl, line 17: do they not provide a tarball with a `./configure`? Generally we prefer that as it is less things that can go wrong.

mxcl:
At some point we could make a test runner that feeds key presses to test these sorts of things.

PR #62, opened 2022-12-06
+lame.sourceforge.io

magnusviri:


magnusviri:
I think I saw this error a few times when I was building it and I don't understand it. If I deleted the built folder and started over the error would go away.

```
downloading
downloading
TLS alert received: AlertMessagePayload {
    level: Fatal,
    description: HandshakeFailure,
}
error: Uncaught (in promise) Error: error sending request for url (https://gigenet.dl.sourceforge.net/project/lame/lame/3.100/lame-3.100.tar.gz#undefined): error trying to connect: received fatal alert: HandshakeFailure
```

mxcl:
We've had issues with source forge URLs before and worked around it by using a mirror that doesn't have these issues.

Switching to curl as the download backend would also help

mxcl:
lol, seg faults on linux x86 but not aarch64

PR #63, opened 2022-12-07
+libjpeg-turbo.org

magnusviri:
This formula is slightly different from the brew version. I'm not completely sure what the differences are because I'm not an expert on the packages I'm building. I'm not sure if I should include links to the brew/apt/yum packages when I submit these. I don't want to give other people more work, but I'm hoping someone else can verify I haven't done something wrong.

### DIFF
--- a/projects/libjpeg-turbo.org/package.yml
+++ b/projects/libjpeg-turbo.org/package.yml
@@ -1,0 +1,36 @@
+distributable:
+  url: https://downloads.sourceforge.net/project/libjpeg-turbo/{{version}}/libjpeg-turbo-{{version}}.tar.gz
+  strip-components: 1
+
+versions:
+  - 2.1.4
+
+build:
+  dependencies:
+    cmake.org: 3
+  script: |
+    cmake -S . -B build -DWITH_JPEG8=1 \
+      -DCMAKE_INSTALL_PREFIX={{prefix}} \
+      -DCMAKE_BUILD_TYPE=Release \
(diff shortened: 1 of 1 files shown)


magnusviri:
I'll be more specific. These are the flags that homebrew has that I didn't include, basically because I don't know what they do, what to replace the strings with, and I'm not sure they're needed.

```
      -DCMAKE_INSTALL_LIBDIR=#{install_libdir}
      -DCMAKE_FIND_FRAMEWORK=#{find_framework}
```

magnusviri:
I just realized that I messed up the test. I'm trying to fix it but I can't figure out how to get the path the src dir. I thought SRCROOT would do it but that points to my "pantry.core" dir. I'm basically trying to find libjpeg-turbo.org/src/v2.1.4/testimages/testimgari.jpg.

jonchang:
> `-DCMAKE_INSTALL_LIBDIR=#{install_libdir}`

This overrides cmake's occasional preference to drop `.so` files in `lib64/` on Linux. Homebrew doesn't support 32-bit so the 32/64 distinction is not preferred.

> `-DCMAKE_FIND_FRAMEWORK=#{find_framework}`

This modifies cmake's order of searching directories on macOS to prefer Unix-style dependencies first, and then only to consider Apple-style Framework directories. This enables Homebrew packagers to use e.g. newer versions of dependencies already included in macOS.

mxcl:
Added in d1784ca25cd6339d30e5b8919e6898a6098ccf1f

PR #66, opened 2022-12-07
rename package after project homepage

jhheider:
@magnusviri: any objection to renaming this package? My motivation is three-fold:
- rhash.sourceforge.net is the canonical project homepage, according to the github repo
- the github namespace is basically a "namespace of last resort" for software that doesn't have a homepage elsewhere
- the capital letters in the URL/package name bother me slightly ;)

Since this doesn't seem to be referenced in other package.yml files, I think you're probably the only one this might impact, though the odds of it impacting others only go up over time.


mxcl:
This is my bad, I should have checked these details before renaming. Thanks @jhheider.

@jhheider please prune the S3 as well.

magnusviri:
Please do! I am not trying to create a lot of work for you guys by creating a bunch of bad packages. I have some level of reassurance that I'm doing it right, but I don't feel like I have any level of authority in this area. So I'm hoping someone else looks at everything I'm making with a critical eye and makes corrections or at least pokes me to fix things and do it better in the future!

jhheider:
We usually do (we like having our code reviewed and questioned, too), but this one got in. Fortunately, the sooner it's caught, the easier it is to fix. Ironically, we caught it because of https://github.com/teaxyz/www/pull/216, so that's a nice bit of synergy.

PR #67, opened 2022-12-07
feat(projects): add rustls

mfts:
Rustls is a library that builds to a `.rlib` file; currently not sure how we'd consume it. 

`curl` can use `rustls` with their C-bindings here: https://github.com/rustls/rustls-ffi


pedrocruzio:
@mfts I would add the help command for the test.

mxcl:
Yeah, real tests where possible pls.

mfts:
closing this in favor of #87

PR #69, opened 2022-12-07
+cli.github.com

jonchang:
Moving from github/cli/cli namespace since it's easier to type and actually has a nice homepage associated with it. It's a github official project so it gets to use the `.github.com` subdomain instead of the `.io`.


mfts:
`gh` package dependency; if merged, this package needs to be updated here https://github.com/teaxyz/demos/pull/8/files#diff-8ef052d63bebd02db6849ed5a7403821ee169ac384fac2fa1dad6ccb24f3ec33R76

mxcl:
Longer term we need support for “renames”

PR #72, opened 2022-12-08
things that don't play nicely with llvm@15

jhheider:
This doesn't work as written, since build-deps.ts gives us both max llvm and our explicit range. https://github.com/teaxyz/pantry.core/pull/76 is probably the fix.


jhheider:
https://github.com/teaxyz/pantry.core/pull/76 is definitely a valid fix.

PR #74, opened 2022-12-09
fix librsvg

jhheider:
@mxcl this is a weird one. It appears to having issues with ${{deps.cairographics.org.prefix}}/lib/libcairo.la, which contains the build paths to the dependencies:

```
# Libraries that this one depends upon.
dependency_libs=' -L/opt/pixman.org/v0.40.0/lib/pkgconfig/../../lib /opt/pixman.org/v0.40.0/lib/libpixman-1.la -lm -L/opt/freedesktop.org/fontconfig/v2.14.0/lib/pkgconfig/../../lib -L/opt/freetype.org/v2.12.1/lib/pkgconfig/../../lib /opt/freedesktop.org/fontconfig/v2.14.0/lib/libfontconfig.la -L/opt/libexpat.github.io/v2.4.9/lib/pkgconfig/../../lib /opt/libexpat.github.io/v2.4.9/lib/libexpat.la -lfreetype -L/opt/libpng.org/v1.6.35/lib/pkgconfig/../../lib -L/opt/zlib.net/v1.2.13/lib/pkgconfig/../../lib /opt/libpng.org/v1.6.35/lib/libpng16.la -lz'
```

Obviously, we can fix this up on a per-*install* basis, but it seems like maybe the answer is to _kill all the .la files_:

https://www.linuxfromscratch.org/blfs/view/svn/introduction/la-files.html

### DIFF
--- a/projects/gnome.org/librsvg/package.yml
+++ b/projects/gnome.org/librsvg/package.yml
@@ -4,6 +4,7 @@ distributable:
 
 versions:
   github: GNOME/librsvg/tags
+  ignore: [2.55.90] # https://gitlab.gnome.org/GNOME/librsvg/-/issues/923 - sigh
 
 dependencies:
   cairographics.org: 1
@@ -16,7 +17,8 @@ build:
   dependencies:
     tea.xyz/gx/cc: c99
     tea.xyz/gx/make: '*'
-    rust-lang.org: 1
+    rust-lang.org/cargo: 0
+    rust-lang.org: ^1.63
     freedesktop.org/pkg-config: ^0.29
     gnome.org/gobject-introspection: 1
     python.org: 3


jhheider:
maybe not: https://github.com/teaxyz/pantry.extra/actions/runs/3660509125/jobs/6187738911

mxcl:
`.la` files aren't often used but yeah, I have as yet not made a “fix” script for them

mxcl:
You gotta figure out who is referencing `/opt/cairographics.org/v1.16.0/lib/libcairo.la` (possibly deeper in the stack). Or ask @jonchang to do it.

We could fix up all `.la` files (I doubt they support a relocatable attribute so we may need to do that as a post-install step), or maybe deleting them all will work.

jhheider:
Yeah, deleting didn't seem to help. I'll look into making them relative. We'll need post-install stuff eventually, but hopefully not for this.

Ironically, this is the only package we have so far gated by this.

mxcl:
I am a little concerned that this *was working* but now isn't, possibly something got a version bump which changed build system stuff.

jonchang:
Most package managers unconditionally blow away `.la` files. I think it would be safe to add something that removes them when bottling (on tea's side) or on install (on user's side), maybe even both (so we don't have to rebottle the world).

jhheider:
The hidden trickster was `fontconfig`, which had this in it's `.la`:
```libtool
# Libraries that this one depends upon.
dependency_libs=' -L/opt/freetype.org/v2.12.1/lib/pkgconfig/../../lib -lfreetype -L/opt/libexpat.github.io/v2.4.9/lib/pkgconfig/../../lib /opt/libexpat.github.io/v2.4.9/lib/libexpat.la'
```
Getting rid of the `.la` files is a lot better than having to fix all that on the fly, _assuming it never causes an issue (and I'm optimistic).

PR #77, opened 2022-12-15
+pnpm.io

getneil:
[pnpm](https://pnpm.io/) a better and faster node package manager


mxcl:
Like our packaging guide recommends, see what brew do: https://github.com/Homebrew/homebrew-core/blob/master/Formula/pnpm.rb

projects/pnpm.io/package.yml
```diff
@@ -0,0 +1,26 @@
+distributable:
+  url: https://github.com/pnpm/pnpm/archive/refs/tags/v{{ version }}.tar.gz
```
mxcl: use: `https://registry.npmjs.org/pnpm/-/pnpm-{{version}}.tgz` per brew formula

projects/pnpm.io/package.yml
```diff
@@ -0,0 +1,26 @@
...
+
+dependencies:
+  nodejs.org: '>=16'
+
+build:
+  dependencies:
+    curl.se: '>=5'
+    nodejs.org: '>=16'
+  script:
+    curl -L https://get.pnpm.io/v6.16.js | node - add --global pnpm@next-7 npm@7
```
mxcl: Seems like the tarball is in a ready to execute format: https://github.com/Homebrew/homebrew-core/blob/master/Formula/pnpm.rb#L29-L33

projects/pnpm.io/package.yml
```diff
@@ -0,0 +1,26 @@
...
+
+provides:
+  - bin/pnpm
+
+dependencies:
+  nodejs.org: '>=16'
+
+build:
+  dependencies:
+    curl.se: '>=5'
```
mxcl: don't forget to drop this

projects/pnpm.io/package.yml
getneil: thank you for this big help.

projects/pnpm.io/package.yml
```diff
@@ -0,0 +1,26 @@
...
+dependencies:
+  nodejs.org: '>=16'
+  npmjs.com: '>=8'
+
+build:
+  script: |
+    npm install
+    export PNPM_HOME=root/Library/pnpm
+    export PATH="$PNPM_HOME:$PATH";
+    export PATH="$(pwd)/bin:$PATH";
```
mxcl: your script doesn't move anything to `{{prefix}}` so nothing is installed. Confused what you're doing anyway, all brew does is move a few files: https://github.com/Homebrew/homebrew-core/blob/master/Formula/pnpm.rb#L29-L33

getneil: sorry did not fully understand the purpose of {{prefix}}, i just wrongly assumed pwd would be where to just put stuff. but thanks im finally getting a different error now.

projects/pnpm.io/package.yml
```diff
@@ -0,0 +1,26 @@
+distributable:
+  url: https://registry.npmjs.org/pnpm/-/pnpm-{{version}}.tgz
+  strip-components: 1
+
+versions:
+  github: pnpm/pnpm
```
getneil, line 6: dont know yet, but i feel im missing something here... hmmmm

projects/pnpm.io/package.yml
```diff
@@ -0,0 +1,26 @@
+distributable:
+  # url: https://registry.npmjs.org/pnpm/-/pnpm-{{version}}.tgz
+  url: https://registry.npmjs.org/pnpm/-/pnpm-7.18.2.tgz
```
getneil: i think pnpm is returning a bunch of weird versions with chars.

mxcl:
Excellent!

mxcl:
I do wonder if you were building locally first? That way is much faster than using CI/CD to evaluate build steps.

getneil:
> I do wonder if you were building locally first? That way is much faster than using CI/CD to evaluate build steps.

i tried first trying to run it directly failed on that, tried to dockerize also (encountered setting up linux on m1)
then i tried to run the entire pipeline using https://github.com/nektos/act the only part that run was the `get-diff`
then I tried using my mac mini intel there some parts started working correctly for the linux part

mxcl:
what did you try? What failed?

If you have issues with tea please tell us about them here or in Slack, if you are having issues then hundred other people also are and they also aren't reporting them.

PR #90, opened 2022-12-26
+bun.sh

### LINKED ISSUE #1: +bun.sh
Let’s try to launch with bun.

Last I looked it needed its own copy of zig to build. Fortunately we have a name spacing system so it can be `bun.sh/zig`

OTOH this could be a good case for just taking their binaries. We can do this provided the runtime dependencies are either zero or well formed. We should also insist on GPG signed binaries for these scenarios since we want to ensure we are getting what they claim they have built (we intend to sign our own bottles too, but this is about levels of trust).

mxcl:
* Closes #1
* Using bun.sh binaries because we'll probs need to do the same for ziglang because ziglang OOMs during builds in CD
* However in general we are happy to use the binaries the project built—they know what they are doing and we should trust them in that.
* HOWEVER we should insist that they sign their binaries. Bun do not.
* ALSO really we should insist on signed sources, see comments in the YAML for more discussion

### DIFF
--- a/projects/bun.sh/package.yml
+++ b/projects/bun.sh/package.yml
@@ -1,0 +1,38 @@
+distributable:
+  url: https://github.com/oven-sh/bun/archive/refs/tags/bun-v{{version}}.tar.gz
+
+warnings:
+  - vendored
+
+versions:
+  github: oven-sh/bun
+  strip: /^Bun /
+
+#FIXME proper system for re-using pre-built binaries
+# we must require the vendor to provide signatures against a published public
+# key. If they don’t then really we should build ourselves or warn the user
+# about the fact.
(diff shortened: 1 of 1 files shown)


mxcl:
well great, hangs in `bun --help` in the x86 macOS runner. Will be hard to debug.

mxcl:
k it works fine locally on an intel Mac (macOS 12 though, not 11)

jhheider:
> k it works fine locally on an intel Mac (macOS 12 though, not 11)

that is painful. limit to earlier release? or did they never support 11?

jhheider:
seems like there were hangs _reported_ in darwin x86-64 in august, but they've minor bumped 3 versions since then: https://github.com/oven-sh/bun/issues/1121

PR #93, opened 2022-12-30
+poetry

mxcl:


jonchang:
The issue here is that in `$prefix/libexec/bin`, there's a poetry.py that itself calls `from poetry.console... import ...`. But python by default includes the script directory in the current search path, so it tries to import stuff from itself and fails, since `poetry` is not a module. The fix that I've implemented to test this in https://github.com/teaxyz/pantry.extra/pull/103 adds the `-P` option to the python shebang, which [restricts python from looking in the script directory](https://docs.python.org/3/using/cmdline.html#cmdoption-P) for its module search path. 

Ideally the `python-venv.sh` build script would be fixed. I suggest:

* do not rename the python script here: https://github.com/teaxyz/pantry.core/blob/a9344e6e6258af68a102593720238bd84e6f0fa2/scripts/brewkit/python-venv.sh#L20
* ... and also don't create this link: https://github.com/teaxyz/pantry.core/blob/a9344e6e6258af68a102593720238bd84e6f0fa2/scripts/brewkit/python-venv.sh#L22-L23
* instead, have this script only exist in `$prefix/bin` as a real file (not a symlink into libexec/bin): https://github.com/teaxyz/pantry.core/blob/a9344e6e6258af68a102593720238bd84e6f0fa2/scripts/brewkit/python-venv.sh#L28-L41

nihaals:
A note about `-P`, it would restrict Poetry to only work with 3.11+, but you may want to use Poetry with an older Python version as its the version used by Poetry when it creates a venv (I believe it defaults to the latest version if it finds multiple installs). One workaround might be creating a venv for Poetry with a recent version but not exposing that recent version to Poetry by not adding it to PATH (unless a specific Python version isn't provided so defaulting to that version) but that might be too hacky.

There is a way to use a specific Python version with Poetry if you have multiple installed, but the expected behaviour would be that it defaults to the version the user has also included in their dependencies as that's their project Python version.

jonchang:
Calling `python -P` is merely a hack to demonstrate the issue. I suggest an alternate fix later on in my comment. With regards to versions I think it would be possible to pin an explicit python version in a call to tea... something like `tea +python-poetry.org +python.org@3.10.8` and then use `poetry env` to use the right version.

nihaals:
> Calling `python -P` is merely a hack to demonstrate the issue. I suggest an alternate fix later on in my comment.

I meant to only add some extra context for if `-P` was used as a temporary solution and merged, I agree that modifying the script would work well.

> With regards to versions I think it would be possible to pin an explicit python version in a call to tea... something like `tea +python-poetry.org +python.org@3.10.8` and then use `poetry env` to use the right version.

I was thinking of something similar in terms of `tea +python-poetry.org +python.org@3.10.8`, although I think it would be nice if running `poetry env` wasn't required and it would automatically use e.g. 3.10.8 to avoid impacting workflows and make it as smooth as possible to use.

mxcl:
@jonchang many thanks, this was the information I needed to implement a minimal fix in the mk-env script. Sorry for the delay in getting to this. I was knee deep working out tea/cli 0.19.

I note that my minimal fix doesn’t work for pipx, so will do the other things you suggested later today.

PR #97, opened 2022-12-31
+gnu.org/{gawk,mpc,mpfr,gcc,binutils,texinfo}

### LINKED ISSUE #95: +gnu.org/gcc
https://gcc.gnu.org/

often needed for rust crates

jhheider:
Done, I think.

- adds gnu.org/{gawk,mpc,mpfr,gcc,binutils,texinfo} - all required to get a working build of gcc
- bootstraps gcc from gcc on linux, and binutils from binutils (for `as`). Once we have live versions, we can remove the bootstrap adaptations
- everything should have tests to give us at least a little confidence in the outcomes
- this should allow us to build stuff (musl cross compiler comes to mind) that **strongly** prefer gcc to clang (though we should do our best to prefer clang).
- this should let us build and provide or cross-link glibc on linuxen that are too old, which should broadly increase our supported OSen.
- i can't think of any more bullet points because I'm just so happy that this builds and tests

~wip; all of these build locally on darwin-aarch64 except glibc and working in docker will be annoying but necessary. might as well smoke out the other builds first.

will depend on https://github.com/teaxyz/pantry.core/pull/83~


projects/gnu.org/binutils/package.yml
```diff
@@ -0,0 +1,54 @@
...
+    gnu.org/texinfo: '*'
+    darwin:
+      tea.xyz/gx/cc: c99
+    linux:
+      gnu.org/gcc: '*'
+  script: |
+    # FIXME: bootstrap - we need `as` from binutils to build with gcc
+    if test "{{hw.platform}}" = "linux"; then
+      if command -v sudo; then
+        sudo apt-get --yes install binutils
```
mxcl, line 22: We'll formalize a “bootstrap” key at some point. Which is allowed to source stuff from other places.

mxcl: But yeah, let's not make a habit of locking ourselves to `apt` based systems.

mxcl:
dang, a 14,000 line diff?!?!

so yeah, since we clone pantries for every user of tea, we may want to optimize that a little later. In theory some diffable database format that we can incrementally update and we keep updated server side would be great.

Ideally in a decentralized, unavailable manner.

projects/gnu.org/gcc/package.yml
```diff
@@ -0,0 +1,123 @@
...
+      - --disable-bootstrap
+      - --disable-nls
+    linux:
+      ARGS:
+        - --disable-multilib
+        - --enable-default-pie
+    darwin:
+      ARGS:
+        # Reliance on CLT hard path is yuck.
+        - --with-sysroot=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
```
mxcl, line 86: this means that the package depends on `apple.com/xcode/clt` or whatever I called it (see hacks.ts)

projects/gnu.org/gcc/package.yml
```diff
@@ -0,0 +1,123 @@
...
+
+provides:
+  # shouldn't need to support these.
+  # - bin/{{hw.arch}}-{{##MANUFACTURER}}-{{hw.platform}}{{##KERNEL-VERSION}}-c++
+  # - bin/{{hw.arch}}-{{##MANUFACTURER}}-{{hw.platform}}{{##KERNEL-VERSION}}-g++
+  # - bin/{{hw.arch}}-{{##MANUFACTURER}}-{{hw.platform}}{{##KERNEL-VERSION}}-gcc
+  # - bin/{{hw.arch}}-{{##MANUFACTURER}}-{{hw.platform}}{{##KERNEL-VERSION}}-gcc-{{version}}
+  # - bin/{{hw.arch}}-{{##MANUFACTURER}}-{{hw.platform}}{{##KERNEL-VERSION}}-gcc-ar
+  # - bin/{{hw.arch}}-{{##MANUFACTURER}}-{{hw.platform}}{{##KERNEL-VERSION}}-gcc-nm
+  # - bin/{{hw.arch}}-{{##MANUFACTURER}}-{{hw.platform}}{{##KERNEL-VERSION}}-gcc-ranlib
```
mxcl: probs should just delete them.

projects/gnu.org/gcc/package.yml
```diff
@@ -0,0 +1,123 @@
...
+  - bin/c++
```
mxcl: we shouldn't provide this. Instead (I think) `tea.xyz/gx/cc` does. This is how llvm does it too. Then we make the gx version able to use gcc or llvm based on which is in `PATH`.

tea provides `TEA_PKGS` now to help with this.

mxcl (approved):
Epic effort. No major issues IMO. Some comments that you can address if you have the energy. Approved.

Though `c++` should be removed IMO.

projects/gnu.org/binutils/package.yml
jhheider: agreed. this is to bootstrap with our current infrastructure, and will be removed once tea can provide gcc for bootstrapping.

jhheider:
> dang, a 14,000 line diff?!?!
> 
> so yeah, since we clone pantries for every user of tea, we may want to optimize that a little later. In theory some diffable database format that we can incrementally update and we keep updated server side would be great.
> 
> Ideally in a decentralized, unavailable manner.

From the maintainer. Other package managers just use the diff by URL, but that feels antithetical to our security model (though there are instances where we do use URL-based work). I can compress it, certainly, and it's just text, but then someone has to open it to see what we're applying, if they care.

projects/gnu.org/gcc/package.yml
jhheider: good point. i'll update the deps.

projects/gnu.org/gcc/package.yml
jhheider: we could name it gc++ to allow it to be located but not the default. not unheard of for gnu packages.

mxcl:
> but that feels antithetical to our security model 

yeah this is correct, and I have previously said that we should commit the patches. However considering users download the whole pantry currently, I feel it's wrong of us to add this TBH.

Idea: could go in another branch. Makes getting it a little painful, but we could add helpers for that

jhheider:
> Idea: could go in another branch. Makes getting it a little painful, but we could add helpers for that

could also have a pantry/blobs repo specifically for finding such stuff. many ways to skin this cat.

PR #102, opened 2023-01-05
fix cli.github.com build

jhheider:
use clang with cgo

### DIFF
--- a/projects/cli.github.com/package.yml
+++ b/projects/cli.github.com/package.yml
@@ -20,6 +20,8 @@ build:
   dependencies:
     go.dev: ^1.18
     tea.xyz/gx/make: '*'
+    linux:
+      tea.xyz/gx/cc: c99 # for cgo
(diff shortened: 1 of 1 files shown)


projects/cli.github.com/package.yml
```diff
@@ -20,6 +20,8 @@ build:
   dependencies:
     go.dev: ^1.18
     tea.xyz/gx/make: '*'
+    linux:
+      tea.xyz/gx/cc: c99 # for cgo
```
mxcl, line 24: oh ok, go.dev used to depend on this, but we made it a companion. Strictly this is not a linux dep, macOS needs it too. gx/cc on Mac checks if the Xcode CLT is installed.

jhheider: 17922c85cfb7ca65b81dc7247a8f16b101f89906

PR #105, opened 2023-01-06
+starship

johnallen3d:
Adds the Starship prompt.

### DIFF
--- a/projects/crates.io/starship/package.yml
+++ b/projects/crates.io/starship/package.yml
@@ -1,0 +1,22 @@
+distributable:
+  url: https://github.com/starship/starship/archive/refs/tags/v{{ version }}.tar.gz
+  strip-components: 1
+
+provides:
+  - bin/starship
+
+versions:
+  github: starship/starship/tags
+  strip: /v/
+
+build:
+  dependencies:
+    cmake.org: ">=3.5"
+    tea.xyz/gx/make: "*"
+    rust-lang.org: ">=1.60"
+    rust-lang.org/cargo: "*"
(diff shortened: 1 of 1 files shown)


johnallen3d:
This is on the [top 300 list](https://github.com/teaxyz/pantry.extra/issues/99).

jhheider:
`CMake Error: CMake was unable to find a build program corresponding to "Unix Makefiles".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.`

Likely needs a build-dep on `tea.xyz/gx/make: '*'`.

mxcl:
First package that probs needs some follow up instructions or the user is left without a clue on how to use it.

`brew` used a `caveats` key. With time it was apparent caveats were usually just “fyi”s, so some equivalent.

Longer term would be great to have some awareness of shell-additions built in to tea in some manner. Presumably via our shell magic additions.

jhheider:
Looks like this is the test failure:
```sh
+ starship version
error: The subcommand 'version' wasn't recognized

  Did you mean 'session'?

  If you believe you received this message in error, try re-running with 'starship -- version'

Usage: starship <COMMAND>

For more information try '--help'
```

Should be an easy fix, if that's all of it.

johnallen3d:
Good catch, thanks @jhheider. I've made the updated... I did review the available commands but couldn't identify an option that would work universally across build systems to test with.

mxcl:
Has a homepage https://starship.rs we should name the package that.

mxcl:
I'll merge and rename

PR #107, opened 2023-01-07
fish wants git for its build

jhheider:
- incorporates https://github.com/teaxyz/pantry.extra/pull/118
- adds git dep for build to clone its own PCRE
- requires https://github.com/teaxyz/pantry.core/pull/103


mxcl:
broken rpath in our git?

```
dyld[44521]: Library not loaded: '@rpath/zlib.net/v1/lib/libz.1.2.13.dylib'
[322](https://github.com/teaxyz/pantry.extra/actions/runs/3864361090/jobs/6587150813#step:8:323)
  Referenced from: '/opt/git-scm.org/v2.39.0/libexec/git-core/git'
[323](https://github.com/teaxyz/pantry.extra/actions/runs/3864361090/jobs/6587150813#step:8:324)
  Reason: tried: '/opt/git-scm.org/v2.39.0/libexec/git-core/../../../zlib.net/v1/lib/libz.1.2.13.dylib' (no such file), '/opt/git-scm.org/v2.39.0/libexec/git-core/../../../zlib.net/v1/lib/libz.1.2.13.dylib' (no such file), '/usr/local/lib/libz.1.2.13.dylib' (no such file), '/usr/lib/libz.1.2.13.dylib' (no such file)
```

mxcl:
```
/opt/git-scm.org/v2.39.0/libexec/git-core/../../../zlib.net/v1/lib/libz.1.2.13.dylib
```

is indeed wrong, needs another `../`

mxcl:
Confirmed locally. This is a gnarly issue to test for since it only happens because macOS doesn’t propagate `DYLD_LIBRARY_PATH` beyond one instantiation for SIP reasons.

edit: Probs we should unset these variables during tests to catch this sort of thing.

jhheider:
Extremely poor hack that seems to work:
```sh
    # FIXME: this is needed due to macOS env security features
    mkdir bin
    cat <<-EOF >>bin/git
    #!/bin/sh

    export DYLD_FALLBACK_LIBRARY_PATH=$DYLD_FALLBACK_LIBRARY_PATH
    $(which git) "\$*"
    EOF
    chmod 555 bin/git
    export PATH=$PWD/bin:$PATH
```

jhheider:
I take it back. Worked locally but not in GHA.

jhheider:
No, the incredibly more heinous and basically unacceptable workaround is:
```sh
    # FIXME: The most hack to work around macOS ENV security hitting CMake.
    ln -s $(tea --prefix)/zlib.net $(tea --prefix)/git-scm.org/
    mkdir $(tea --prefix)/git-scm.org/gnu.org
    ln -s $(tea --prefix)/gnu.org/gettext $(tea --prefix)/git-scm.org/gnu.org/
```

mxcl:
lol no work around we need to fix git

jhheider:
> lol no work around we need to fix git

I _think_ that means all hard links need to be in the same directory. So, bin/git can't be a link to libexec/git-core/git (since they need different @rpaths), but bin/git and bin/git-whatever can be. Which, I believe, means the code we have that fixed hardlinks needs to check for identical dirnames in addition to its other checks.

The good news is when we get this right, we can add it to the `test:` block for git to check all the rpaths.

mxcl:
> I think that means all hard links need to be in the same directory. So, bin/git can't be a link to libexec/git-core/git (since they need different @rpaths), but bin/git and bin/git-whatever can be. Which, I believe, means the code we have that fixed hardlinks needs to check for identical dirnames in addition to its other checks.

eek yes, you're probs right.

projects/fishshell.com/package.yml
```diff
@@ -26,6 +27,12 @@ build:
 
     mkdir build
     cd build
+
+    # FIXME: The most hack to work around macOS ENV security hitting CMake.
+    ln -s $(tea --prefix)/zlib.net $(tea --prefix)/git-scm.org/
```
mxcl: `$TEA_PREFIX` is exposed FYI, but this is fine.

mxcl:
K! I got a solution, if we drop the `GIT_PREFIX/libexec/git-core` directory then the rpath is the same number of `..`s from both locations. On it…

jhheider:
this will depend on https://github.com/teaxyz/pantry.core/pull/103 and I will wrap in https://github.com/teaxyz/pantry.extra/pull/118

jhheider:
Thanks for the smart `git` fix, @mxcl .

PR #108, opened 2023-01-09
+redis.io

getneil:


projects/redis.io/package.yml
```diff
@@ -0,0 +1,31 @@
...
+  - bin/redis-server
+  - bin/redis-cli
+  - bin/redis-benchmark
+
+dependencies:
+  openssl.org: ^1
+
+build:
+  dependencies:
+    gnu.org/make: '>=4.3'
```
jhheider: Does it require GMake specifically, or is `tea.xyz/gx/make: '*'` sufficient?

projects/redis.io/package.yml
```diff
@@ -0,0 +1,31 @@
...
+    freedesktop.org/pkg-config: '>=0.29'
+    tea.xyz/gx/cc: c99
+  script:
+    PREFIX={{prefix}}
+    BUILD_TLS=yes
+    make install
```
jhheider: I'd change this to:
```yaml
script: make install
env:
  PREFIX: ${{prefix}}
  BUILD_TLS: yes
```

projects/redis.io/package.yml
```diff
@@ -0,0 +1,31 @@
...
+    # TODO: configure default redis.conf if necessary
+test:
+  script: |
+    redis-server --test-memory 2
+    redis-cli --version
+    redis-benchmark --version
```
jhheider: Good number of tests. Two notes:

- if `script:` is the only subkey, you can pass the script directly on the `test:` key
- is it worth spinning up the server, running a couple of set/gets then spinning it down, to be sure it's working?

jhheider (approved):
Looks good. Mostly just style questions/comments. Eventually, we'll want some kind of service runner for `tea` bins, but no need to hold up packaging for it (indeed, having some servers packaged should make it easier to write/test).

projects/redis.io/package.yml
getneil: added.

projects/redis.io/package.yml
getneil: thanks

projects/redis.io/package.yml
```diff
@@ -0,0 +1,47 @@
...
+    echo $redis_value
+
+    redis-cli shutdown
+
+    if [ $redis_value = "value123" ]; then
+      echo "redis works!"
+    else
+      echo "redis install doesnt work!"
+      set -e
+    fi
```
jhheider: simpler exit condition might just be `test "$redis_value" = "value123"` and let the error code tell the story.

jhheider: or, even just replace the latter half with `test "$(redis-cli --raw GET key123)" = "value123"`

jhheider:
I'll fix that build script error :/

projects/redis.io/package.yml
getneil: tested replacement and it works

projects/redis.io/package.yml
getneil: that is much cleaner. added it.

mxcl:
lovely job here. The test especially is magnificent. Good understanding of tea. I wonder about the cores key in build tho. Is that something we added?

jhheider:
Yes, for larger requirements: https://github.com/teaxyz/pantry.zero/wiki/Structure-of-a-package.yml#cores

That note should say it defaults to 2 if omitted...

jhheider:
Wiki updated.

PR #110, opened 2023-01-09
+sl (bonus: +tcl+expect)

mxcl:


mxcl:
```
gcc -O -o sl sl.c -lncurses
ld.lld: error: unable to find library -lncurses
```

our ncurses only provides libncurses*w*. Not sure about this history on this and if we can provide both or what. /cc @jonchang

jonchang:
Yeah, technically, `libncursesw` is for wide character support (i.e., unicode and its encodings), while `libncurses` is for legacy encodings. In practice they're basically compatible with each other and giving an app `libncursesw` when it asks for `libncurses` (or even `libcurses`) will not end the world.

> The ncurses and ncursesw libraries are reasonably source-compatible. That is, an application written for “ncurses” will build with “ncursesw”. But it will behave differently in response to your locale settings. (Some distributors, who do not care about the differences, have chosen to merge the names together as “ncurses”).

https://invisible-island.net/ncurses/ncurses.faq.html

jhheider:
This should be tested, but I have faint memories of ncursesw being symlinked as ncurses to support software that is not looking for it.

Edit: yes, that seems to be the last line of faq. Knew I'd seen it.

jhheider:
I might have accidentally packaged `Tcl` and `expect` on my quest for a proper test for this 300-line C program.

jhheider:
You good with adding a non-curses `-v` flag for testing-sake, @mxcl ?

mxcl:
epic effort.

Yes, I think we should prefer *not* to add stuff solely for tests, but until we have a nice idea on how to test these sorts of things, this is harmless. `sl` doesn't actually invoke `ls` after, unlike `gti`.

I note that brew’s test is `sl -c`: https://github.com/Homebrew/homebrew-core/blob/master/Formula/sl.rb#L32-L34

jhheider:
> I note that brew’s test is `sl -c`: https://github.com/Homebrew/homebrew-core/blob/master/Formula/sl.rb#L32-L34

I saw that too. GHA using our test-harness doesn't like opening terminal by ncurses, and it defeated my best efforts at working around it, sadly. But I _believe_ expect is the right way (unless it's actually running in a headless `tmux`, but that seems messy, and silly for *this* program).

PR #111, opened 2023-01-10
+github.com/jart/blink

jhheider:
https://github.com/jart/blink

No releases/tags yet, but we can package the current commit as v0.0.0 until they're ready.

### DIFF
--- a/projects/github.com/jart/blink/package.yml
+++ b/projects/github.com/jart/blink/package.yml
@@ -12,6 +12,9 @@ versions:
   - 0.0.0
 
 build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
   script: |
     make -j {{hw.concurrency}}
     mkdir {{prefix}}/bin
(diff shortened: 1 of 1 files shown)


projects/github.com/jart/blink/package.yml
```diff
@@ -0,0 +1,32 @@
...
+  url: https://github.com/jart/blink/archive/13df12124d69aba8a7f74803715af36ed629b349.tar.gz
+  strip-components: 1
+
+provides:
+  - bin/blink
+  - bin/blinkenlights
+
+versions:
+  # github: jart/blink/tags  FIXME once they start tagging/releasing
+  - 0.0.0
```
mxcl, line 12: fun

projects/github.com/jart/blink/package.yml
```diff
@@ -0,0 +1,32 @@
...
+  test: |
+    make check
+    make check2
+    make emulates
+
+test:
+  dependencies:
+    curl.se: '*'
+  script: |
+    curl -O https://raw.githubusercontent.com/jart/blink/13df12124d69aba8a7f74803715af36ed629b349/third_party/cosmo/tinyhello.elf
```
mxcl, line 31: probs worth us copying this into our repo. Though I admit we have done this in other places maybe we shouldn't. But yeah. Links die.

jhheider: Hopefully, linked to a specific GitHub commit that matches our distributable.url should be as safe as we can get. But I did consider loading it.

PR #113, opened 2023-01-11
+dav1d

kevmo314:


mxcl:
oof, super weird segfault on linux x86-64.

Currently we cannot merge if some platforms fail to build, but we should figure out a solution at some point /cc @jhheider

jhheider:
My very low-IQ solution for numactl (linux only) was:
```sh
if test "{{hw.platform}}" = "darwin"; then
  mkdir -p "{{prefix}}"
  touch "{{prefix}}/not-available-on-darwin"
  exit 0
fi
```

which is very poor.

mxcl:
Sadly not a good enough workaround for things that `provides:` at this time. At some point we will need to allow bottles to go through if they only support platform X and not Y since in reality this is a thing (albeit much rarer than it used to be).

jhheider:
It'll be less rare once windows native is ready, unfortunately.

PR #116, opened 2023-01-13
+xc

mfts:
package `xc` package because we will mention it in the new readme of cli so it's just appropriate to have it packaged.

### DIFF
--- a/projects/xcfile.dev/package.yml
+++ b/projects/xcfile.dev/package.yml
@@ -4,7 +4,7 @@ distributable:
 
 versions:
   github: joerdav/xc/tags
-  strip: /v/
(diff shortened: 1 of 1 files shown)


mxcl:
Check some of the other pantry entries that use `go` to fix this issue. You have to add a few flags to the go compiler.

mxcl:
eg https://github.com/teaxyz/pantry.extra/blob/main/projects/charm.sh/gum/package.yml

jhheider:
green!

mxcl:
nice work everyone!

mfts:
Thanks for the fix and wow idea for a test 💡

PR #118, opened 2023-01-13
fishshell.com: drop cmake/ncurses patch

jonchang:
Once https://github.com/teaxyz/pantry.core/pull/92 is green and merged


jhheider:
Incorporated into https://github.com/teaxyz/pantry.extra/pull/107. Thanks, @jonchang !

Issue #99, opened 2023-01-03
The Top 300


mxcl:
### These are the top 300 packages in the world

> Well actually, 540+ because the list is topologically expanded to include all dependencies

Can you help us package these? (Start at the top! We ordered them so later pkgs depend on earlier ones)

- [ ] https://github.com/teaxyz/pantry.extra/issues/591
- [ ] https://github.com/teaxyz/pantry.extra/issues/594
- [ ] https://github.com/teaxyz/pantry.extra/issues/599
- [ ] https://github.com/teaxyz/pantry.extra/issues/687
- [ ] https://github.com/teaxyz/pantry.extra/issues/645
- [ ] https://github.com/teaxyz/pantry.extra/issues/671
- [ ] https://github.com/teaxyz/pantry.extra/issues/371
- [x] https://github.com/teaxyz/pantry.extra/issues/197
- [ ] https://github.com/teaxyz/pantry.extra/issues/718
- [ ] https://github.com/teaxyz/pantry.extra/issues/735
- [ ] https://github.com/teaxyz/pantry.extra/issues/730
- [ ] https://github.com/teaxyz/pantry.extra/issues/742
- [ ] https://github.com/teaxyz/pantry.extra/issues/743

<details><summary>completed</summary>

- [x] https://github.com/teaxyz/pantry.extra/issues/662
- [x] +pkg-config
- [x] https://github.com/teaxyz/pantry.extra/issues/498
- [x] https://github.com/teaxyz/pantry.extra/issues/499
- [x] https://github.com/teaxyz/pantry.extra/issues/500
- [x] +sqlite
- [x] https://github.com/teaxyz/pantry.extra/issues/501
- [x] +xz
- [x] https://github.com/teaxyz/pantry.extra/issues/502
- [x] https://github.com/teaxyz/pantry.extra/issues/503
- [x] https://github.com/teaxyz/pantry.extra/issues/504
- [x] https://github.com/teaxyz/pantry.extra/issues/206
- [x] https://github.com/teaxyz/pantry.extra/issues/505
- [x] https://github.com/teaxyz/pantry.extra/issues/218
- [x] https://github.com/teaxyz/pantry.extra/issues/201
- [x] https://github.com/teaxyz/pantry.extra/issues/506
- [x] +ninja 
- [x] +libpng 
- [x] https://github.com/teaxyz/pantry.extra/issues/193
- [x] +pcre2 
- [x] +gettext 
- [x] +meson 
- [x] +freetype 
[...]

PR #119, opened 2023-01-13
+nmap

mfts:
#99 

using the two-digit versions workaround like here: https://github.com/teaxyz/pantry.core/pull/98

### DIFF
--- a/projects/nmap.org/package.yml
+++ b/projects/nmap.org/package.yml
@@ -1,0 +1,28 @@
+distributable:
+  url: https://nmap.org/dist/nmap-{{version.raw}}.tgz
+  strip-components: 1
+
+versions:
+  - v7.93
+
+build:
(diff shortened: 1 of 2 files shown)


jhheider:
all green! good work, @mfts !

projects/nmap.org/package.yml
```diff
@@ -0,0 +1,32 @@
+distributable:
+  url: https://nmap.org/dist/nmap-{{version.raw}}.tgz
+  strip-components: 1
+
+versions:
+  - 7.93
+
+dependencies:
+  openssl.org: '*'
```
mxcl, line 9: Yeah, probs not true though. At some point we will package openssl 3 and I'm pretty sure it’s API is very different to 1.1.1

mxcl:
Super glad we have this.

PR #120, opened 2023-01-16
prql-lang.org 0.4.0 changes some things; we must adapt

jhheider:
@mfts the only constant in life is change.

### DIFF
--- a/projects/prql-lang.org/package.yml
+++ b/projects/prql-lang.org/package.yml
@@ -10,10 +10,25 @@ build:
     rust-lang.org/cargo: '*'
   script: |
     cd ./prql-compiler
-    cargo install --path . --root {{prefix}}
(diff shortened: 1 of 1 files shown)


projects/prql-lang.org/package.yml
```diff
@@ -10,10 +10,25 @@ build:
...
+  else
+    COMPILER=prqlc
+  fi
+  echo "from employees | filter has_dog | select salary" | $COMPILER compile
 
 provides:
-  - bin/prql-compiler
\ No newline at end of file
+  - bin/prql-compiler
+  - bin/prqlc
```
mxcl, line 34: Fun that we may need version keys for this soon.

jhheider: it's true. the interesting thing about doing this, as long as no conflicts are created, is that the change is transparent to the user.

PR #124, opened 2023-01-17
+zoxide

tnxz:


tnxz:
this needs to evaluate the output of `zoxide init $SHELL` which varies, how can this be accomplished any thoughts?

mxcl:
> this needs to evaluate the output of zoxide init $SHELL which varies, how can this be accomplished any thoughts?

you can force the shell to bash (we provide the package). You could do that for each.

Still the test mainly should verify that the package doesn't segfault (surprisingly common) then after that test any functionality that may be missing, eg. with python we should test that all its modules that depend on tea packages actually function (we still don't 😒). Eg. with git we test that it is relocatable since we had to put work into ensuring it was.

With cargo stuff we can be fairly sure it's fine if it runs.

Better tests *always welcome* though.

tnxz:
ok so there was no need for anything fancy just based on your shell you could add the init script say `zsh`
`eval "$(zoxide init zsh)" ` in `~/.zshrc` after the tea magic script, then `zoxide` is installed on startup and just works

so any program that requires an init script to be in the shell startup file just needs to go after the tea magic script.

mxcl:
sorry I thought you were asking about the `test` key in the yaml.

In terms of integrating it, yes, your suggestion seems correct.

I'm tempted to add functionality for integration like this. I'm also tempted to make the system flexible enough that others can do it by adding arbitrary keys to package yaml.

sanchitram1:
> you can force the shell to bash (we provide the package). You could do that for each.

What package is this?

I'm packaging up `zsh-syntax-highlighting`, and one of the steps is to add `source $(pwd)/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh` to `.zshrc`, and I keep running into a `Read-only file system` error.

PR #125, opened 2023-01-17
+lsd

johnallen3d:


projects/github.com/peltoche/lsd/package.yml
```diff
@@ -0,0 +1,19 @@
+distributable:
+  url: https://github.com/Peltoche/lsd/archive/refs/tags/{{ version }}.tar.gz
```
jhheider, line 2: I think you want a `strip: 1` key here. It's likely unpacking into a `lsd-{{version}}` directory (or similar) and that will strip the first path segment off, so the build code is in the root.

Edit: er, yes, `strip-components:`.

johnallen3d: Thanks for that explanation. I've seen (and included) that line but didn't fully understand the implications. This built OK locally without it so I thought perhaps it wasn't required.

jhheider: It's _basically_ default behavior, given how things are packaged, but:

a) it's good to be explicit
b) it'd be annoying, at a minimum, to unroll if someone tarballed their files into the archive root

jhheider:
Looks good. Thanks @johnallen3d !

projects/github.com/peltoche/lsd/package.yml
mxcl: It does however need to be less cryptic. Was thinking of:

```yaml
uses: gnu.org/tar
   with: { strip-components: 1 }
```

To copy GitHub actions a bit.

PR #127, opened 2023-01-18
+xsv

johnallen3d:


mxcl:
We’ve typically been adding crates to the `crates.io` namespace. However I prefer namespacing them this way in several ways. And now that referencing the package explicitly is not a UX issue I'm not fussed.

At some point we'll add “redirects” and clean up.

Though I'm hoping for a longer term solution where we can automatically bottle crates without having to maintain package.ymls (same for python, npm, ruby, etc.)

mxcl:
Thanks for all your contributions!

johnallen3d:
Argh, I knew that! 🤦‍♂️ I'd happily move this if desired.

mxcl:
well for consistency we should but we already are somewhat inconsistent and there are some issues with renaming currently.

I'll let @jhheider decide.

mxcl:
> However I prefer namespacing them this way in several ways

By this I meant I prefer GitHub.com namespacing as then you get attachment to the *user*. My preference is only slight though and we already went the “stick them on crates.io” route.

The crates.io was chosen because we do intend to try to stop “repackaging” crates and be able to infer from crates.io the registry.

PR #131, opened 2023-01-18
+dagger

johnallen3d:


projects/dagger.io/package.yml
```diff
@@ -0,0 +1,34 @@
...
+    mkdir -p "{{ prefix }}"/bin
+    mv dagger "{{ prefix }}"/bin
+  dependencies:
+    go.dev: ^1.18
+  env:
+    LDFLAGS:
+      - -s
+      - -w
+      - -X go.dagger.io/dagger/version.Version={{ version }}
+      - -X go.dagger.io/dagger/version.Revision=tea
```
johnallen3d, line 25: I tried to follow the Homebrew formula here but the version isn't printing as expected.

```sh
❯ /opt/homebrew/bin/dagger version
dagger 0.3.7 (2bd84ee24) darwin/arm64
❯ ~/.tea/dagger.io/v0.3.9/bin/dagger version
dagger devel () darwin/arm64
```

johnallen3d: Actually, I was looking at the forumla from `homebrew-core` but the latest is actually from [their tap](https://github.com/dagger/homebrew-tap/blob/main/dagger.rb). It looks like in this case they're pulling pre-built binaries from GitHub.

projects/dagger.io/package.yml
```diff
@@ -11,13 +11,13 @@ provides:
...
 build:
   script: |
-    export CGO_ENABLED=0
     go build -v -ldflags="$LDFLAGS" ./cmd/dagger
     mkdir -p "{{ prefix }}"/bin
     mv dagger "{{ prefix }}"/bin
   dependencies:
     go.dev: ^1.18
   env:
+    CGO_ENABLED: 0
```
johnallen3d, line 20: TIL! 👏

jhheider:
another one for the books!

johnallen3d:
Thanks for your help on this one!

PR #133, opened 2023-01-18
+surrealdb

johnallen3d:
Draft PR of a package without an official release (still in beta).

Related discussion: https://github.com/orgs/teaxyz/discussions/321

/cc @mxcl @jhheider


jhheider:
`error: linker `aarch64-linux-gnu-gcc` not found`

Dunno if `build.env.LD: ld` will fix that, but it's worth trying.

PR #138, opened 2023-01-20
+gleam+erlang

### LINKED ISSUE #342: +erlang
Formulae: https://formulae.brew.sh/formula/erlang

#138

jhheider:
- gleam.run compiles to >
- erlang.org

~(epmd segfaults when trying to bind inet in gha ubuntu. ugh.)~

resolves #342

### DIFF
--- a/projects/erlang.org/package.yml
+++ b/projects/erlang.org/package.yml
@@ -1,0 +1,87 @@
+distributable:
+  url: https://github.com/erlang/otp/releases/download/OTP-{{ version.raw }}/otp_src_{{ version.raw }}.tar.gz
+  strip-components: 1
+
+provides:
+  - bin/ct_run
+  - bin/dialyzer
+  - bin/epmd
+  - bin/erl
+  - bin/erlc
+  - bin/escript
+  - bin/run_erl
+  - bin/to_erl
+  - bin/typer
+
+interprets:
+  extensions: erl
+  args: escript
+
+versions:
+  github: erlang/otp
+  strip: /^OTP /
+
+dependencies:
+  openssl.org: '*'
+  invisible-island.net/ncurses: '*'
+
+runtime:
+  env:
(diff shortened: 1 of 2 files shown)


philipgiuliani:
Whats the status of this PR? I am interested in adding elixir-lang, but need erlang for it.

jhheider:
Stalled for erlang segfaulting. Feel free to look at it, if you want.

philipgiuliani:
Hi @jhheider,

I've done some experiments. I noticed that the only pipeline that fails, is using a pipeline hosted by GitHub, while the other ones are self-hosted.

I have tried to run the tests locally inside a container using the `ubuntu:latest` image and the tests passed as expected.

The error message we see in the CI means that the beam cannot connect to epmd. I guess the segfault we see is also from epmd which cannot start. The `escript` executes fine, but returns `econnrefused`.

epmd, by default, is binding to the public interface (0.0.0.0), it's possible that this is not allowed inside a Github CI runner? I couldn't find any resource/docs about that.

So one test I would suggest is to add `-kernel inet_dist_use_interface {127,0,0,1}` inside the test script. If this doesn't work we could try starting `epmd` as a test with the debug flag (`epmd -d`).

jhheider:
> So one test I would suggest is to add `-kernel inet_dist_use_interface {127,0,0,1}` inside the test script. If this doesn't work we could try starting `epmd` as a test with the debug flag (`epmd -d`).

let's try it. it seemed like maybe it was trying to create docker instances, which might be a problem.

jhheider:
Part of it was segfaults on x86-64 without PIE (as we often see in compiled C). Passing now.

mxcl:
beautiful work

mxcl:
Guess we now need https://hex.pm

jhheider:
😵

PR #140, opened 2023-01-23
+pytest.org

getneil:
this is required by `six` which might be required by `awscli`

### DIFF
--- a/projects/pytest.org/package.yml
+++ b/projects/pytest.org/package.yml
@@ -1,0 +1,28 @@
+distributable:
+  url: https://github.com/pytest-dev/pytest/archive/refs/tags/{{ version }}.tar.gz
+  strip-components: 1
+
+versions:
+  github: pytest-dev/pytest
+
+dependencies:
+  python.org: 3.7
+
+build:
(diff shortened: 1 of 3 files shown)


projects/pytest.org/package.yml
```diff
@@ -0,0 +1,28 @@
...
+  script: |
+    python-venv.sh {{prefix}}/bin/pytest
+
+provides:
+  - bin/pytest
+
+test:
+  dependencies:
+    python.org: 3.7
+    pip.pypa.io: 22.3
```
jhheider: Only need to define python dep at the top level (the other two are redundant).

Python should likely be >=3.7 and pip probably '*'.

Otherwise, it looks good.

jhheider (changes requested):
Better use of versions

projects/pytest.org/package.yml
getneil: suggestions applied.

jhheider:
looks good! thanks, @getneil !

mxcl:
Thanks @getneil !

PR #142, opened 2023-01-23
cgo needs a compiler

jhheider:


projects/ipfs.tech/package.yml
```diff
@@ -11,8 +11,16 @@ versions:
 build:
   dependencies:
     tea.xyz/gx/make: '*'
+    tea.xyz/gx/cc: c99
     go.dev: 1
+    gnu.org/patch: '*'
   script: |
+    if test "{{hw.platform}}" = "linux"; then
```
mxcl, line 18: This is just linux aarch64 right? We shouldn't apply fixes to more platforms than required.

mxcl: Also for all other go tools we do this in env, odd this requires a patch.

jhheider: yeah, it's because they don't pass any flags in their makefiles. I started with our SOP. ;)

jhheider: and our comments elsewhere say it's linux; I'll see if it's only aarch64.

jhheider: nope, x86_64 needs it too.

mxcl: oh ok.

PR #146, opened 2023-01-27
Gitignore `.DS_Store`

davdroman:
Bottling on macOS without gitignoring `.DS_Store` is a bit of a nightmare.

CleanShot 2023-01-27 at 20 58 47@2x

This should help alleviate things.

As a side note, maybe this should also be done on pantry.core and pantry.zero.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store


jhheider:
In general, we make those kind of systems-based exclusions at the user level rather than the repo. See: https://github.com/teaxyz/pantry.core/tree/main/projects/git-scm.org. Otherwise, you need to do it for every project you ever work on.

davdroman:
@jhheider TIL about user-level gitignoring. Thanks!

For anyone coming to this thread in the future:

```sh
$ mkdir -p ~/.config/git
$ echo .DS_Store > ~/.config/git/ignore
```

mxcl:
I’m not too fussed about allowing this, FYI, because most people don’t know you can global gitignore. And `.DS_Store` files suck.

PR #147, opened 2023-01-28
+gnu.org/glibc

### LINKED ISSUE #2131: Cannot be used for older `glibc` libs
Hi,

I tried setting `tea` up in Amazon Linux 2 (which I think is using an older version of `GLIBC`), and I'm getting the following errors:
```
/home/farzadmf/.tea/tea.xyz/v*/bin/tea: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by /home/farzadmf/.tea/tea.xyz/v*/bin/tea)
/home/farzadmf/.tea/tea.xyz/v*/bin/tea: /lib64/libc.so.6: version `GLIBC_2.27' not found (required by /home/farzadmf/.tea/tea.xyz/v*/bin/tea)
/home/farzadmf/.tea/tea.xyz/v*/bin/tea: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by /home/farzadmf/.tea/tea.xyz/v*/bin/tea)
```

jhheider:
- [x] debootstrap binutils/gcc
- [x] build glibc
- [x] use glibc without segfault

@mxcl // @jonchang I'd love some eyes on this. Currently, things using the new glibc like to segfault. I'm guessing I need to runtime.env.$some_env_vars to make it first on the list of `libc`s looked for.

### DIFF
--- a/projects/gnu.org/glibc/package.yml
+++ b/projects/gnu.org/glibc/package.yml
@@ -1,0 +1,114 @@
+distributable:
+  url: https://ftp.gnu.org/gnu/glibc/glibc-{{ version.raw }}.tar.gz
+  strip-components: 1
+
+versions:
+  url: https://ftp.gnu.org/gnu/glibc/
+  match: /glibc-[0-9]+(\.[0-9]+)+\.tar\.gz/
+  strip:
+    - /^glibc-/
+    - /\.tar\.gz$/
+
+platforms: linux
+
+build:
+  dependencies:
+    gnu.org/make: '>=3.79'
+    gnu.org/gawk: '>=3'
+    gnu.org/gcc: '*'
+    gnu.org/gettext: '*'
+    gnu.org/texinfo: '*'
+    gnu.org/bison: '*'
+    perl.org: '*'
+    python.org: ^3.13
+  working-directory: build
+  script:
+    - ../configure $ARGS
+    - make all
+    - make install
+
+    - run: for s in $SCRIPTS; do sed -i 's|{{prefix}}|"$(cd "$(dirname "$0")/.." \&\& pwd)"|' $s; done
+      working-directory: ${{prefix}}/bin
+
+    - run: ln -s ../lib/glibc/ld-{{ version.marketing }}.so ld.so
+      working-directory: ${{prefix}}/bin
+  test: make test
+  env:
+    SCRIPTS:
+      - catchsegv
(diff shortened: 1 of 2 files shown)


jhheider:
I probably need to move the binutils/gcc stuff out to peel out the bootstraps.

jhheider:
This passes testing, though I'm far from sure it's actually testing what I think it is. Next step is to make `tea` for linux selectively use it for `tea`-provided binaries *iff* the system's `glibc` is too old. 👀

projects/gnu.org/glibc/package.yml
```diff
@@ -0,0 +1,126 @@
...
+    test "{{hw.platform}}" = "darwin" && exit 0
+    case "{{hw.arch}}" in
+      x86-64) ARCH=x86_64;;
+      aarch64) ARCH=aarch64;;
+      *)
+        echo "Unsupported architecture {{hw.arch}}"
+        exit 1
+        ;;
+    esac
+    gcc \
```
mxcl: `-nostdinc` & `-nostdlib`

mxcl (changes requested):

projects/gnu.org/glibc/package.yml
jhheider: that looks less promising. you sure we want both?

mxcl: it'll stop it using the system headers and libraries… which is surely what we want?

jhheider: Possibly -nodefaultlibs?

https://stackoverflow.com/questions/10119520/g-static-initialization-and-nostdlib

mxcl: Oh ok I see. gcc considers glibc’s headers and libraries the system libraries…?

The goal is for gcc to only use glibc and tea libs, nothing from /usr/lib or /lib or /include or /usr/include etc.

jhheider: which means, I think, providing the init loader from libgcc. we'll see.

felipecrs:
Out of curiosity, any ideas on how big (MB) a bottle of glibc would look like?

mxcl:
this pr has become epic in scope and effort

felipecrs:
no wonder, it basically solves one of the world's biggest problems (i.e allows to run anything on alpine)

PR #150, opened 2023-01-28
+zsh

mxcl:


mxcl:
Should have commented that I just added all the flags from brew’s formula

https://github.com/Homebrew/homebrew-core/blob/master/Formula/zsh.rb

PR #157, opened 2023-01-30
+packaging templates

jhheider:


mxcl:
Yes if there's a way to instantiate them into package ymls. No otherwise. Unless I'm missing context, I'm not sure they add much, but they do add maintenance burden. Actual package.ymls are better for learning.

jhheider:
> Yes if there's a way to instantiate them into package ymls.

It shouldn't be _hard_ to write a tea script that uses charm.sh to grab and substitute the inputs.

> No otherwise. Unless I'm missing context, I'm not sure they add much, but they do add maintenance burden. Actual package.ymls are better for learning.

I agree that there's more options to look at in the wild. I figured this would reduce the learning curve by giving a quick place to start (since the base idioms are pretty static based on language). Happy to shelve if not desirable.

jhheider:
You convinced me (I think). Closing this in favor of [this wip](https://github.com/teaxyz/pantry.core/blob/%2Bnew.sh/scripts/new.sh), that I'm stalling for the moment. I'll make a issue in case anyone wants to work on it.

PR #160, opened 2023-01-31
fish: patch command_not_found_handler

### LINKED ISSUE #344: fish command not found magic gives exit code `127`
fish doesn't return the exit code of tea, it always returns 127. This is a sub par experience for tea users. Can we work around this? If not, should we propose a change to the fish developers?

jhheider:
resolves https://github.com/teaxyz/cli/issues/344


mxcl (approved):

mxcl:
great stuff!

PR #166, opened 2023-02-03
+rx

mfts:
`rx` is an extensible, modern and minimalist pixel editor implemented in rust

### DIFF
--- a/projects/crates.io/rx/package.yml
+++ b/projects/crates.io/rx/package.yml
@@ -1,0 +1,19 @@
+distributable:
+  url: https://github.com/cloudhead/rx/archive/refs/tags/v{{version}}.tar.gz
+  strip-components: 1
+
+provides:
+  - bin/rx
+
+versions:
+  github: cloudhead/rx/tags
+  strip: /v/
+
+build:
+  dependencies:
+    rust-lang.org/cargo: '*'
+  script:
+    cargo install --locked --path . --root {{prefix}}
+
+test:
+  rx -h


projects/crates.io/rx/package.yml
```diff
@@ -0,0 +1,19 @@
...
+  strip: /v/
+
+build:
+  dependencies:
+    rust-lang.org/cargo: '*'
+  script:
+    cargo install --locked --path . --root {{prefix}}
+
+test:
+  rx -h
```
jhheider, line 19: Can we add a better test? -h doesn't always prove the binary will run.

jhheider: I tried
```yaml
test:
  dependencies:
    curl.se: '*'
  script: |
    rx -h
    curl http://www.libpng.org/pub/png/PngSuite/basn6a08.png >test.png
    rx -u $FIXTURE test.png
  fixture: |
    e test.png
    p/clear
    p/add #ff0000
    w test2.png
    q
```
but it a) opens a window, and b) won't quit from init. So, I think we might be stuck.

jhheider:
For some reason I merged this, even though it doesn't build (why?).

Requires X11 libraries. Those could be packaged, though it's going to be a bear. But feel free to try it, if you want, @mfts .

mfts:
I was already wondering why it got merged so quickly. Yes, X11 is gonna be a big one. I'll give it a try.

PR #169, opened 2023-02-04
+yarnpkg.com

getneil:


projects/yarnpkg.com/package.yml
```diff
@@ -0,0 +1,39 @@
...
+    yarnScript=$(cat scripts/bin/yarn)
+    replace="../.."
+    next=..
+    echo "${yarnScript/$replace/$next}" >> {{prefix}}/bin/yarn
+    chmod +x {{prefix}}/bin/yarn
+
+    cp {{prefix}}/bin/yarn {{prefix}}/bin/yarnpkg
+
+    if test "{{hw.platform}}" = "linux"; then
+      $SUDO apt-get --yes install libatomic1
```
getneil: need help on this im trying to find the first source or how to package this.

jhheider: it's _very_ low level. I believe we require it as a matter of course, so it shouldn't be necessary to install it.

jhheider:
In theory, you might be able to install this using `npm install . --global --prefix={{prefix}}`.

getneil:
> In theory, you might be able to install this using `npm install . --global --prefix={{prefix}}`.



> In theory, you might be able to install this using `npm install . --global --prefix={{prefix}}`.

that should have been the first thing i tried.

jhheider:
It's not always obvious; I referred back to my work on `npmjs.com`. That's something that would be well-served by https://github.com/teaxyz/pantry.core/issues/142.

getneil:
> npm install . --global --prefix={{prefix}}

unfortunately it still requires the libatomic1.

jhheider:
That should be on all the images, unless something has changed. Almost everything requires it.

getneil:
> That should be on all the images, unless something has changed. Almost everything requires it.

working for me now. i replaced my docker image w/ `ghcr.io/teaxyz/infuser`

jhheider:
#teamwork

jhheider:
and, I think, that might be on the top300.

PR #180, opened 2023-02-06
+gnutls+p11-kit+libtasn1+libunistring+nettle

jhheider:
(gnutls requires too many things; skipped guile because it looks awful; does anyone use guile?)

ref https://github.com/teaxyz/pantry.extra/issues/99


jhheider:
wow, I can't believe that worked.

PR #183, opened 2023-02-07
fix(brotli)

### LINKED ISSUE #372: I get `bad-yaml` error for real packages that I know to be in the pantry
These packages have not been previously installed. Both work correctly with the long-form `tea +pkg pkg`
```sh
$ sl
error: Uncaught (in promise) Error: bad-yaml
    if (!isArray(node)) throw new Error("bad-yaml");
                              ^
    at Object.getProvides (file:///opt/tea.xyz/src/v0.22.3/src/hooks/usePantry.ts:133:31)
```

```sh
$ xc system
error: Uncaught (in promise) Error: bad-yaml
    if (!isArray(node)) throw new Error("bad-yaml");
                              ^
    at Object.getProvides (file:///opt/tea.xyz/src/v0.22.3/src/hooks/usePantry.ts:133:31)
```

tea v0.22.3

jhheider:
fixes https://github.com/teaxyz/cli/issues/372

### DIFF
--- a/projects/github.com/google/brotli/package.yml
+++ b/projects/github.com/google/brotli/package.yml
@@ -17,7 +17,7 @@ build:
     make install
 
 provides:
-  bin/brotli
+  - bin/brotli
 
 test:
   fixture:


jhheider:
This is a fortran-grade single-hyphen bug.

mxcl:
My bad. At v1 we can have a YAML spec checking thingy.

jhheider:
Yeah, I considered that option, but not while we're still evolving it.

PR #184, opened 2023-02-09
+imagemagick

jhheider:


projects/imagemagick.org/package.yml
```diff
@@ -0,0 +1,53 @@
+distributable:
+  #FIXME: they use a weird versioning scheme
+  url: https://github.com/ImageMagick/ImageMagick/archive/refs/tags/7.1.0-61.tar.gz
+  strip-components: 1
+
+versions:
+  #FIXME: they use a weird versioning scheme
+  # github: ImageMagick/ImageMagick
+  - 7.1.0.61
```
jhheider, line 9: @mxcl you ok with this for now? They use the pre-release identifier as a super-patch, or something.

mxcl: Fine for now, we will need to figure it out one day tho.

projects/imagemagick.org/package.yml
```diff
@@ -0,0 +1,53 @@
...
+      --prefix={{ prefix }} \
+      --disable-silent-rules
+    make --jobs {{ hw.concurrency }} install
+
+test:
+  dependencies:
+    curl.se: '*'
+  script: |
+    curl "https://upload.wikimedia.org/wikipedia/commons/6/6a/PNG_Test.png" -o a.png
+    identify a.png | grep "a.png PNG 3120x3920 3120x3920+0+0 8-bit sRGB 7.02443MiB 0.000u"
```
mxcl, line 53: newlines at the end of files is the convention. Let’s be consistent and do that. Not the end of the world if we forget, but it's a worthwhile goal.

projects/imagemagick.org/package.yml
```diff
@@ -0,0 +1,53 @@
...
+
+build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+    freedesktop.org/pkg-config: '*'
+  script: |-
+    # seemingly `./configure` doesn't use pkg-config for libjpeg
+    export CPPFLAGS=`pkg-config --cflags libjpeg`
+    export LDFLAGS=`pkg-config --libs libjpeg`
```
mxcl, line 41: Could be `env`

mxcl (approved):

projects/imagemagick.org/package.yml
jhheider: I was thinking about that. Back ticks should interpret at the right time, right?

PR #185, opened 2023-02-09
+github.com/clever/microplane

johnallen3d:
TODO:

- [ ] Add a better test

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,8 +74,11 @@ with their pull request then you can use GitHub’s CLI:
 ```
 $ gh pr checkout 123
 
-# or you can copy paste the URL
+# or you can copy paste the URL:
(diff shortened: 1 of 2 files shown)


jhheider:
> * [ ]  Add a better test

Hm. The docs suggest it needs a github token to be useful. Maybe something like this:

```yaml
test:
  dependencies:
    cli.github.com: '*'
  scripts: |
    export GITHUB_TOKEN=$(gh auth login) ## dunno if this will Just Work™
    mp init "org:teaxyz"
    mp clone
    mp status
```

johnallen3d:
Thanks @jhheider, I'll give this a try.

mxcl:
The problem here I believe is that the GITHUB_TOKEN isn't scoped to clone anything. Or something like that.

What's the solution :/

PR #187, opened 2023-02-10
+vim.org

### LINKED ISSUE #188: +vim.org
related to Top 300

getneil:
resolves #188

### DIFF
--- a/projects/vim.org/package.yml
+++ b/projects/vim.org/package.yml
@@ -1,0 +1,45 @@
+distributable:
+  url: https://github.com/vim/vim/archive/refs/tags/v{{version}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: vim/vim/tags
+  strip: /v/
+
+
+provides:
+  - bin/vim
+  - bin/vi
+
+dependencies:
+  python.org: '>=3.11'
+
+build:
+  dependencies:
+    cmake.org: ^3
+  script: |
+    ./configure \
+      --prefix={{prefix}} \
+      --enable-multibyte \
+      --with-tlib=ncurses \
+      --with-compiledby=tea.xyz \
+      --enable-cscope \
+      --enable-terminal \
+      --enable-perlinterp \
+      --enable-rubyinterp \
+      --enable-python3interp \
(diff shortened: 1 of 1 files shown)


getneil:
the error doesnt seem helpful atm, but this builds and test correctly in my local.

Screenshot 2023-02-10 at 4 17 22 PM

issue seems to be useGithubApi related, graphql query. ...investigating

jhheider:
Given that there's 15,000 tags, and the API returns them oldest first, it might just be too many. I can't easily see what the oldest tags look like, but it might be that the API limits out before it finds one it can read. This *looks* like it should scan them all, though: https://github.com/teaxyz/cli/blob/12f482cf8ffaa99f646317513ca1230e113f9391/src/hooks/useGitHubAPI.ts#L50-L91

getneil:
I see. It might be better to manually specify versions right?

projects/vim.org/package.yml
```diff
@@ -0,0 +1,44 @@
...
+      --enable-luainterp
+    make
+    make install prefix={{prefix}}
+    if test "{{hw.platform}}+{{hw.arch}}" = "linux+x86-64"; then
+      cd {{prefix}}/bin
+      ln -s vim vi
+    fi
+test:
+  script: |
+    vim --version
```
getneil: original intended test:
```
test:
  script: |
    vim --version
    vi --version
    echo "hello world" >> ./test.txt
    echo ":s/hello/test/" >> ./commands.vim
    echo ":wq" >> ./commands.vim
    vim -T dumb -S ./commands.vim ./test.txt
    test "$(cat ./test.txt)" = "test world"
```

still figuring out but it freezes on `vim -T dum ...`

mxcl: probs waiting on stdin input even though I assume that's what `-T dumb` is meant to prevent…?

jhheider: I suspect `-T dumb` is the terminal to use; I found great difficulty testing ncurses applications even with the terminal dummied out.

jhheider:
Beautiful teamwork, everyone!

PR #388, opened 2023-02-13
+github.com/mamba-org/mamba

### LINKED ISSUE #6: +conda
Many who work with Python use conda. Example, OpenBB uses pip and conda

ddobrinskiy:
# Intro

[mamba](https://mamba.readthedocs.io/en/latest/installation.html) is a FAST package manager for data science

While mainly aimed at python, it supports most scientific computing languages: julia, R, python, etc

It is now the default tool for most data scientists to manage their environments.

Mamba has now de-facto replaced conda, because:

- it mimics the API
- conda is very slow, while mamba is not

## Motivation

As a Data Scientist I often have to take my code written locally with mamba as the pkg manager, and run it in CI/prod.

Getting the dependencies and docker images correct is a pain, especially since most mamba users are data scientists, not developers.

## No dependencies

mamba's core idea is to be 100% self-contained distribution for python/julia/R/etc package management

So having external dependencies goes against how it is structured, we need to build it using only the `.sh` distribution files from mamba-forge

By packaging mamba with `tea`, I am hoping to simplify my workflows, and maybe even get wider adoption of `tea` within the Data Science community

## p.s.

resolves https://github.com/teaxyz/pantry.extra/issues/6


ddobrinskiy:
Current blocker: can't build locally

```bash
❯ ./pantry.core/scripts/bottle.ts github.com
error: Uncaught (in promise) Error: missing GPG_PASSPHRASE
  throw new Error(message)
        ^
    at panic (https://raw.githubusercontent.com/teaxyz/cli/v0.23/src/utils/safe-utils.ts:8:9)
    at file:///home/david/proj/tea_stuff/pantry.core/scripts/bottle.ts:39:59
```

jhheider:
If this is a python-venv, you might want to use [python-venv.sh](https://github.com/teaxyz/brewkit/blob/main/share/brewkit/python-venv.sh); see https://github.com/teaxyz/pantry.extra/blob/main/projects/youtube-dl.org/package.yml. In any case, it looks like we'll need `python.org: ^3.something` as a runtime dependency.

ddobrinskiy:
@jhheider mamba's core idea is to be 100% self-contained distribution for python/julia/R/etc package management

Having external dependencies goes against how it is structured.

So I'll do more experiments on how to build it properly.

p.s. I also updated the PR description to make the purpose of this PR clearer

ddobrinskiy:
Current status: builds and works locally

```bash
# add mamba and it's bin folder to shell
❯ ~/proj/pkgdev/pantry.extra/tea.out/github.com/mamba-org/mamba/v22.11.1.4/bin/mamba init zsh
modified      /home/david/.zshrc

==> For changes to take effect, close and re-open your current shell. <==

Added mamba to /home/david/.zshrc

❯ mamba --version
mamba 1.1.0
conda 22.11.1

❯ echo $CONDA_PREFIX
/home/david/proj/pkgdev/pantry.extra/tea.out/github.com/mamba-org/mamba/v22.11.1.4

❯ which python
/home/david/proj/pkgdev/pantry.extra/tea.out/github.com/mamba-org/mamba/v22.11.1.4/bin/python
```

ddobrinskiy:
Current status: local tests work, CLI tests fail with

```bash
+ mamba --version
/home/runner/.tea/github.com/mamba-org/mamba/test-v22.11.1.4/xyz.tea.test.sh: /home/runner/.tea/github.com/mamba-org/mamba/v22.11.1.4/bin/mamba: /opt/github.com/mamba-org/mamba/v22.11.1.4/bin/python: bad interpreter: No such file or directory
```

This is because `mamba` file is actually just a bash script that links to mamba-built dependencies

And when we build `mamba.sh`, the `mamba` file is **created** with hard-coded (?) links to `bin` location

CI builds it's dependencies in `/opt/github.com/mamba-org/mamba/v22.11.1.4/bin/`, so that's where `mamba` file is looking for its  `python`

While at test-time, the artifacts are stored in `/home/runner/.tea/github.com/mamba-org/mamba/v22.11.1.4/bin/mamba`, and running it tries to call python from `/opt/...` and thus fails.

Not sure what's the correct fix here, need to dig deeper into how mamba looks for the relevant `bin`:
* is it hardcoded into `mamba` at build?
* or is it dependent on some env variable that we can spoof at test-time?

Upd: bad news, `bin/mamba` is actualy a python script with a hard-coded shebang generated at build time



Maybe an appropriate solution is to redirect the original `/opt...` directory with a symlink at test time?

mxcl:
This is awesome! Thank you.

Use our `fix-shebangs.ts` script. See other package.ymls for usage examples.

projects/github.com/mamba-org/mamba/package.yml
```diff
@@ -23,14 +23,22 @@ build:
...
 
+    sleep 1
     pwd
+    sleep 1
     ls
 
     # install mamba
     chmod +x mamba.sh
     ./mamba.sh $ARGS #FIXME when running on gh actions, this fails
                      # ^ /opt/github.com/mamba-org/mamba/v22.11.1.4/bin/python: bad interpreter: No such file or directory
```
mxcl: does it need python as a *build* dependency?

ddobrinskiy: No, it comes with it's own instance of python

This particular piece of code has been updated and no longer has a #FIXME

ddobrinskiy:
Current status:

* DONE:
  * [x] passes CI
  * [x] builds & tests ok locally
  * [x] did some testing of tea-packaged mamba/python in action locally
  * [x] move from `curl` to `aria2c` to simplify code and local testing (aria2c supports download caching, could not get it to work with curl)

* TBD
  * [x] extensive testing of mamba/python locally (on linux)
  * [x] test locally on macos m1
  * [x] clean up code

@mxcl FYI

ddobrinskiy:
**FIXME**: 

`fix-shebangs.ts` breaks local installation, investigating

```bash
❯ mamba --version
/usr/bin/env: ‘python’: No such file or directory
```

ddobrinskiy:
**FIXME**: 

`fix-shebangs.ts` breaks local installation, investigating

```bash
❯ mamba --version
/usr/bin/env: ‘python’: No such file or directory
```

So there is a catch 22 here:

after installation of mamba, the user should run `mamba init "$(basename "${SHELL}")"` to add it to PATH et cetera.

however to run this for the first time, python is needed.

Python is already part of mamba's package bin folder, but at first installation it's not in PATH yet. And we can't add it to path without using some version of python for  initial run of `mamba init`.


Honestly, this error confuses me in several ways
* shouldn't `tea` be invoked when my terminal could not find a working python instance?
* why is `pantry.extra/tea.out/github.com/mamba-org/mamba/v22.11.1.4/condabin` in my path? Does it appear during build? why not `mamba/v22/bin` then, but `condabin`?
  * should try this on a clean linux install
* if the reader ever feels the compulsion to give up add `python` as dependency, **DON'T**:  this would _technically_ solve a lot of problems with build and setup, but I'm reluctant because this goes against the philosophy of mamba/conda, i.e. this being a self-sufficient and alternative distribution of python

ddobrinskiy:
UPD: this whole comment is no longer relevant, see https://github.com/teaxyz/pantry.extra/pull/388#issuecomment-1434675630

@mxcl This PR is ready for review.

There is one thing that bothers me though:

After installing mamba, the user is expected to run `mamba init "$(basename "${SHELL}")"` to add it to their shell properly.

First, remember that `mamba` is not a binary - it is a python script.

So for this to work, it is very important that the version of `python` that `mamba` script finds is using exactly the python distribution was packaged with it. By default, this is solved by hard-coded shebangs that the installer autogenerates at installation.

However this breaks tea tests. So we did `fix-shebangs.ts`, but now `mamba` does not have hardcoded shebangs, and so when running it after a fresh install, it does not see the very specific python distribution from above.

Currently I fix it by modifying PATH (see below), and after that everything just works without a hitch.

But this bothers me, and should be somehow fixed in the future - I will experiment more after this PR is merged (I want to see how tea behaves when I'm installing mamba via running the magic)

```bash
❯ export PATH="$HOME/proj/pkgdev/pantry.extra/tea.out/github.com/mamba-org/mamba/v22.11/bin/python:$PATH"
❯ mamba init zsh

```

ddobrinskiy:
Second thing that bothers me is that the code downloads installation `.sh` script at build time (different files depending on platform), and it does not depend on the `distributable.url`

So another thing to ideally fix in the future

ddobrinskiy:
Upd: The correct way to init mamba for daily usage is the following:

```bash
tea +github.com/mamba-org/mamba sh          # activate a tea venv with mamba & conda-packaged-python
tea $ mamba init "$(basename "${SHELL}")"   # updates your .zshrc or .bashrc to make mamba & conda usable
```

Also added this to source code as a comment

jhheider:
Does that produce a different outcome from just running `mamba init $shell_name`? The tea environment should be present during the invocation.

ddobrinskiy:
> Does that produce a different outcome from just running `mamba init $shell_name`? The tea environment should be present during the invocation.

@jhheider Yeah, this actually seems to work. I simplified the comment sections in the code.

p.s. the behavior below is exactly what I would want as a data scientist. Especially if I was less experienced with paths et cetera, I would want things to "just work"

```
❯ pwd
/home/david/proj/pkgdev/pantry.extra
❯ tea build github.com/mamba-org/mamba
❯ mamba init zsh # changes ~/.zshrc
❯ which python
python not found
❯ source ~/.zshrc
❯ which python
/home/david/proj/pkgdev/pantry.extra/tea.out/github.com/mamba-org/mamba/v22.11.1.4/bin/python
```

ddobrinskiy:
If I understand tea cli correctly so was, [this whole installation page](https://mamba.readthedocs.io/en/latest/installation.html) with it's various instructions can be replaced by a tea one-liner:

` sh <(curl tea.xyz) mamba init "$(basename "${SHELL}")"`

mxcl:
> If I understand tea cli correctly so was, [this whole installation page](https://mamba.readthedocs.io/en/latest/installation.html) with it's various instructions can be replaced by a tea one-liner:
> 
> ` sh <(curl tea.xyz) mamba init "$(basename "${SHELL}")"`

Awesome :) The dream is coming true.

projects/github.com/mamba-org/mamba/package.yml
```diff
@@ -0,0 +1,54 @@
...
+    # download mamba installation script
+    aria2c -c -o mamba.sh "https://github.com/conda-forge/miniforge/releases/download/22.11.1-4/Mambaforge-22.11.1-4-$PLATFORM.sh"
+
+    # install mamba
+    chmod +x mamba.sh
+    ./mamba.sh $ARGS
+
+    fix-shebangs.ts {{prefix}}/bin/*
+
+    #FIXME: add caveats
```
mxcl, line 31: yeah it's on my “TODO” to figure out how we can communicate this sort of thing.

mxcl (approved):
Fabulous contribution.

kendu3310:
nice

PR #390, opened 2023-02-14
+digip.org/jansson

lino-levan:
This is blocked by https://github.com/teaxyz/pantry.core/issues/18 afaik

### DIFF
--- a/projects/digip.org/jansson/fixture.c
+++ b/projects/digip.org/jansson/fixture.c
@@ -1,0 +1,11 @@
+#include <jansson.h>
+#include <assert.h>
+int main()
+{
+  json_t *json;
+  json_error_t error;
+  json = json_loads("\"foo\"", JSON_DECODE_ANY, &error);
+  assert(json && json_is_string(json));
+  json_decref(json);
+  return 0;
+}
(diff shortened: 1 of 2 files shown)


lino-levan:
I may just be stupid and nothing is blocking this /shrug. Seems that the build is failing on linux though which is a bit unlucky.

jhheider:
Looks like the build is passing an unknown flag to llvm's linker, `--default-symver`. So, if it's known to build with llvm, we need to figure out how to stop that flag from passing. Worst case, it might mean monkey-patching some of `configure` or `Makefile*`

mxcl:
brew’s superenv would know which toolchain was being used and strip out invalid flags.

We could do the same in our `tea.xyz/gx/cc` proxy package.

Notably I remember writing “superenv” and it was a months long trial. Fixing things broke other things etc.

Another option is compiling with `gcc` which presumably this package expects…?

jhheider:
I wonder if `tea.xyz/gx/cc` needs some special logic to allow using it to shim system vs. llvm.org vs. gnu.org/gcc vs. something in the future. Possibly by using the `value` in the yaml. It would have solved some past issues, but, as you noted, getting it right is massive trial-and-error.

mxcl:
GHA checks out the repo the action is in, you don’t have to do it.

Though the path is `github.action_path` I believe.

Good work here.

PR #395, opened 2023-02-14
+libxslt

mfts:
chatgpt said it was a requirement for postgres, so i built it. turns out it's not a requirement. but the build process was easy (similar to libxml2)

### DIFF
--- a/projects/gnome.org/libxslt/package.yml
+++ b/projects/gnome.org/libxslt/package.yml
@@ -27,8 +27,10 @@ test:
     tea.xyz/gx/cc: c99
     gnome.org/libxml2: '*'
   script: |
-    cc `xslt-config --cflags --libs` test.c
+    cc `xslt-config --cflags --libs` test.c $CFLAGS
     ./a.out
+  env:
+    CFLAGS: -I{{pkgs.gnome.org/libxml2.prefix}}/include
 provides:
   - bin/xslt-config
   - bin/xsltproc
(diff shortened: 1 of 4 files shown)


mfts:
"Insanity is doing the same thing over and over and expecting different results." 

will continue later on debugging linux test

jhheider:
Nested headers are annoying. Some tools are resiliant, and some are not. See e.g. https://github.com/teaxyz/pantry.core/blob/adaaf3af9417051f02a211b1c3d25c5c0c1314e9/projects/python.org/package.yml#L107-L108. Although, often, simply adding `pkgconfig` as a build dep is sufficient.

projects/gnome.org/libxslt/package.yml
```diff
@@ -0,0 +1,36 @@
...
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+  script: |
+    ./configure $ARGS
+    make --jobs {{ hw.concurrency }}
+    make install
+  env:
+    ARGS:
+      - --prefix={{prefix}}
+      - --without-python  # mysterious build failure in GHA macOS-11
```
mxcl: the comment is copy pasted from libxml2. Does it apply? Or did you just YOLO paste?

mfts: Applies here as well. Without it it uses Python and crashes the build

mfts: But yes I checked the `libxml2` build first because the docs say the build instructions are very similar.

mxcl: k good stuff ☺️

mfts:
@mxcl how is `xslt-config.patch.in` used? Is that file automatically called by `xslt-config`?

mxcl:
@mfts you can see the patch being applied on the first line of the build script

mfts:
@mxcl I only looked at the `test`. Learned something new 👍

PR #396, opened 2023-02-15
+crates.io/exa

penguwin:
Adds [exa](https://github.com/ogham/exa) a modern replacement for `ls`.

### DIFF
--- a/projects/crates.io/exa/package.yml
+++ b/projects/crates.io/exa/package.yml
@@ -1,0 +1,25 @@
+distributable:
+  url: https://github.com/ogham/exa/archive/refs/tags/v{{version}}.tar.gz
+  strip-components: 1
+
+provides:
+  - bin/exa
+
+versions:
+  github: ogham/exa/tags
+  strip: /^v/
+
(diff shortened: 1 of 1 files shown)


projects/crates.io/exa/package.yml
```diff
@@ -0,0 +1,25 @@
...
+provides:
+  - bin/exa
+
+versions:
+  github: ogham/exa/tags
+  strip: /^v/
+
+dependencies:
+  darwin:
+    zlib.net: ^1
```
mxcl, line 14: Is this really only required on mac? If so could use a comment explaining it.

penguwin: It seems to be required, without `zlib` the CI fails. After further investigation, it seems that I also need `pandoc` at build time for additional functionality, as well as `libgit2` on linux. I'll take a closer look

mxcl:
Merged because having it available is more useful than it being perfect. OSS is about getting the person for whom this doesn't quite work to submit the next PR. 👊🏻

PR #398, opened 2023-02-15
+postgresql.org

mfts:
add postgresql.org

### DIFF
--- a/projects/postgresql.org/package.yml
+++ b/projects/postgresql.org/package.yml
@@ -75,15 +75,15 @@ test:
     mkdir -p ./data
 
     if test "{{ hw.platform }}" = "linux"; then
-      INITDB="$(command -v initdb)"
-      chmod 777 data
-      chmod 777 .
-      sudo -u nobody $INITDB -D $(pwd)/data
+      chown nobody. data
+      chmod 755 .
+      sudo -u nobody {{prefix}}/bin/initdb -D $(pwd)/data
+      sudo -u {{prefix}}/bin/pg_ctl -D ./data -l logfile start
     else
       initdb -D ./data
+      pg_ctl -D ./data -l logfile start
     fi
 
-    pg_ctl -D ./data -l logfile start
     createdb test
     psql -c 'create table test (id int);' test
     dropdb test
--- a/projects/postgresql.org/package.yml
+++ b/projects/postgresql.org/package.yml
@@ -75,7 +75,7 @@ test:
     mkdir -p ./data
 
     if test "{{ hw.platform }}" = "linux"; then
-      $INITDB="$(command -v initdb)"
+      INITDB="$(command -v initdb)"
       chmod 777 data
       chmod 777 .
       sudo -u nobody $INITDB -D $(pwd)/data
--- a/projects/postgresql.org/package.yml
+++ b/projects/postgresql.org/package.yml
@@ -75,7 +75,7 @@ test:
     mkdir -p ./data
 
     if test "{{ hw.platform }}" = "linux" && command -v sudo; then
-      SUDO="sudo -E -u nobody"
+      SUDO="sudo -E -u nobody PATH=$PATH"
 
       $SUDO initdb -U nobody -D ./data
       $SUDO pg_ctl -D ./data -l logfile start
--- a/projects/postgresql.org/package.yml
+++ b/projects/postgresql.org/package.yml
@@ -72,16 +72,23 @@ provides:
 
 test:
   script: |
-    mkdir -p ./data
+    apt-get install -y sudo
+    cat >test.sh <<EOF
+    #!/bin/sh
+
+    $(env)
+
+    $(cat $FIXTURE)
+    EOF
 
     if test "{{ hw.platform }}" = "linux"; then
-      su nobody -c "initdb -D ./data"
+      sudo -u nobody bash -x test.sh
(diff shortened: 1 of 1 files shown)


projects/postgresql.org/package.yml
```diff
@@ -0,0 +1,82 @@
...
+    postgres -V
+
+    # this won't work because it requires 'su' priviledges
+    # mkdir -p ./data
+    # initdb -D ./data
+    # pg_ctl -D ./data -l logfile start
+    # createdb test
+    # psql -c 'create table test (id int);' test
+    # dropdb test
+    # pg_ctl -D ./data stop
```
mfts: This actually works in darwin, but fails in linux. Not sure how I can write a better test that doesn't requires sudo priviledges.

jhheider: What's the specific failure? Can you work around it by altering where the db, sock, or which port is used? I know root privileges are generally required for ports under 1024 (though psql is usually 5432), and for writing stuff like `/var/run/postgres.sock`.

mfts: The error locally is 

```
initdb: error: cannot be run as root
initdb: hint: Please log in (using, e.g., "su") as the (unprivileged) user that will own the server process.
```

jhheider: Oh, wow. It won't run as root. Hm. Does `sudo -u nobody 'initdb -D ./data'` work?

jhheider: If you need to, you could write the script as a fixture, and just do:
```sh
test "{{ hw.platform }}" = "linux" && SUDO="sudo -u nobody"
$SUDO sh $FIXTURE
```
which feels really bad, so definitely add a comment.

mfts: I tried that as well. 

```
+ su -u nobody -c 'initdb -D ./data'
Try 'su --help' for more information.
```

jhheider: sudo -p preserves environment, so `SUDO="sudo -p -u nobody"`, maybe.

jhheider: All really annoying, though.

mfts: Will look at it later again. sometimes it helps to step away from keyboard :)

jhheider: s_sometimes_pretty much always_

projects/postgresql.org/package.yml
```diff
@@ -0,0 +1,97 @@
...
+test:
+  script: |
+    mkdir -p ./data
+
+    if test "{{ hw.platform }}" = "linux"; then
+      adduser postgres --disabled-password --gecos ""
+      chown -R postgres:postgres ./data
+      PASSWD=$(date | md5sum | cut -c1-8)
+      echo "postgres:$PASSWD" | chpasswd
+      su postgres -c 'initdb -D ./data'
```
mfts: hmm the ci doesn't like this one either... in local docker it works. and `su nobody -c 'initdb -D ./data'` doesn't work. maybe it works in the ci?

jhheider: I think you want `sudo -u nobody initdb -D ./data`. The runners have pw-less sudo, much like GHA runners.

mfts: That's good to know. 

I cannot really debug this locally because locally the linux container works differently and `sudo: command not found` error is thrown. 

In GH CI it says that `initdb` cannot be found
https://github.com/teaxyz/pantry.extra/actions/runs/4250986462/jobs/7392829427#step:8:120

jhheider: Oops, right, try `-p` to sudo as well to preserve environment.

jhheider:
Initdb (maybe?) seems to be trying to run adduser. We should prevent that if possible.

projects/postgresql.org/package.yml
```diff
@@ -0,0 +1,93 @@
...
+    fi
+
+    if test "{{ hw.platform }}" = "darwin"; then
+      initdb -D ./data
+      pg_ctl -D ./data -l logfile start
+      createdb test
+      psql -c 'create table test (id int);' test
+      dropdb test
+      pg_ctl -D ./data stop
+    fi
```
jhheider: This might be simpler:
```sh
  if test "{{ hw.platform }}" = "linux"; then
    SUDO="sudo -u -p nobody"
  fi

  $SUDO initdb -U nobody -D ./data
  $SUDO pg_ctl -D ./data -l logfile start
  $SUDO createdb test
  $SUDO psql -c 'create table test (id int);' test
  $SUDO dropdb test
  $SUDO pg_ctl -D ./data stop
```

mfts:
All of the sudden the builds are failing on Linux. That wasn't the case before.

jhheider:
Looks like a missing dep:
```sh
***
ERROR: `flex' is missing on your system. It is needed to create the
file `scan.c'. You can either get flex from a GNU mirror site
or download an official distribution of PostgreSQL, which contains
pre-packaged flex output.
***
```

Dunno why it was passing before. I don't like _that_.

mfts:
strange because on #447 `flex` is reported missing although it's listed as a dep

jhheider:
punted the test. it's _de minimis_ on linux. no need to hold this up longer. thanks, @mfts .

PR #403, opened 2023-02-16
Update wumpus to 1.9

magnusviri:


magnusviri:
This isn't building from the source. I've submitted an [issue](https://gitlab.com/esr/wumpus/-/issues/1).

mxcl:
Looks like we need `xmlto` or to skip building the docs https://github.com/teaxyz/pantry.extra/actions/runs/4196952855/jobs/7278644336#step:8:94

magnusviri:
I've started work on xmlto. I don't know when I'll finish.

jhheider:
You should be able to skip the manpage (for now) with:
```sh
make wumpus superhack
mkdir -p "{{prefix}}"/bin
mv wumpus superhack "{{prefix}}"/bin
```

magnusviri:
Looks like it works now. I added a test too. I'm not sure why the processes die in the test, I was going to kill them, but oh well.

PR #404, opened 2023-02-17
+github.com/create-dmg/create-dmg

getneil:


jhheider:
https://github.com/teaxyz/pantry.extra/blob/44f830645202d067023bdd155f1a4914aaa5aa04/projects/freedesktop.org/libbsd/package.yml#L16-L32

getneil:
> https://github.com/teaxyz/pantry.extra/blob/44f830645202d067023bdd155f1a4914aaa5aa04/projects/freedesktop.org/libbsd/package.yml#L16-L32

thanks that worked. there might be a bug in linux i needed to do `mkdir -p {{prefix}}`

also the package doesnt have a configure step its just a bash script .

jhheider:
> also the package doesnt have a configure step its just a bash script .

Looks like it has a `Makefile`, but it's not doing a lot. You might be able to simplify with `make prefix={{prefix}} install`.

getneil:
> > also the package doesnt have a configure step its just a bash script .
> 
> Looks like it has a `Makefile`, but it's not doing a lot. You might be able to simplify with `make prefix={{prefix}} install`.

that worked!

PR #412, opened 2023-02-18
Fix qrencode

### LINKED ISSUE #409: qrencode not working
qrencode is not working as expected: it does not generate an image

```bash
❯ qrencode 123456789 -o test.png
PNG output is disabled at compile time. No output generated.
```

Notably, while it seems to pass the CI tests, it actually fails in the same way: https://github.com/teaxyz/pantry.extra/actions/runs/4179109052/jobs/7239159814#step:8:345 

initial PR: https://github.com/teaxyz/pantry.extra/pull/393

davdroman:
Here's an attempt to fix qrencode based on imagemagick's own fix https://github.com/teaxyz/pantry.extra/blob/16ba42c8b4fd7ce76ede36dfb8ec46de0c9d025e/projects/imagemagick.org/package.yml#L40-L41

Hopefully closes #409

### DIFF
--- a/projects/fukuchi.org/qrencode/package.yml
+++ b/projects/fukuchi.org/qrencode/package.yml
@@ -29,7 +29,7 @@ build:
   env:
     ARGS:
       - --prefix="{{prefix}}"
-    CPPFLAGS: $(pkg-config --cflags libpng)
+    CFLAGS: $(pkg-config --cflags libpng)
     LDFLAGS: $(pkg-config --libs libpng)
 
 test: |
--- a/projects/fukuchi.org/qrencode/package.yml
+++ b/projects/fukuchi.org/qrencode/package.yml
@@ -29,7 +29,6 @@ build:
   env:
     ARGS:
       - --prefix={{prefix}}
-      - --libdir={{prefix}}/lib
     CFLAGS: $(pkg-config --cflags libpng)
     LDFLAGS: $(pkg-config --libs libpng)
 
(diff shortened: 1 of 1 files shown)


davdroman:
~~Hmmm... doesn't seem to work for me locally, same PNG notice as in #409. Any ideas @mxcl @jhheider?~~

I figured it out! `--jobs {{ hw.concurrency }}` was causing the issue. That along with the C flags fixed it.

jhheider:
Weird. Concurrent make jobs shouldn't cause build issues.

mxcl:
> Concurrent make jobs shouldn't cause build issues.

I't more common that you'd think, especially with makefiles. Which is modern build systems are designed to abstract concurrency so our stupid meat brains don't have to think about it.

mxcl:
I note brew doesn't “deparallize” (brew adds `--jobs x` automatically)

https://github.com/Homebrew/homebrew-core/blob/24b557524b975cd7c5f458c147a1abcc9f511acf/Formula/qrencode.rb

mxcl:
we also need to fix the test, presumably by verifying the png is created and valid:

https://github.com/teaxyz/pantry.extra/actions/runs/4179109052/jobs/7239159814#step:8:345

davdroman:
> we also need to fix the test

@mxcl done [28cdc7e](https://github.com/teaxyz/pantry.extra/pull/412/commits/28cdc7eb7ed9d690185725040aef8556d3c1c789)

mxcl:
Tremendous test

PR #416, opened 2023-02-20
+pcre2

mfts:


mxcl:
k so, as soon as we need this we should merge though I would rather keep a single `package.yml` though that will require some work at the `tea/cli` layer.

Also should be `projects/pcre.org/2/package.yml`

mxcl:
Yeah this is almost identical to the other PCRE so for now what I'm going to do is a hack.

mxcl:
I decided that our name spacing system easily supports multiple package yamls and other solutions require more thought or perhaps are just too complex and not worth it.

PR #421, opened 2023-02-22
fix(solidity)

jhheider:
woo, my PR made it in!


mxcl:
us the new king of OSS

PR #425, opened 2023-02-25
Add golangci-lint

regrowr:


mxcl:
We prefer not to run installers, but we’re being lax about this for now.

regrowr:
> We prefer not to run installers, but we’re being lax about this for now.

Ok... that would be my hermit binary install mindset, but also golanci-lint themselves [recommend](https://golangci-lint.run/usage/install/#install-from-source) a binary install.

That said, I'll circle back and look at addressing their issues in the context of tea... soon-ish.

mxcl:
I think we're ok taking binaries, especially if the project says we should.

However eventually we'll require they be signed, or the user will need to opt in.

Essentially our users trust us to vet the binaries. So we build them ourselves and then say: look we took the sources the project provided. You trust that we didn't modify them and we built it.

Admittedly there's opportunity for malicious intent in the sources so I dunno really how much we're buying.

PR #426, opened 2023-02-25
Add github.com/evilmartians/lefthook

regrowr:


mxcl:
Doesn't use sources, but I’m not sure we're fussed. At some point I think we will require binaries that we don’t build to be signed. But that's not yet.

regrowr:
> Doesn't use sources, but I’m not sure we're fussed. At some point I think we will require binaries that we don’t build to be signed. But that's not yet.

That would be my hermit binary install mindset... again.

I'll circle back and look at a source build... soon-ish.

PR #430, opened 2023-02-28
+graphviz.org

jonchang:
Only includes "recommended" dependencies as listed on their website: https://graphviz.org/download/source/#dependencies


jonchang:
I think this may need poppler. The documentation for graphviz suggests pangocairo is sufficient for PDF output, but it seems this is not the case.

jonchang:
Same `flex` issue as https://github.com/teaxyz/pantry.extra/pull/398. Going to try to debug this on Docker later since I can't repro this on my x86-64 linux box.

mxcl:
wonderful

PR #435, opened 2023-03-01
libxml2: make it so most includes just work

mxcl:


mxcl:
hmm, any clue @jhheider https://github.com/teaxyz/pantry.extra/actions/runs/4303385056/jobs/7503089508#step:7:168

mxcl:
Confused by this: https://github.com/teaxyz/pantry.extra/actions/runs/4303385056/jobs/7503089508#step:7:168

jhheider:
> hmm, any clue @jhheider https://github.com/teaxyz/pantry.extra/actions/runs/4303385056/jobs/7503089508#step:7:168

Interesting. I wonder if `--unlink` would fix it, and if any other method would as well. I don't see where it's finding the conflict, unless it's unpacking over an older install which was a dir not a link (most likely).

Probably not too risky to try `--unlink` for this specific spot.

jhheider:
There you go, buddy. All green.

mxcl:
Sweet thanks.

In general the trick I'm doing here is a good one for making more builds (ours and end-user) just work.

Maybe it's a bit dirty. The other option is supplementing `env` (`CPATH`, etc.), cleaner but there'll be cases where a tool doesn't check the env.

PR #437, opened 2023-03-01
fix(aws.amazon.com/cli)

relsunkaev:
[This](https://github.com/teaxyz/pantry.extra/blob/ff249734f360a5a1c429539bc138fec3a6827d93/projects/aws.amazon.com/cli/package.yml#L21) line creates an `_awscrt` symlink to the `awscrt` directory, mistakenly thinking that the srcs are referencing this directory when importing `_awscrt`. Meanwhile, `_awscrt` is actually a compiled binary that is placed in the site-packages directory on install and the symlink breaks imports from that binary.

```yaml
script: |
    ...
    # oddly the srcs import both names for this and we have to symlink or it doesn't work
    ln -s awscrt _awscrt
```


mxcl:
Thanks. I thought it was odd we had to do this. Even more odd that in fact we didn't at all.

We could use a better test for this pkg.

PR #439, opened 2023-03-02
+eigen.tuxfamily.org

sanchitram1:
One of the homebrew packages, see https://github.com/teaxyz/pantry.extra/issues/372

### DIFF
--- a/projects/eigen.tuxfamily.org/package.yml
+++ b/projects/eigen.tuxfamily.org/package.yml
@@ -6,23 +6,23 @@ versions:
   - 3.4.0
 
 build:
-  working-directory: ./eigen-build 
+  working-directory: build
   dependencies:
     tea.xyz/gx/cc: c99
     tea.xyz/gx/make: '*'
     cmake.org: '^3'
   script: |
-    pwd
     cmake .. -DCMAKE_INSTALL_PREFIX={{prefix}}
-    make install 
+    make install
   test:
     make test
 
 test:
   dependencies:
     tea.xyz/gx/cc: c99
+    freedesktop.org/pkg-config: ^0.29
   env:
-    cxxflags: "-I{{prefix}}/include/eigen3"
+    cxxflags: $(pkg-config --cflags eigen3)
   fixture: |
     #include <iostream>
     #include <Eigen/Dense>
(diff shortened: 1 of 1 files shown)


projects/eigen.tuxfamily.org/package.yml
```diff
@@ -0,0 +1,44 @@
...
+  url: https://gitlab.com/libeigen/eigen/-/archive/{{version}}/eigen-{{version}}.tar.gz
+  strip-components: 1
+
+versions:
+  - 3.4.0
+
+build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
```
mxcl, line 12: `cmake.org: ^3`

sanchitram1:
- Replaced the `mkdir...` and `cd...` steps with a `working-directory: eigen-build` key in yaml
- Added `cmake.org` as a build & test dependency
- Since eigen is a header-only library, no need for the `provides` tag (no binaries)
- Since it's header only, you have to specify eigen3 in the include path. This is stored in `cxxflags`
- Replaced diff with a regular string comparison.

sanchitram1:
Closes #372

projects/eigen.tuxfamily.org/package.yml
```diff
@@ -0,0 +1,48 @@
...
+  - 3.4.0
+
+build:
+  working-directory: ./eigen-build 
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+    cmake.org: '^3'
+  script: |
+    pwd
```
mxcl: innocuous but shouldn't be here

projects/eigen.tuxfamily.org/package.yml
```diff
@@ -0,0 +1,48 @@
...
+    cmake .. -DCMAKE_INSTALL_PREFIX={{prefix}}
```
mxcl, line 15: Should add `-DCMAKE_BUILD_TYPE=Release`

mxcl:
Note I'm doing a newish thing here where if the includes are eg. `include/NAMESPACE/foo` we make NAMESPACE a symlink to `include` and mv foo to `include/foo`.

This is for robustness, it will fix a lot of potential failures in builds for both us and our users.

We can do this because we don’t install things.

PR #443, opened 2023-03-02
add kubectl

ABevier:


mxcl:
Hack is probs acceptable but I'd rather we pig’d rsync

PR #446, opened 2023-03-03
+xiph/rav1e

sanchitram1:
Video encoder for av1. Test is just `rav1e --version`


projects/github.com/xiph/rav1e/package.yml
```diff
@@ -0,0 +1,27 @@
...
+   url: https://github.com/xiph/rav1e/archive/refs/tags/v{{ version }}.tar.gz
+   strip-components: 1
+
+versions:
+  github: xiph/rav1e/tags  # reads github tags from github
+  strip: /^(?!v).*/
+
+dependencies:
+  rust-lang.org: ">=1.60"
+  rust-lang.org/cargo: "*"
```
mxcl: It doesn't have run time dependency on rust or cargo.

projects/github.com/xiph/rav1e/package.yml
```diff
@@ -0,0 +1,27 @@
+distributable:
+   url: https://github.com/xiph/rav1e/archive/refs/tags/v{{ version }}.tar.gz
+   strip-components: 1
+
+versions:
+  github: xiph/rav1e/tags  # reads github tags from github
+  strip: /^(?!v).*/
```
mxcl: Hmm, what's this about?

mxcl (changes requested):

projects/github.com/xiph/rav1e/package.yml
sanchitram1: Some of the pre-release tags have p<date> like p20230101.  They were breaking the build, so I excluded them when pulling versions.  If we wanted to include them, how would we do so?

projects/github.com/xiph/rav1e/package.yml
sanchitram1: Fixed

projects/github.com/xiph/rav1e/package.yml
jhheider: `versions.ignore: /^p/`

jhheider: Oh, sorry, you want to _include_ them? `versions.strip: /^p/`.

jhheider:
What is the additional file for? Is it test input? If so, you can remove the `test.fixture` key and reference it directly in your test.

projects/github.com/xiph/rav1e/package.yml
mxcl: He doesn't want to include them, he thought that I was telling him to include them.

But no! I was indeed wondering what you were doing because yes, `ignore` is the correct solution here.

projects/github.com/xiph/rav1e/package.yml
```diff
@@ -0,0 +1,25 @@
...
+
+dependencies:
+  nasm.us: 2.14.02
```
mxcl: why so specific?

PR #451, opened 2023-03-03
+az (azure-cli)

### LINKED ISSUE #302: +azure-cli (490/548)
Formulae: https://formulae.brew.sh/formula/azure-cli

mfts:
closes #302


mfts:
The test is actually not working properly. It works locally on macOS. I guess it's something with the virtualenv. Will continue to check

projects/microsoft.com/azure-cli/package.yml
```diff
@@ -0,0 +1,58 @@
...
+        executable = \$TEA_PYTHON
+      EOSH
+      find "\$VIRTUAL_ENV"/bin -maxdepth 1 -type f | xargs \\
+        sed -i.bak "1s|.*|#!\$VIRTUAL_ENV/bin/python|"
+      rm "\$VIRTUAL_ENV"/bin/*.bak
+      ln -sf "\$TEA_PYTHON" "\$VIRTUAL_ENV"/bin/python
+      exec "\$VIRTUAL_ENV"/bin/az "\$@"
+    EOF
+
+    chmod +x ../bin/"az"
```
jhheider: Does our existing `python-venv.sh` script not work for this?

mfts: no because in the root of the repo there's no setup.py. azure has three subrepos (see line 27-29)

mfts: It works locally, but I feel like something's not right here

jhheider: Hm. What if you ran them back-to-front:
```sh
for venv in azure-cli{-telemetry,-core,}; do
  python-venv.sh "{{prefix}}"/az
done
```
?

Just thinking that gets the most battle-testing in terms of venv management.

mfts: Doesn't work like that unfortunately, or I don't understand where you'd like to add it. We need some things that `python-venv.sh` is doing but not all

jhheider: Roger. I figure that'd build all three parts into the venv, and overwrite the final linkage each time; but if it's doesn't work, it doesn't work.

mfts: Ok I may have found a way: 

```
for venv in azure-cli{-telemetry,-core,}; do
  SRCROOT={{prefix}}/../src-v{{version}}/src/"$venv" python-venv.sh "{{prefix}}"/bin/az
done
```

mfts:
Cool it worked and tests pass too

projects/microsoft.com/azure-cli/package.yml
jhheider: Magical. I love standardization.

projects/microsoft.com/azure-cli/package.yml
```diff
@@ -0,0 +1,28 @@
...
+
+build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+    rust-lang.org: '*'
+    freedesktop.org/pkg-config: '*'
+  script: |
+    for venv in azure-cli{-telemetry,-core,}; do
+      SRCROOT={{prefix}}/../src-v{{version}}/src/"$venv" python-venv.sh "{{prefix}}"/bin/az
```
jhheider: I _think_ `{{prefix}}/../src-v{{version}}/src/"$venv"` is the same thing as `./src/"$venv"`.

mfts: throws an error if i do that

```
Error: invalid absolute path: ./src/azure-cli-telemetry
    at new Path (file:///opt/tea.xyz/src-v0.24.9/src/vendor/Path.ts:37:19)
    at from_env (file:///opt/tea.xyz/src-v0.24.9/src/app.ts:175:22)
    at injection (file:///opt/tea.xyz/src-v0.24.9/src/app.ts:165:16)
    at file:///opt/tea.xyz/src-v0.24.9/src/app.ts:34:36
```

jhheider: Hm. Guess that only works for TEA_PANTRY_PATH. Other packages (like `tea.xyz/gx/cc`) suggest that $SRCROOT might be set by default. I don't know if this better:
```sh
_SRCROOT="$SRCROOT"
for ......
   SRCROOT="$_SRCROOT"/src/"$venv" ...
...
```

That's not much of an improvement, but it protects against future changes.

mfts: I don't really understand this `_SRCROOT="$SRCROOT"` but it works :)

jhheider: Ha. Captures the original `$SRCROOT` before we start overwriting it.

PR #453, opened 2023-03-04
+libarchive

mfts:
closes #229

PR #454, opened 2023-03-04
+tesseract

### LINKED ISSUE #234: +tesseract (224/548)
Formulae: https://formulae.brew.sh/formula/tesseract
Source: https://github.com/tesseract-ocr/

mfts:
- requires: #453 
- requires: #449 

closes #234


projects/tesseract-ocr.github.io/package.yml
```diff
@@ -0,0 +1,48 @@
...
+      - --prefix="{{prefix}}"
+      - --disable-dependency-tracking
+      - --datarootdir="{{prefix}}"/share
+
+provides:
+  - bin/tesseract
+
+test:
+  script: |
+    export TESSDATA_PREFIX="{{prefix}}"/share/tessdata/
```
jhheider: should this be a top-level `runtime.env.TESSDATA_PREFIX` key?

mxcl: If it's required for the test then it will indeed be required at runtime.

mfts: @jhheider could you help me understand what you mean by `runtime.env`?

do I add an `env` key in the first level of the package.yml? I couldn't find a package that has a similar setup 

```sh
...
dependencies:
    - ...

env:
    - TESSDATA_PREFIX="{{prefix}}"/share/tessdata/

build:
...
```

I tried that locally but adding `echo $TESSDATA_PREFIX` to test and running `pkg test` is empty

mxcl: using dots is a common shorthand for dictionary key specification. So he means:

```yaml
runtime:
  env:
    TESSDATA_PREFIX: foo
```

Which was functionality added ~December.

mfts:
Seems like the error occurs because the `runtime.env.TESSDATA_PREFIX` appends the same path twice

https://github.com/teaxyz/pantry.extra/actions/runs/4377779719/jobs/7661731801#step:9:202 

I tested it locally and only adding a single path, like `TESSDATA_PREFIX=/Users/builder/.tea/tesseract-ocr.github.io/v5.3.0/share/tessdata/` solves the problem

mxcl:
This is fixed in 0.25 (unreleased).

mxcl:
The fixture is 30kB, which is quite a lot considering I stopped Jacob from adding a 14kB patch.

The pantry is downloaded by everyone using tea, so ideally we keep it lean.

What options we got? Can we shrink it? We could download it if it's available elsewhere.

projects/tesseract-ocr.github.io/package.yml
```diff
@@ -0,0 +1,51 @@
+distributable:
+   url: https://github.com/tesseract-ocr/tesseract/archive/refs/tags/{{version}}.tar.gz
+   strip-components: 1
+
+versions:
+  github: tesseract-ocr/tesseract/tags
```
mxcl: do releases/tags so we get push updates

mxcl (changes requested):

mfts:
> What options we got? Can we shrink it? We could download it if it's available elsewhere.

I'll wget it in the test. No need to have it permanently in the package.yml

mxcl (approved):

PR #457, opened 2023-03-05
+git3

mfts:
git3.sh


mfts:
Build fails but works locally. cannot really understand the error. I'm editing the `scripts` key but not the yarn lockfile.

mfts:
I don't understand this issue: https://github.com/teaxyz/pantry.extra/actions/runs/4366203230/jobs/7635906200#step:6:151

Locally it works on both platforms. I'm changing the `scripts` in the package.json.

mxcl:
no clue from my end

jhheider:
`yarn install --frozen-lockfile`?

https://stackoverflow.com/questions/58482655/what-is-the-closest-to-npm-ci-in-yarn

mfts:
> `yarn install --frozen-lockfile`?

That should be `yarn install --immutable` since berry, but when using the --immutable flag, Yarn will throw an error if there's a discrepancy between the `yarn.lock` file and the `package.json` file. 
Not sure why they are out of sync in the first place.

jhheider:
Maybe remove the lockfile?

mfts:
doesn't like that either

jhheider:
maybe some kind of force-update lock file?

mfts:
I can only update lock file if i have node_modules and i don't have node_modules because I need to run `yarn install` first. very strange.

I'm gonna sleep and struggle on tomorrow

jhheider:
possibly with `npm`, though you're supposed to use one or the other, not both.

mxcl:
Probs worth opening the issue upstream. If you followed their instructions and it doesn't work it's over to them.

jhheider:
looks like maybe `classic.yarnpkg.com` might work.

jhheider:
However, this seems like a problem: https://git3.sh/git3.w3q/git3-cli/src/branch/master/git3.pkg.json. Maybe we can just inject our targets...

jhheider:
I suspect fix-elf/fix-machos are causing: https://github.com/vercel/pkg/issues/321

PR #436, opened 2023-03-01
+double-conversion

mfts:
closes https://github.com/teaxyz/pantry.extra/issues/258

PR #461, opened 2023-03-05
+folly

mfts:
- requires #436 
- requires #458 
- requires #460

### DIFF
--- a/projects/facebook.com/folly/package.yml
+++ b/projects/facebook.com/folly/package.yml
@@ -1,0 +1,63 @@
+distributable:
+   url: https://github.com/facebook/folly/releases/download/{{version.tag}}/folly-{{version.tag}}.tar.gz
+   strip-components: 0
+
+versions:
+  github: facebook/folly
+
+dependencies:
+  boost.org: '*'
+  gflags.github.io: '*'
+  google.com/glog: '*'
+  libevent.org: '*'
+  lz4.org: 1
+  openssl.org: ^1.1
+  tukaani.org/xz: 5
+  facebook.com/zstd: 1
+  gnu.org/coreutils: 9
+  google.github.io/snappy: '*'
+  google.com/double-conversion: ^3
+
+build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+    cmake.org: ^3.0.2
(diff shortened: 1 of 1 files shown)


mxcl:
maybe this https://github.com/Homebrew/homebrew-core/blob/e39672cc04741441793c0b95751ed76ec1b929aa/Formula/snappy.rb#L34

mfts:
> maybe this https://github.com/Homebrew/homebrew-core/blob/e39672cc04741441793c0b95751ed76ec1b929aa/Formula/snappy.rb#L34

Good idea. I tried that, unfortunately, the error appears much earlier in the build process 

The error is [searchable, but old](https://github.com/facebook/folly/issues/847).

jhheider:
https://github.com/facebook/folly/issues/847

mfts:
@jhheider i found that ticket. It doesn't tell me anything new or anything I'm not already doing.

jhheider:
At least one of the suggestions was doing *only* `mkdir _build` and *not* `cd _build` (which seems crazy but is work trying).

mfts:
> which seems crazy but is work trying)

very crazy, and that doesn't work either unfortunately

mxcl:
> > maybe this https://github.com/Homebrew/homebrew-core/blob/e39672cc04741441793c0b95751ed76ec1b929aa/Formula/snappy.rb#L34
> 
> Good idea. I tried that, unfortunately, the error appears much earlier in the build process
> 
> The error is [searchable, but old](https://github.com/facebook/folly/issues/847).

How did you try it? The linked fix is for `snappy` not this? Did you apply the fix to snappy, build snappy and then rebuild folly?

mfts:
> Did you apply the fix to snappy, build snappy and then rebuild folly?

Yes I applied to patch to snappy, build snappy and then build folly - all locally. 
My understanding is that if a dep is built locally, the local build process for another package uses the local dep over the tea-hosted dep

mxcl:
> My understanding is that if a dep is built locally, the local build process for another package uses the local dep over the tea-hosted dep

Yep that is correct.

mxcl:
lol figured it out. This is the first tarball that requires `strip-components: 0`

mxcl:
This was gnarly. Facebook did a botch job on their build system that used regexes, our build dir has a `+` in it which broke it.

PR #463, opened 2023-03-06
+quarto.org

ddobrinskiy:
[quarto.org](https://quarto.org) is a framework to render scientific documents from markdown docs or jupyter in python, julia and other languages


ddobrinskiy:
Ok, this one also needs more work to be relocatable, will come back to it later

mxcl:
It seems to download prebuilt binaries of a bunch of stuff it needs. To be in tea it would need to use our packages instead.

PR #466, opened 2023-03-06
+micro

davdroman:


davdroman:
Can't make it work... closing for now.

jhheider:
A bug was released in brewkit (currently pulling 0.5.1 and releasing 0.5.2), so if it was just that test issue, I wouldn't give up yet.

jhheider:
This is an interesting one. You need to get `-buildmode=pie` into the `ADDITIONAL_GO_LINKER_FLAGS`. possibly `make ADDITIONAL_GO_LINKER_FLAGS='-buildmode=pie'`, but that might just overwrite it. It might be necessary to use `sed` to inject it directly into the makefile with, for example, `sed -i.bak -e 's/ADDITIONAL_GO_LINKER_FLAGS = /ADDITIONAL_GO_LINKER_FLAGS = -buildmode=pie /' Makefile`.

davdroman:
@jhheider cool thanks for the tips! I'll take a look tomorrow.

jhheider:
Try:
```sh
sed -i.bak -e 's/^ADDITIONAL_GO_LINKER_FLAGS = /ADDITIONAL_GO_LINKER_FLAGS = "-buildmode=pie" /' Makefile
rm Makefile.bak
```

and see if that gets it.

magnusviri:
I think that fixed it. I don't know what that does though. I actually figured out that the command builds and runs, but as soon as the build script finishes, it doesn't work anymore. I'm assuming this is because pkg build modifies the binary. This step perhaps?

```
  fixing up built products
doing SLOW rpath fixes…
```

Anyway, that sed command seemed to fix it. My package.yml file is so hacked up that I have to reduce it down to the minimum to see what is really important.

jhheider:
It just adds the appropriate flag to the Makefile, since there doesn't seem to be a better way to do so.

magnusviri:
It's all fixed. I'll see if I can figure out how to add to this pull request.

magnusviri:
@jhheider I don't know what `-buildmode=pie` does. I think it's only required on Linux, so I'm going to put it in an if statement.

magnusviri:
I finally read that link. The segfault only happens on Linux.

jhheider:
It is optional on linux, but it's always on on Ubuntu. It's a hardened memory runtime that provides additional alignment guarantees, so if the system uses it, binaries must as well, but not vice versa.

I don't recommend being interested but the technologies are called position-independent code and position-independent executable if you want to know more.

magnusviri:
Well, I tried to follow [these instructions](https://github.com/magnusviri/tea_pantry/tree/main#working-on-other-peoples-pull-requests) but it was harder than that. I had to do this.

```
TEA_PANTRY_PATH=/opt/tea/tea.xyz/var/pantry tea gh repo set-default
TEA_PANTRY_PATH=/opt/tea/tea.xyz/var/pantry tea gh pr checkout 466
```

After I made my change and committed, I tried to do a push and I got this.

```
git push
remote: Permission to davdroman/pantry.extra.git denied to magnusviri.
fatal: unable to access 'https://github.com/davdroman/pantry.extra.git/': The requested URL returned error: 403
```

I've never done this type of thing before so I'm not sure how to proceed. Is it possible to push it to my fork and have this pull request reference my fork? I kind of doubt it... IDK.

magnusviri:
> I don't recommend being interested but the technologies are called position-independent code and position-independent executable if you want to know more.

LOL, I'm kind of interested, but there's a lot of things between where I'm at and those specifics. When I switch to Linux as my daily driver (been on my to-do list for over 3 years now) then I might start looking into that type of thing more.

magnusviri:
Here's [my branch](https://github.com/magnusviri/tea_pantry/tree/micro-editor.github.io)

jhheider:
There's a lot you can do with git remotes; possibly the easiest way to do it without adding remotes is to use the `edit file` feature on the web and just paste in your updated file.

PR #470, opened 2023-03-07
+libavif

mxcl:


mxcl:
@jhheider so… we now have to rebuild *everything*? https://github.com/teaxyz/pantry.extra/actions/runs/4354535585/jobs/7609968386#step:6:302

jhheider:
There will be things that need rebuilding, but they would have failed rebuilding (or testing on the expanded matrix) anyway, I believe.

Things built often should already be fixed.

mxcl:
Cool. Wasn’t being accusatory. More like: oh noe!

jhheider:
I'd love OSes to be more resilient in this regard, but that's probably a pipe dream.

mxcl:
@jhheider I rebuilt the dep but it still fails. strangely

jhheider:
Try [this](https://github.com/teaxyz/pantry.extra/blob/cfd38daa87d67b5cd2cdd824d42bc45b95ccff2b/projects/soliditylang.org/package.yml#L52-L54).

The CMAKE docs suggest it shouldn't be necessary, but I seem to find it is. Perhaps we should set those in `tea.xyz/gx/cc:runtime.env` as well?

Edit: ugh, no, those are cli inputs, not env vars.
Edit2: maybe we need to take a page from homebrew and have `{{ cmake_std_args }}`.

PR #476, opened 2023-03-08
Fix incorrect formatting in the README

revolter:
Preview: https://github.com/revolter/pantry.extra/blob/52ee05aa7229eae15036c7f2075e9598f7a46103/README.md#contributing

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,8 +47,8 @@ $ git push origin my-new-package
 $ gh pr create
 ```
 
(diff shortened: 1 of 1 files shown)


mxcl:
Interesting that this worked. I have used `\` to do this in the past. Thanks!

revolter:
Ah, wow, I didn't know that. That's great to know, since I don't like trailing spaces because they're invisible.

PR #480, opened 2023-03-08
+render

davdroman:
Attempting to add https://github.com/render-oss/render-cli

jhheider edit: requires https://github.com/teaxyz/brewkit/pull/71


jhheider:
Sometimes deno.lock is very annoying. This fix isn't exactly elegant, but it works.

PR #483, opened 2023-03-08
+vde

ABevier:
`vde` is dependency of `qemu` which is a dependency of `podman`

### DIFF
--- a/projects/virtualsquare.org/vde/package.yml
+++ b/projects/virtualsquare.org/vde/package.yml
@@ -35,4 +35,4 @@ build:
       - --prefix="{{prefix}}"
 
 test: |
-  echo $(vde_switch -v) | grep "VDE {{ version }}"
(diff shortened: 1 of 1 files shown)


projects/virtualsquare.org/vde/package.yml
```diff
@@ -0,0 +1,38 @@
...
+    ./configure $ARGS
+    make install
+  env:
+    ARGS:
+      - --disable-debug
+      - --disable-dependency-tracking
+      - --prefix="{{prefix}}"
+
+test: |
+  echo $(vde_switch -v) | grep "VDE {{ version }}"
```
jhheider: The `echo $()` can be removed, right?

mxcl: Merging anyway, but yeah, looks sus

PR #487, opened 2023-03-09
+libusb

ABevier:
`libusb` is a dependency of `qemu` which is a dependency of `podman`

### DIFF
--- a/projects/libusb.info/package.yml
+++ b/projects/libusb.info/package.yml
@@ -1,0 +1,82 @@
+distributable:
+  url: https://github.com/libusb/libusb/releases/download/v{{ version }}/libusb-{{ version }}.tar.bz2
+  strip-components: 1
+
+versions:
+  github: libusb/libusb/releases/tags 
+  strip: /^v/
+
+build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+    gnu.org/autoconf: ^2
+  script: |
+    ./configure $ARGS
+    make install
+  env:
+    ARGS:
+      - --prefix="{{prefix}}"
+      - --disable-dependency-tracking
+      - --disable-debug
+
+test:
+  dependencies:
+    tea.xyz/gx/cc: c99 
+  script: |
+    mv $FIXTURE test.c
+    gcc test.c -lusb-1.0 "-I{{ prefix }}/include/libusb-1.0" -o test
+    ./test
+  # This fixture comes from the examples provided by libusb
+  fixture: |
+    #include <stdio.h>
+    #include "libusb.h"
+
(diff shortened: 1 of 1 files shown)


jhheider:
Homepage is `libusb.info` it looks like.

projects/libusb.info/package.yml
```diff
@@ -0,0 +1,82 @@
...
+      - --prefix="{{prefix}}"
+      - --disable-dependency-tracking
+      - --disable-debug
+
+test:
+  dependencies:
+    tea.xyz/gx/cc: c99 
+  script: |
+    mv $FIXTURE test.c
+    gcc test.c -lusb-1.0 "-I{{ prefix }}/include/libusb-1.0" -o test
```
jhheider, line 28: Instead of the `-I` flag, we should prefer this on the build script:

```
cd "{{prefix}}/lib"
mv libusb-1.0/* .
rmdir libusb-1.0
ln -s . libusb-1.0
```

projects/libusb.info/package.yml
```diff
@@ -0,0 +1,82 @@
...
+  strip: /^v/
+
+build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+    gnu.org/autoconf: ^2
+  script: |
+    ./configure $ARGS
+    make install
```
jhheider, line 16: Don't forget `-j {{ hw.concurrency }}` to take advantage of parallelizable builds.

projects/libusb.info/package.yml
```diff
@@ -0,0 +1,82 @@
+distributable:
+  url: https://github.com/libusb/libusb/releases/download/v{{ version }}/libusb-{{ version }}.tar.bz2
+  strip-components: 1
+
+versions:
+  github: libusb/libusb/releases/tags 
+  strip: /^v/
```
jhheider, line 7: I would add an `ignore: /-rc/` to save us issues in the future.

jhheider:
And that test failure suggests it will require `systemd.io` be packaged for linux. That's likely bothersome.

ABevier:
> And that test failure suggests it will require `systemd.io` be packaged for linux. That's likely bothersome.

Yeah, that's troublesome.  Not sure how to easily fix that.

jhheider:
Well, there's a lo-fi trick for omitting systems: https://github.com/teaxyz/pantry.extra/blob/main/projects/freedesktop.org/libbsd/package.yml. `systemd.io` has a ton of deps, but, looking at the homebrew formula, we have most of them already.

projects/libusb.info/package.yml
mxcl: the version parser already rejects prereleases

PR #488, opened 2023-03-09
+pnpm.io: pnpx and node 19 support

relsunkaev:
- add `pnpx` support
- add nodejs.org^19 support

### DIFF
--- a/projects/pnpm.io/package.yml
+++ b/projects/pnpm.io/package.yml
@@ -11,7 +11,7 @@ provides:
   - bin/pnpx
 
 dependencies:
-  nodejs.org: '^14,^16,^18'
+  nodejs.org: '>=14 <20'
 
 build:
   script: |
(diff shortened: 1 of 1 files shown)


projects/pnpm.io/package.yml
```diff
@@ -8,13 +8,15 @@ versions:
 
 provides:
   - bin/pnpm
+  - bin/pnpx
 
 dependencies:
-  nodejs.org: '^14,^16,^18'
+  nodejs.org: '>=14 <20'
```
mxcl, line 14: @jhheider gunna merge this, but be aware in case it causes issues! In which case pls fix.

mxcl (approved):

projects/pnpm.io/package.yml
jhheider: Won't find fixes if we don't break stuff. Let's do it.

PR #492, opened 2023-03-09
+slirp

### LINKED ISSUE #330: +libslirp (153/548)
Formulae: https://formulae.brew.sh/formula/libslirp

ABevier:
Closes #330

### DIFF
--- a/projects/freedesktop.org/slirp/package.yml
+++ b/projects/freedesktop.org/slirp/package.yml
@@ -1,0 +1,52 @@
+distributable:
+  url: https://gitlab.freedesktop.org/slirp/libslirp/-/archive/v{{ version }}/libslirp-v{{ version }}.tar.gz
+  strip-components: 1
+
+versions:
+  - 4.7.0
+
+dependencies:
+  gnome.org/glib: ^2
+
+build:
+  dependencies:
+    mesonbuild.com: ^0.63
+    ninja-build.org: 1
+    freedesktop.org/pkg-config: ^0.29
+    tea.xyz/gx/cc: c99
+  script: |
+    meson build $ARGS
+    ninja -C build install all
+    cd "{{ prefix }}/lib"
+    mv pkgconfig/* .
+    rmdir pkgconfig
+    ln -s . slirp
+  env:
+    ARGS:
+      - -Ddefault_library=both 
+      - --prefix={{ prefix }}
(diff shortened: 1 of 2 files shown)


projects/freedesktop.org/libslirp/package.yml
```diff
@@ -0,0 +1,48 @@
+distributable:
+  url: https://gitlab.freedesktop.org/slirp/libslirp/-/archive/v{{ version }}/libslirp-v{{ version }}.tar.gz
```
jhheider: looks like we should name this `freedesktop.org/slirp` unless there's a project by that name. maybe. I tend to prefer the undecorated name, but we have both.

projects/freedesktop.org/libslirp/package.yml
```diff
@@ -0,0 +1,48 @@
...
+    tea.xyz/gx/cc: c99
+  script: |
+    mv $FIXTURE test.c
+    gcc test.c -lslirp -o test
+    ./test
+  fixture: |
+    #include <stdio.h>
+    #include <string.h>
+    #include <stddef.h>
+    #include <slirp/libslirp.h>
```
jhheider: This suggests the build script should end with:
```sh
cd "{{prefix}}/lib"
mv slirp/* .
rmdir slirp
ln -s . slirp
```

mxcl: I’ve been doing this when I find situations where other packages just do `#include <libslirp.h>` indicating they expected the headers to be installed un-namespaced. So we’re trying to be just works for ourselves and for end-users.

I expect this happened because the namespacing was introduced after not being namespaces for years. Or some distro's remove the namespacing and the package owner uses such a distro.

I’m not sure we should always do it… dunno.

projects/freedesktop.org/slirp/package.yml
```diff
@@ -0,0 +1,52 @@
...
+    ninja -C build install all
+    cd "{{ prefix }}/lib"
+    mv pkgconfig/* .
+    rmdir pkgconfig
+    ln -s . slirp
+  env:
+    ARGS:
+      - -Ddefault_library=both 
+      - --prefix={{ prefix }}
+      - --libdir={{ prefix }}/lib
```
mxcl, line 28: probs not required.

PR #497, opened 2023-03-13
rework for single pantry

jhheider:


mxcl:
k so safe to merge?

After will new-version builds happen here? If not can you make that happen.

I will then archive pantry.core and begin the final history transfers.

mxcl (approved):

jhheider:
I'm ready to make that happen. I believe you'll want to merge at the same time as the rename.

I believe CLI will need a small tweak for sync, but GitHub name redirection should cover for a bit.

jhheider:
It all seems to be working. on to phase `cur++`.

PR #748, opened 2023-03-14
+minikube +go-bindata

### LINKED ISSUE #331: +minikube (398/548)
Formulae: https://formulae.brew.sh/formula/minikube

penguwin:
This PR adds [minikube](https://github.com/kubernetes/minikube) and it's dependency [go-bindata](https://github.com/kevinburke/go-bindata/) to the pantry.

I had some issues getting `go-bindata` to compile in tea as it's latest release is not using go modules - which caused me further issues with `$GOROOT`, `$GOPATH` and name clashes 😅. I ended up patching in a `go.mod` file comparably to how it's done in the [nixpkgs](https://github.com/NixOS/nixpkgs/blob/9d18216aa3638f2f403cfeb2b45fc21ef822dc3c/pkgs/development/tools/go-bindata/default.nix#LL18) collection. Since the package doesn't have any external dependencies this ended up being simpler, lmk if there's a better way to work around this.

Resolves #331

### DIFF
--- a/projects/github.com/kevinburke/go-bindata/package.yml
+++ b/projects/github.com/kevinburke/go-bindata/package.yml
@@ -1,0 +1,39 @@
+distributable:
+  url: https://github.com/kevinburke/go-bindata/archive/refs/tags/v{{version}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: kevinburke/go-bindata
+
+provides:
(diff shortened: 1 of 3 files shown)


mxcl:
We're not familiar with better ways as yet! Though potentially setting some env would do it

PR #749, opened 2023-03-14
+ncdu

### LINKED ISSUE #691: +ncdu (449/548)
Formulae: https://formulae.brew.sh/formula/ncdu

jhheider:
resolves https://github.com/teaxyz/pantry.extra/issues/691


jhheider:
might need to use the C version:
- we don't seem to be able to get PIC/PIE through to LLVM and it looks like those features aren't exposed in zig these days (references are from 2019/2020)
- it doesn't appear to output MachO headers that ruby-machos can understand and work with. this was the same issue we had trying to build zig ourselves.

marking this draft until someone (possibly me) has more ideas.

PR #757, opened 2023-03-15
+kubectx

### LINKED ISSUE #694: +kubectx (453/548)
Formulae: https://formulae.brew.sh/formula/kubectx

penguwin:
Closes #694

### DIFF
--- a/projects/kubectx.dev/package.yml
+++ b/projects/kubectx.dev/package.yml
@@ -1,0 +1,25 @@
+distributable:
+  url: https://github.com/ahmetb/kubectx/archive/refs/tags/v{{version}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: ahmetb/kubectx
+
+provides:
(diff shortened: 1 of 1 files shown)


mxcl:
> Stable versions of kubectx and kubens are small bash scripts that you can find in this repository.
> 
> Starting with v0.9.0, kubectx and kubens are now rewritten in Go. They should work the same way (and we'll keep the bash-based implementations around) but the new features will be added to the new Go programs. Please help us test this new Go implementation by downloading the binaries from the [Releases page →](https://github.com/ahmetb/kubectx/releases)

So we should either depend on `gnu.org/bash` or compile the `go` versions (or use the go binaries they provide).

penguwin:
Thanks for looking over this, i totally missed that part. Added `gnu.org/bash` as dependency :+1:

mxcl:
Thank you!

PR #786, opened 2023-03-16
add bcrypt

mfts:
building this because `libxcrypt` throws test errors on linux. https://github.com/teaxyz/pantry.extra/actions/runs/4432685158/jobs/7777001117#step:9:56

according to gpt-4, 
> If the `crypt()` function is causing errors on both Linux aarch64 and x86_64 systems, it's possible that the `crypt()` function on these systems does not support the use of the "`$2b$`" prefix for the bcrypt algorithm.

so i wanna build `libxcrypt` with bcrypt support but i'm not so sure this is the right approach. 

worst case another package added

### DIFF
--- a/projects/bcrypt.sourceforge.net/package.yml
+++ b/projects/bcrypt.sourceforge.net/package.yml
@@ -1,0 +1,32 @@
+distributable:
+  url: https://bcrypt.sourceforge.net/bcrypt-{{version.marketing}}.tar.gz
+  strip-components: 1
+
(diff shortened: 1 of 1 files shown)


mxcl:
[needs zlib](https://github.com/teaxyz/pantry/actions/runs/4433036217/jobs/7777703247#step:6:133)

mfts:
I included `zlib` as a runtime dep

mxcl:
probs needs to be told where to look for zlib

PR #789, opened 2023-03-16
README: Fix link to wiki

fredrik:
The wiki link was pointing the wrong way, should be https://github.com/teaxyz/pantry.extra/wiki

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,6 @@ $ pkg edit
 ```
 
 
-[wiki]: https://github.com/teaxyz/pantry/wiki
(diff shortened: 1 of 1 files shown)


mfts:
@fredrik good catch. thanks for the PR.

However, we are actually going to rename `pantry.extra` to `pantry` very soon, so this is technically correct soon. I hold off on merging your PR.

mxcl:
Renamed it this morning! Sorry for the period where the links were wrong, non ideal but tricky logistically to do better.

PR #806, opened 2023-03-17
libtelnet, used by remote_cmds (which includes Apple's telnet)

magnusviri:


magnusviri:
I made this a draft because the test is not much of a test. It just includes one of the header files and sees if it compiles. I don't actually use any of the libtelnet code.

mxcl:
requiring `xcodebuild` will be fun. @jhheider just a heads up since this apparently will be required for some stuff :/

jhheider:
Well, I'd love to find an end-run around that. Is this library apple only? That seems crazy.

magnusviri:
Apple doesn't even ship telnet anymore. The gnu version works great. Maybe we can just shelf the Apple version. I doubt they will maintain it anyway. See https://github.com/teaxyz/pantry/pull/825

jhheider:
If we don't (until we do) need it, happy to shelve old code.

PR #815, opened 2023-03-18
+nano-editor.org

### LINKED ISSUE #717: +nano (495/548)
Formulae: https://formulae.brew.sh/formula/nano

getneil:
resolves #717

### DIFF
--- a/projects/nano-editor.org/package.yml
+++ b/projects/nano-editor.org/package.yml
@@ -5,16 +5,17 @@ versions:
   - 7.2
 
 dependencies:
-  gnu.org/gettext: '*'
-  invisible-island.net/ncurses: '>=6.0'
+  gnu.org/gettext: "*"
+  invisible-island.net/ncurses: ">=6.0"
 
 provides:
   - bin/nano
 
 build:
   dependencies:
-    freedesktop.org/pkg-config: '*'
-    gnu.org/automake: '*'
+    freedesktop.org/pkg-config: "*"
+    gnu.org/automake: "*"
+    gnu.org/make: "*"
     tea.xyz/gx/cc: "*"
   script: |
     ls
--- a/projects/nano-editor.org/package.yml
+++ b/projects/nano-editor.org/package.yml
@@ -14,8 +14,8 @@ provides:
 
 build:
   dependencies:
-    freedesktop.org/pkg-config: "*"
-    gnu.org/automake: "*"
+    freedesktop.org/pkg-config: ^0.29
+    gnu.org/automake: ^1.16
     gnu.org/make: "*"
     tea.xyz/gx/cc: "*"
   script: |
--- a/projects/nano-editor.org/package.yml
+++ b/projects/nano-editor.org/package.yml
@@ -1,0 +1,40 @@
+distributable:
+  url: https://www.nano-editor.org/dist/v{{ version.major }}/nano-{{ version.major }}.{{ version.minor }}.tar.gz
+
+versions:
+  - 7.2
+
+dependencies:
+  gnu.org/gettext: '*'
+  invisible-island.net/ncurses: '>=6.0'
+  freedesktop.org/pkg-config: '*'
+  tea.xyz/gx/cc: "*"
+
+provides:
+  - bin/nano
+
+build:
+  dependencies:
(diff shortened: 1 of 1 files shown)


projects/nano-editor.org/package.yml
```diff
@@ -0,0 +1,38 @@
...
+  dependencies:
+    freedesktop.org/pkg-config: "*"
+    gnu.org/automake: "*"
+    gnu.org/make: "*"
+    tea.xyz/gx/cc: "*"
+  script: |
+    cd nano-$VERSION && ./configure $ARGS
+    make install
+  env:
+    VERSION: "{{ version.major }}.{{ version.minor }}"
```
jhheider: I believe this can be:
```yaml
script: |
  cd nano-{{ version.marketing }}
  ./configure $ARGS
  make install
```

You can _try_:
```yaml
working-directory: nano-{{ version.marketing }}
```
instead, but I don't know if that passes through the de-moustaching.

projects/nano-editor.org/package.yml
```diff
@@ -0,0 +1,38 @@
+distributable:
+  url: https://www.nano-editor.org/dist/v{{ version.major }}/nano-{{ version.major }}.{{ version.minor }}.tar.gz
```
jhheider: You can use `nano-{{ version.marketing }}.tar.gz` or `{{ version.raw }}`, depending on which you think is correct. It's probably more correct to use `{{ version.raw }}` throughout.

projects/nano-editor.org/package.yml
getneil: that made it nicer. thank you

projects/nano-editor.org/package.yml
```diff
@@ -0,0 +1,38 @@
...
+provides:
+  - bin/nano
+
+build:
+  dependencies:
+    freedesktop.org/pkg-config: "*"
+    gnu.org/automake: "*"
+    gnu.org/make: "*"
+    tea.xyz/gx/cc: "*"
+  working-directory: nano-{{ version.raw }}
```
mxcl: Or just `strip-components: 1` at the distributable level like all other packages do.

projects/nano-editor.org/package.yml
```diff
@@ -0,0 +1,38 @@
...
+dependencies:
+  gnu.org/gettext: "*"
+  invisible-island.net/ncurses: ">=6.0"
+
+provides:
+  - bin/nano
+
+build:
+  dependencies:
+    freedesktop.org/pkg-config: "*"
```
mxcl: Should be `~0.29`

projects/nano-editor.org/package.yml
```diff
@@ -0,0 +1,38 @@
...
+    gnu.org/automake: "*"
```
mxcl: Almost certainly should be `^2`. I doubt automake will ever release another major version but I am certain that if they do it will be radically different and break everything that assumed otherwise.

projects/nano-editor.org/package.yml
```diff
@@ -0,0 +1,38 @@
...
+      - --sysconfdir="{{prefix}}"
+      - --enable-color
+      - --enable-extra
+      - --enable-multibuffer
+      - --enable-nanorc
+      - --enable-utf8
+
+test:
+  script: |
+    nano --version
```
mxcl: do a `test $(nano --version) = {{version}}` we've had a lot of packages that don’t contain their version info lately

projects/nano-editor.org/package.yml
getneil: we have only bottled 1.16.5 atm

projects/nano-editor.org/package.yml
getneil: i added a grep test here nano was not returning just the version in `--version`

getneil: i also noticed something strange. 

at the top the version.raw == 7.2
```
distributable:
  url: https://www.nano-editor.org/dist/v{{ version.major }}/nano-{{ version.raw }}.tar.gz
  strip-components: 1

versions:
  - 7.2
 ```

but when used in test block version.raw became 7.2.0

mxcl:
gw

PR #826, opened 2023-03-21
projects/plasmasturm.org

magnusviri:
This doesn't have a release or tag so I hardcoded the version.


projects/plasmasturm.org/rename/package.yml
```diff
@@ -0,0 +1,29 @@
...
+build:
+  script: |
+    pod2man rename rename.1
+    mkdir -p "{{ prefix }}"/bin
+    mv rename "{{ prefix }}"/bin
+    mkdir -p "{{ prefix }}"/share/man/man1
+    mv rename.1 "{{ prefix }}"/share/man/man1
+  env:
+    ARGS:
+      - --prefix="{{prefix}}"
```
mxcl: unused

projects/plasmasturm.org/rename/package.yml
```diff
@@ -0,0 +1,29 @@
+distributable:
+  url: https://github.com/ap/rename/archive/refs/heads/master.tar.gz
```
mxcl: Should be: https://github.com/ap/rename/archive/0da9faa241093c90a40b15a9ff4b36e9858d4516.tar.gz IMO

mxcl (changes requested):

projects/plasmasturm.org/rename/package.yml
magnusviri: I removed the unused env var

projects/plasmasturm.org/rename/package.yml
magnusviri: Turns out the developer was using tags, I just didn't notice. I switched it to use tags.

PR #829, opened 2023-03-21
+docbook.org

magnusviri:
I created this exactly the same way [Homebrew does](https://github.com/Homebrew/homebrew-core/blob/388cf58311b12654b7392950ec3f1e6082b38564/Formula/docbook.rb). I don't know enough about docbook to know if this is how it really should be done. Please advise if I need to change it. I packaged this so that I can package [xmlto](https://pagure.io/xmlto/), so that I can update the [wumpus](https://github.com/teaxyz/pantry/tree/main/projects/catb.org/wumpus) recipe to 1.9 and gitlab, because 1.9 requires xmlto to create the man page.

### DIFF
--- a/projects/docbook.org/package.yml
+++ b/projects/docbook.org/package.yml
@@ -1,0 +1,42 @@
+versions:
+  - 4.1.2
+  - 4.2
+  - 4.3
+  - 4.4
+  - 4.5
+  - 5.0
+  - 5.1
+
+build:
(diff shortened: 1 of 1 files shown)


mxcl:
Seemed like it was a good start. Docbook seems like a weird one, but it may be necessary to package it like this at least at first based on how the tools use it.

PR #830, opened 2023-03-21
+docbook.org

magnusviri:


magnusviri:
I created this exactly the same way [Homebrew does](https://github.com/Homebrew/homebrew-core/blob/388cf58311b12654b7392950ec3f1e6082b38564/Formula/docbook.rb). I don't know enough about docbook to know if this is how it really should be done. Please advise if I need to change it. I packaged this so that I can package [xmlto](https://pagure.io/xmlto/), so that I can update the [wumpus](https://github.com/teaxyz/pantry/tree/main/projects/catb.org/wumpus) recipe to 1.9 and gitlab, because 1.9 requires xmlto to create the man page.

projects/docbook.org/package.yml
```diff
@@ -0,0 +1,49 @@
...
+build:
+  dependencies:
+    curl.se: '*'
+    info-zip.org/unzip: '*'
+  script: |
+    if [ -e "docbook" ]; then
+        rm -r docbook
+    fi
+    mkdir -p docbook/xml
+    VERSIONS=(4.1.2 4.2 4.3 4.4 4.5 5.0 5.1) #{{ versions }}
```
mxcl, line 19: nice future thinking comment

projects/docbook.org/package.yml
```diff
@@ -0,0 +1,49 @@
...
+  - 4.5
+  - 5.0
+  - 5.1
+
+build:
+  dependencies:
+    curl.se: '*'
+    info-zip.org/unzip: '*'
+  script: |
+    if [ -e "docbook" ]; then
```
mxcl, line 15: yeah we really need a smart-clean thing in brewkit

mxcl (approved):

mxcl:
I looked into this and I think we may need a runtime env var, or perhaps to not include all versions together (brew does it because it doesn't support independent versions as well, but well… this may be what we need… we'll find out!)

Thanks

Issue #681, opened 2023-03-13
+rustup-init (434/548)


mfts:
Formulae: https://formulae.brew.sh/formula/rustup-init

PR #832, opened 2023-03-21
+rustup-init

apwadkar:
I wasn't entirely sure which folder this should be under. Seems like `rust-lang.org` might be better, but I wasn't sure if that would conflict with the other packages in there.

This builds and produces a binary, but it shows that the binary is a modified version, and I'm not entirely sure what the reason for that is.


mxcl:
* I do think the name should be `rust-lang.org/rustup`
* Looks like you need to depend on `freedesktop.org/pkg-config: ~0.29`

> This builds and produces a binary, but it shows that the binary is a modified version, and I'm not entirely sure what the reason for that is.

Is that what `--version` says? Perhaps you need to add a git tag or something to set the version info?

apwadkar:
> - I do think the name should be `rust-lang.org/rustup`

I've moved it to `rust-lang.org/rustup-init` since this produces the `rustup-init` binary not `rustup` and is the package name for #681.

> Is that what `--version` says? Perhaps you need to add a git tag or something to set the version info?

When I build locally and with the Linux docker container, `rustup-init --version` returns something like `rustup-init unknown (bd020d138 2023-03-21)`. But when looking at [CI output](https://github.com/teaxyz/pantry/actions/runs/4483000846/jobs/7881836388), it shows the right output: `rustup-init 1.25.2 (2023-03-21)`.

mxcl:
Thank you for your contribution! The package will be available shortly (actions will rebuild), you will need to `tea -S` first.

jhheider:
closes #681

PR #838, opened 2023-03-22
fish: tests must specify they want tea in the PATH now

mxcl:
This is more proper. One can imagine a future where tools alter their behavior if `tea` is available because we are useful. So one should opt into testing that.

### DIFF
--- a/projects/fishshell.com/package.yml
+++ b/projects/fishshell.com/package.yml
@@ -3,7 +3,7 @@ distributable:
   strip-components: 1
 
 versions:
-  github: fish-shell/fish-shell/tags
+  github: fish-shell/fish-shell/releases/tags
 
 dependencies:
   gnu.org/gettext: '*'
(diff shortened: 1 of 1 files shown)


mxcl:
@jhheider new test verifies that fish command not found handler will not redirect stdout to a file. Can you take a look? if we cannot easily fix it then open a ticket and comment out the test thanks.

jhheider:
This doesn't seem to be at all trivial, and ChatGPT seems just as stymied as I. It doesn't seem like this is an intended function of either functions or the command_not_found handler. In a normal function, you might be able to use `(read)` to grab stdin, or possibly some `isatty` tricks. These seem like consequences of fish design decisions below the level of user-customizability.

mxcl:
shit. Well without it the magic is less. Though I have plans for a more powerful magic that may make this ok.

Why is fish so popular? Genuine question. I barely heard of it before all this.

jhheider:
> Why is fish so popular? Genuine question. I barely heard of it before all this.

Its focus is on the interactive experience. I won't speak for others, but for me its syntax more closely matches programming languages, its functions are powerful (possibly why they don't _need_ to play nicely with pipes), and it has fewer quoting issues due to use lists by default internally.

mxcl:
So the failure on linux is repeatable locally. It's like the magic doesn’t work at all since fish apparently is complaining about go:

```
~/xyz.tea.fixture (line 10):
  go version
  ^
```

PR #841, opened 2023-03-23
[fix]: add chatblade version

mfts:
chatblade added versioned releases, so updating the `package.yml` respectively


jhheider:
Oh, thank god. I hate the unversioned dance.

PR #849, opened 2023-03-23
fix(python)

jhheider:
looks like python needs a db lib on linux


mxcl:
lol you can't just add a dep to python without review mate.

Where's the justification for this in the PR AND the commit?

jhheider:
Sorry, tunnel vision as usual. Mirroring homebrew's config. Looks like there were changes in the db defaults for 3.11 per the docs.

PR #850, opened 2023-03-23
+lcov

ABevier:
Adds the `lcov` package for creating test coverage reports.

### DIFF
--- a/projects/github.com/linux-test-project/lcov/package.yml
+++ b/projects/github.com/linux-test-project/lcov/package.yml
@@ -1,0 +1,46 @@
+distributable:
+  url: https://github.com/linux-test-project/lcov/releases/download/v{{ version.major }}.{{ version.minor }}/lcov-{{ version.major }}.{{ version.minor }}.tar.gz
+  strip-components: 1
+
+versions:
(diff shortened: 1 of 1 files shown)


projects/github.com/linux-test-project/lcov/package.yml
```diff
@@ -0,0 +1,46 @@
+distributable:
+  url: https://github.com/linux-test-project/lcov/releases/download/v{{ version.major }}.{{ version.minor }}/lcov-{{ version.major }}.{{ version.minor }}.tar.gz
```
jhheider: `v{{ version.major }}.{{ version.minor }}` is either `v{{ version.marketing }}` (if you're specifically dropping the patch level, or `v{{ version.raw }}` if you're dealing with a project that might have 2- or 3-digit versions.

jhheider:
That test error isn't great. We probably need to build more backends or plugins with `llvm`.

ABevier:
> That test error isn't great. We probably need to build more backends or plugins with `llvm`.

From my understanding lcov is used for gcc and llvm-cov is used for clang, so I'm going to explicitly require gcc.

PR #851, opened 2023-03-24
+tldr.sh

vladdeSV:


vladdeSV:
First PR, and would appreciate some help. This is based on the [Homebrew formula](https://github.com/Homebrew/homebrew-core/blob/ba882e339248cbecb1d27c233e04a083df517741/Formula/tldr.rb#L20-L23).

Some issues I do not know how to solve:
1. `+libzip.org` is [not bottled for linux+aarch64](https://tea.xyz/+libzip.org/). I am unsure what to do.

1. I cannot compile without `libzip-dev`. How do I specify this in package.yml?

1. If I were to run the application without `libzip4` installed I get this type of error: `./tldr: error while loading shared libraries: libzip.so.4: cannot open shared object file: No such file or directory`. What is the solution here?

mxcl:
> I cannot compile without libzip-dev. How do I specify this in package.yml?

We don’t separate `-dev` packages so depending on libzip is enough

[The error is you need to add `curl.se: '*'` as a dep](https://github.com/teaxyz/pantry/actions/runs/4512442583/jobs/7945919183?pr=851#step:6:163)

@jhheider any clue why we don't have lib zip for linux aarch64?

jhheider:
```sh
$ curl -s dist.tea.xyz/bottles.txt | grep libzip
libzip.org/darwin/aarch64/v1.9.2.tar.xz
libzip.org/darwin/x86-64/v1.9.2.tar.xz
libzip.org/linux/x86-64/v1.9.2.tar.gz
```
mysterious. ancient history build failure? I'll trigger a rebuild.

jhheider:
Rebuilt cleanly. Build now passing. Test failing on ubuntu-latest. Should be able to replace the `test:` as written with:
```yaml
test: tldr --version | grep {{ version }}
```
though that doesn't explain the failure. I wonder if there's a `grep` or redirection issue on ubuntu-latest, or if something (segfault?) is being masked.

vladdeSV:
Changed the test to the simpler variant you provided.

> though that doesn't explain the failure

Should I change the test to something else then? It does pass on other systems.

What would an approproiate action here be?

jhheider:
Based on the [build logs](https://github.com/teaxyz/pantry/actions/runs/4515124860/jobs/7952051027?pr=851) it seems to want but not need `git` and `pkg-config` for building. Let's try this as a test (I suspect something is segfaulting):
```yaml
test: |
  tldr --version
  tldr ls
  tldr --version | grep {{ version }}
```

vladdeSV:
> (I suspect something is segfaulting)

You are correct – it's the binary. From my local machine (Debian WSL2):

```
testbeds/tldr.sh-1.5.0+linux/xyz.tea.test.sh: line 21:   813 Segmentation fault      tldr --version
```

Pushing changes to see output on different machines ASAP.

jhheider:
https://github.com/tldr-pages/tldr-c-client/blob/821e08c4df46e817409f026c0b7f2887c6489c3e/Makefile#L15-L20

Looks like the Makefile is overriding our flags. Try this:
```sh
  script: make $ARGS --jobs {{ hw.concurrency }} install
  env:
    ARGS:
      - PREFIX="{{prefix}}"
    linux:
      ARGS:
        - CFLAGS="-ggdb -O0 -ftrapv $CFLAGS"
        - CPPFLAGS="$CPPFLAGS"
        - LDFLAGS="$LDFLAGS"
```
Segfaults on linux almost always mean the `-fPIC/-pie` flags didn't make it through.

vladdeSV:
Hm. I'm at the point where I'm blindly copying. I don't really understand what effect the changes will have.

I'd hate to leave this half-finished, but right now I don't know where to go.

Maybe not the best project for a first PR 😅😬

jhheider:
Hm, yeah, that's the shell quoting thing. Let me see.

jhheider:
This doesn't have the shell-quoting issue (tested):
```yaml
  script: |
    make PREFIX="{{prefix}}" \
      CFLAGS="-ggdb -O0 -ftrapv $CFLAGS" \
      CPPFLAGS="$CPPFLAGS" \
      LDFLAGS="$LDFLAGS" \
      --jobs {{ hw.concurrency }} \
      install
```

Bash's quoting rules are kind of the worst for this.

jhheider:
As for what the changes are _doing_, the Makefile sets a series of flags unless overrides are passed (similar to `PREFIX`). We need to ensure that our `-fPIC/-pie` flags from `tea.xyz/gx/cc` make it in because ubuntu and others turn on PIE/PIC by default, so we see segfaults on binaries compiled without them.

I can't recommend not digging into PIE/PIC enough, as its a warren. But that's our hard-fought insight into this issue.

jhheider:
*Much* better. Needs `curl.se` to be a runtime dependency, rather than a build dependency (for `libcurl`). Otherwise, I bet this is done.

jhheider:
Feels good when it all goes green.

jhheider:
Thanks, @vladdeSV !

vladdeSV:
Thanks for the help @jhheider!

jhheider:
Any time!

PR #859, opened 2023-03-25
+github.com/jaseg/lolcat

magnusviri:
https://raw.githubusercontent.com/jaseg/lolcat/master/LOLCat-Rainbow.jpg

This lolcat clone is an attempt to reduce the world's carbon dioxide emissions by optimizing inefficient code. It's >10x as fast and <0.1% as large as the original one.

c-lolcat << python-lolcat << ruby-lolcat

### DIFF
--- a/projects/github.com/jaseg/lolcat/package.yml
+++ b/projects/github.com/jaseg/lolcat/package.yml
@@ -20,4 +20,4 @@ provides:
 
 test:
   script: |
-    test "$(lolcat --version)" = "lolcat {{version.raw}}, (c) 2023 jaseg"
+    lolcat --version | grep "lolcat version {{version.marketing}}"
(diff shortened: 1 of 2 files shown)


magnusviri:
I don't understand why the test is still failing. This is the test.

```
test "$(lolcat --version)" = "lolcat {{version.raw}}, (c) 2023 jaseg"
```

This is the output.

```
test 'lolcat version 1.4, (c) 2023 jaseg' = 'lolcat 1.4.0, (c) 2023 jaseg'
```

Is version.raw not supported in the test section?

jhheider:
It certainly should be. I believe the problem is that `test` is getting the supplied version ("`tea` version"), which is `1.4.0`. Using `{{ version.marketing }}` will solve it for this specific software, but we should investigate a full fix.

You're missing the word "version" in your test string, fyi. You might also want to use `grep` instead of an exact test, since the copyright year is likely to drift in ways that aren't precisely predictable (can't use `date +%Y`, for example, because building `1.4` won't work next January).

mxcl (approved):

PR #885, opened 2023-03-27
more fix(python)

jhheider:
@mxcl did that fixture ever work? Adding the `bdb` dep made it work on 3.11, but it still doesn't work on 3.9 (https://github.com/teaxyz/pantry/actions/runs/4533934635/jobs/7987503539)

Commenting out that test and the questionable "fix" until we can work this out.


mxcl:
you're asking me to remember things and that is tricky.

jhheider:
don't I know it? so much juggling.

PR #890, opened 2023-03-27
+github.com/psf/black

magnusviri:
This isn't done.


magnusviri:
This requires https://github.com/teaxyz/brewkit/pull/66.

mxcl:
brewkit v0.20 has the venv script rewritten in python with support for specifying extras:

```
python-venv.py {{prefix}}/bin/foo --extra foo --extra bar
```

PR #1005, opened 2023-03-29
+abseil.io

magnusviri:


magnusviri:
Linux is getting this error:

```
+ cmake -S . -B build -DCMAKE_CXX_STANDARD=17 -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_RPATH=/opt/abseil.io/v20230125.2.0/bla -DCMAKE_BINARY_DIR=/opt/abseil.io/v20230125.2.0/bin -DABSL_PROPAGATE_CXX_STD=ON -DCMAKE_INSTALL_PREFIX=/opt/abseil.io/v20230125.2.0 -DCMAKE_INSTALL_LIBDIR=/opt/abseil.io/v20230125.2.0/lib -DCMAKE_BUILD_TYPE=Release -DCMAKE_FIND_FRAMEWORK=LAST -DCMAKE_VERBOSE_MAKEFILE=ON -Wno-dev -DBUILD_TESTING=OFF
CMake Error: CMake was unable to find a build program corresponding to "Unix Makefiles".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
Error: Process completed with exit code 1.
```
I have no clue what to do about this. Help?

jhheider:
I believe you need to depend on `tea.xyz/gx/cc: c99` to get a c/c++ compiler in there. You'll need it for your test node as well.

magnusviri:
I added this and I think it's right but it still doesn't work.

```
build:
  dependencies:
    cmake.org: ^3
    linux:
      tea.xyz/gx/cc: c99
```

jhheider:
If linux needs a c-compiler (and make), then we should require them on both:
```yaml
build:
  dependencies:
    cmake.org: ^3
    tea.xyz/gx/cc: c99
    tea.xyz/gx/make: '*'
```

magnusviri:
> If linux needs a c-compiler (and make), then we should require them on both:
> 
> ```yaml
> build:
>   dependencies:
>     cmake.org: ^3
>     tea.xyz/gx/cc: c99
>     tea.xyz/gx/make: '*'
> ```

It tested without make so I didn't include it there.

Why would you require it on both? I'm mostly a programmer who checks if it works rather than understand why it works. So I'm one of them. I guess I'm trying to understand this.

mxcl:
Is both here build and test? they are unrelated. test does not inherit anything from the build node. Both inherit from the higher level dependencies node though.

projects/abseil.io/package.yml
```diff
@@ -0,0 +1,42 @@
...
+  dependencies:
+    cmake.org: ^3
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+
+  script: |
+    cmake -S . -B build \
+      -DCMAKE_CXX_STANDARD=17 \
+      -DBUILD_SHARED_LIBS=ON \
+      -DCMAKE_INSTALL_RPATH={{ prefix }}/bla \
```
mxcl, line 18: seems spurious?

projects/abseil.io/package.yml
```diff
@@ -0,0 +1,42 @@
...
+      -DCMAKE_BINARY_DIR={{ prefix }}/bin \
```
mxcl, line 19: probs not necessary since this will be the default

projects/abseil.io/package.yml
```diff
@@ -0,0 +1,42 @@
...
+      -DABSL_PROPAGATE_CXX_STD=ON \
+      -DCMAKE_INSTALL_PREFIX={{ prefix }} \
+      -DCMAKE_INSTALL_LIBDIR={{ prefix }}/lib \
```
mxcl, line 22: probs not necessary since this will be the default

projects/abseil.io/package.yml
```diff
@@ -0,0 +1,42 @@
...
+      -DCMAKE_BUILD_TYPE=Release \
+      -DCMAKE_FIND_FRAMEWORK=LAST \
```
mxcl, line 24: if this is required needs a comment

mxcl (approved):
Comments are not mandatory changes since it works. But worth reading ffr.

projects/abseil.io/package.yml
magnusviri: I copied the Homebrew formula. I don't even know what this library does, much less if these flags are required. I might not be the best person to build packages...

jhheider:
> It tested without make so I didn't include it there.
> 
> Why would you require it on both? I'm mostly a programmer who checks if it works rather than understand why it works. So I'm one of them. I guess I'm trying to understand this.

`tea.xyz/gx/*` ensure that the relevant tool is present. It works on macOS because Apple ships `make` with the OS, but another compliant `darwin` platform could not. So, it is requiring `make` on macOS, it just always[citation needed] finds it.

PR #835, opened 2023-03-21
fix linux-utils on mac

mxcl:
Closes #831

PR #1046, opened 2023-03-30
fix(util-linux)

jhheider:
clean up header locations


jhheider:
closing in favor of #835

PR #1121, opened 2023-03-30
+mop-tracker

sanchitram1:
Financial dashboard on terminal

### DIFF
--- a/projects/github.com/mop-tracker/mop/add-version-flag.patch
+++ b/projects/github.com/mop-tracker/mop/add-version-flag.patch
@@ -1,0 +1,40 @@
+From 2fe7f1e0f29aa69520d569887a45627afc06bac6 Mon Sep 17 00:00:00 2001
+From: Sanchit Ram Arvind <sanchitram@gmail.com>
+Date: Mon, 3 Apr 2023 23:43:12 -0400
+Subject: [PATCH] Add version flag
+
+---
+ cmd/mop/main.go | 10 ++++++++++
+ 1 file changed, 10 insertions(+)
+
+diff --git a/cmd/mop/main.go b/cmd/mop/main.go
+index 4636339..c87bfdc 100644
+--- a/cmd/mop/main.go
++++ b/cmd/mop/main.go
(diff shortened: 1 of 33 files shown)


jhheider:
my guess is you won't be able to test a curses-style binary that needs a TTY. this is a hole in our testing process currently. hopefully you can use `--version` or `--help` to ensure the binary doesn't segfault, at least.

sanchitram1:
The only flag it has is `-profile string` to add a new location for the `.moprc` it uses to pre-load the stocks.  Don't think it has a `-version` or a `-help`(?). 

I actually think that all packages should provide both the above flags at a minimum.  I can raise an issue / PR in the original repo to see if there's any other way / flags they provide?

jhheider:
I had this problem with [sl](https://github.com/teaxyz/pantry/blob/main/projects/github.com/mtoyoda/sl/version.patch). I chose to add the `version` flag myself with a patch.

projects/github.com/mop-tracker/mop/add-version-flag.patch
```diff
@@ -0,0 +1,40 @@
...
+
+diff --git a/cmd/mop/main.go b/cmd/mop/main.go
+index 4636339..c87bfdc 100644
+--- a/cmd/mop/main.go
++++ b/cmd/mop/main.go
+@@ -42,6 +42,7 @@ Enter comma-delimited list of stock tickers when prompted.
+ 
+ <r> Press any key to continue </r>
+ `
++var teaVersion = "Mop 1.0.0"
```
jhheider: Two things:
- this should probably return `2023.3.30`.
- this should likely be passed through `sed` to set on the fly:
  - `sed 's/^var teaVersion = .*$/var teaVersion = "{{ version }}"/' props/add-version-flag.patch | patch -p1`

sanchitram1: I had reworked it locally to use the version that was set in the Makefile as 1.0.0.  However, since we are using Calver, I can make sure it pull it from the calver listed in the package.yaml.

PR #1122, opened 2023-03-30
+openvpn

sanchitram1:
- +openvpn Without pkcs11-helper, still works


jhheider:
looks like you have deps using openssl ^1.1. will openvpn compile with it? most things that are up-to-date will use either (but tea won't put both in an single environment).

sanchitram1:
How would you force a dependency to use `openssl~3`?  I could edit the dependency's .yml file (obviously making sure that it is compatible with ~3?

I could also see if `openvpn` compiles with ~1.1 -- that way, there's no conflicts.  I changed that locally, and it built.

jhheider:
Right, I think you need to use openssl^1.1 for openvpn. The alternative is, for any package that supports it, to make their version string '^1.1 || 3'. Then things that _can_ use either will, and any that have strict requirements will lock to theirs.

sanchitram1:
Made the change, but think this is still failing because it contains commits from `mop-tracker` as well **sigh**.

Removed `mop-tracker` from it, and re-checking to see if the builds & test work.

jhheider:
Not sure about the library it's complaining about, but brew includes:
```ruby
  on_linux do
    depends_on "linux-pam"
    depends_on "net-tools"
  end
```
My guess is it's `net-tools`.

sanchitram1:
Only for linux though?  So this would be:
```yaml
dependencies:
  tea.xyz/gx/cc: c99
  tea.xyz/gx/make: '*'
  freedesktop.org/pkg-config: ^0.29
  linux/aarch64:
    net-tools: '*'
  linux/x86-64:
    net-tools: '*'
```

sanchitram1:
(have to confirm whether it's '*', but assuming it is)

jhheider:
> Only for linux though? So this would be:
> 
> ```yaml
> dependencies:
>   tea.xyz/gx/cc: c99
>   tea.xyz/gx/make: '*'
>   freedesktop.org/pkg-config: ^0.29
>   linux/aarch64:
>     net-tools: '*'
>   linux/x86-64:
>     net-tools: '*'
> ```

```yaml
...
  linux:
    sourceforge.net/net-tools: '*'
```

Should be correct, if it is `net-tools`, once it's packaged.

jhheider:
@sanchitram1 net-tools has been packaged: https://github.com/teaxyz/pantry/pull/1922

sanchitram1:
Looking at it today!

sanchitram1:
Per homebrew, this depends on `libcap-ng`, `libnl`, `linux-pam`, and `net-tools` for linux.  #1999 covers `libnl`.

sanchitram1:
2 down, 2 more to go.  `libcap-ng` (#2030) looks straightforward, i'll give it a go.

`linux-pam`, acc. to homebrew, seems more intense.  Per its script, it depends on:
```rb
depends_on "libnsl"
depends_on "libprelude"
depends_on "libtirpc"
depends_on "libxcrypt"
```

When I built the homebrew top 300, seems like I was missing some linux-specific dependencies when constructing the complete graph.  I don't think we have these packaged?

jhheider:
If you want to publish what you have now, you can always make it
```yaml
platforms:
  - darwin
  # FIXME: linux support will require the following additional packages:
  # libnsl, libprelude, libtirpc, libxcrypt
```

sanchitram1:
Donezo

PR #1186, opened 2023-03-31
fix(git) (and others)

### LINKED ISSUE #1185: ❌ build issues: git-scm.org
Running log of build failures for git-scm.org

jhheider:
curl.se should depend on curl.se/ca-certs

Currently, curl gets ca-certs as a transitory dependency from openssl (and git likewise), but I don't think we should depend on that chaining. Moreover, git needs to know where to _find_ ca-certs, since it doesn't look like it uses the `SSL_CERT_FILE` envvar.

Requesting review for changing deps (cascade risk).

### DIFF
--- a/projects/curl.se/package.yml
+++ b/projects/curl.se/package.yml
@@ -8,6 +8,7 @@ versions:
 
 dependencies:
   openssl.org: ^1.1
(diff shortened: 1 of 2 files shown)


mxcl:
good spots

PR #1216, opened 2023-04-02
+stripe

leonsilicon:
The reason why the tests fail is because running `stripe --version` will display `stripe version master` because of https://github.com/stripe/stripe-cli/blob/master/pkg/version/version.go#L19, but I'm not sure how to work around that.

### DIFF
--- a/projects/stripe.com/package.yml
+++ b/projects/stripe.com/package.yml
@@ -20,7 +20,7 @@ build:
 
   env:
     LDFLAGS:
-      [-s, -w, "-X=main.Version={{version}}"]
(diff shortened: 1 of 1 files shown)


jhheider:
Try:
```sh
sed -i.bak -e 's/^var Version = "master"$/var Version = "{{ version }}"/' pkg/version/version.go
rm pkg/version/version.go.bak
```

jhheider:
Actually, reading the comment in that file, and https://github.com/stripe/stripe-cli/blob/master/.goreleaser/mac.yml, it looks like `-X github.com/stripe/stripe-cli/pkg/version.Version={{ version }}` (or similar; I don't know if that's a logical path or a file path) is cleaner.

jhheider:
Great!

PR #1247, opened 2023-04-04
+gops

sanchitram1:
- Any version go.dev for building & running
- Test is to print the go process that is used to run gops.


projects/mop-tracker/version.patch
```diff
@@ -0,0 +1,49 @@
+diff --color -u a/Makefile b/Makefile
```
jhheider: This file probably doesn't go in this pr...

projects/github.com/google/gops/package.yml
```diff
@@ -0,0 +1,43 @@
...
+# contents of both:
+#
+# * https://github.com/teaxyz/pantry
+#
+# as a learning guide and reference. For example, if you are package a `go`
+# project, then browse the listings and find a go package.
+
+
+# this is the source tarball. Swap the version out for {{version}} or we won’t
+# be able to upgrade the package when it updates!
```
jhheider: I'd pull out the template comments, but this looks good otherwise.

jhheider:
Bam! Thanks, @sanchitram1 !

PR #1253, opened 2023-04-04
Add git subtree support to git-scm.org

leonsilicon:


jhheider:
Nice. Looks like the test needs a reword, but this a good addition.

jhheider:
Looking good. Thanks, @leondreamed .

PR #1276, opened 2023-04-06
+whisper.cpp

sanchitram1:
- is `mv $SRCROOT/examples {{prefix}}/share` needed?  It contains use cases.
- what if you want to use gpt4all instead of llama?
- currently breaks at fetch model at `tea +sentencepiece upgrade-model.py 7B`. I get the same error when trying `llama.cpp --help`
```sh
Traceback (most recent call last):
  File "/Users/sanch/.tea/github.com/ggerganov/whisper.cpp/v1.2.1/tbin/upgrade-model.py", line 7, in <module>
    from sentencepiece import SentencePieceProcessor
  File "/Users/sanch/.tea/google.com/sentencepiece/v0.1.97/lib/python3.11/site-packages/sentencepiece/__init__.py", line 13, in <module>
    from . import _sentencepiece
ImportError: cannot import name '_sentencepiece' from partially initialized module 'sentencepiece' (most likely due to a circular import) (/Users/sanch/.tea/google.com/sentencepiece/v0.1.97/lib/python3.11/site-packages/sentencepiece/__init__.py)
```


mxcl:
> is mv $SRCROOT/examples {{prefix}}/share needed? It contains use cases.

You can do it for sure.

> what if you want to use gpt4all instead of llama?

Is that possible? I figured this would be finely dependent on its own model/

> currently breaks at fetch model at tea +sentencepiece upgrade-model.py 7B. I get the same error when trying llama.cpp --help

Is it the same magnet? In which case the model fetch script really should be its own pkg or this should depend on llama.cpp. We don’t repeat ourselves in pkging.

And… well we need to find a new torrent that works. The one I used worked for me, then for nobody else and now doesn't work for me either.

sanchitram1:
You're right about `gpt4all`, I had mis-read the models section on their repo.  

`whisper.cpp` provides shell scripts with `./models` to retrieve the model that you want to use.  I used `base.en`, but you can get tiny, small, medium, or large.  We could (should) ask the user which one they want to use.

Where do I place the model?  Right now, it's in `tbin`, but whenever you call whisper.cpp, you'd need to tell it where the model is? 

edit: Can we change the default directory it looks for the model within the build script?

sanchitram1:
Ok, on reviewing gpt4all & llama.cpp thoroughly, I understand what you're placing in the bin vs. the tbin.

The build products we obtain from make are placed in the `tbin`.  The `props` contain the actual executable placed in the bin, that just calls the build product we put in `tbin`.  It's here we can specify params like where the model is located.  `tbin` also contains the `model-fetch` script from `props`, where we can figure out what model the someone wants to use, and whether we have it.

mxcl:
> whisper.cpp provides shell scripts with ./models to retrieve the model that you want to use. I used base.en, but you can get tiny, small, medium, or large. We could (should) ask the user which one they want to use.

I would start by just fetching whatever most people prefer (will take some research). Later we can consider a way to choose

sanchitram1:
Illegal instruction?

mxcl:
> Illegal instruction?

It crashed. Check for a newer version of whisper.cpp, this version is broken.

sanchitram1:
Needed sdl2 for whisper.command and whisper.stream, so it included all the previous commits until it was added.

There are no tests for whisper.command & whisper.stream, but both do work.  What's the best way to perform multiple tests within a script?

magnusviri:
I just tried it and it's really cool. I'm eager to build something that dings every time I say "um".

magnusviri:
> There are no tests for whisper.command & whisper.stream, but both do work. What's the best way to perform multiple tests within a script?

Run them with -h (since they don't have -v).

magnusviri:
I'm seriously thinking the AI community needs a tool to download models. I've been working on git-lfs so I can download files from hugging face. Being able to easily download models is actually my number one goal right now. I don't think I've got time to take on such a tool. This PR downloaded the whisper model without any trouble at all.

sanchitram1:
Yet I'm still getting an Illegal Instruction...

Seems like others using whisper had the issue, and they suggested a fix, which still makes this action fail.  I posted about it there [here](https://github.com/ggerganov/whisper.cpp/issues/597#issuecomment-1515613431).  

For the download model -- whisper has a file `download-ggml-model.sh` that does a great job of interfacing with hugging face.

mxcl:
I'm finishing this up.

Also changing the model download directory to match alpaca.cpp, not sure why you set it to be inside the pkg prefix, that would mean it would have to redownload the model every time the pkg updates.

Changing the main binary to be `whisper.cpp` which accepts `command` or `stream` as sub-commands. This means we deviate from the README for the project, but you cannot take `main` as your binary name lol. and `whisper.main` which idiomatic is also not imo the right choice for the project.

This means in order to use our whisper.cpp you have to be savvy to how it works, so we'll need to figure out a solution for that longer term.

sanchitram1:
Thanks for closing it out, and for a couple of the comments, and [this](https://github.com/teaxyz/pantry/blob/581dd709d9631c3e2ca4d364686121ee90177822/projects/github.com/ggerganov/whisper.cpp/whisper-fetch#L14).  Guessing this is why I was running into the Illegal Instruction error?

I agree on the savviness being a pre-requisite.  Running `whisper.cpp --help` does cover most of the use cases.

PR #1277, opened 2023-04-06
fix(+python-poetry.org): use `version` variable in url

relsunkaev:
Currently the url is hardcoded which results in installing the same version (1.3.1) even if the cli is indicating that it's installing some other version.

The codespace also auto-formatted the yaml, hope that's okay.

### DIFF
--- a/projects/python-poetry.org/package.yml
+++ b/projects/python-poetry.org/package.yml
@@ -12,11 +12,12 @@ build:
   dependencies:
     cmake.org: '*'
     linux/x86-64:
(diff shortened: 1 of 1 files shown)


jhheider:
once this is merged, i'll rebuild the other versions we have up.

jhheider:
1.3.2, 1.4.0, and 1.4.2 all built. you might need to `rm -rf ~/.tea/python-poetry.org ~/.tea/tea.xyz/var/www/python-poetry.org*`.

PR #1279, opened 2023-04-06
+mozilla.org/mozjpeg

magnusviri:


magnusviri:
AI did most of the work!

magnusviri:
My computers is arm, which is why the Linux test passed.

jhheider:
Yup, me too. Every time.

PR #1299, opened 2023-04-08
gha tidy

mxcl:


mxcl:
Tidy. Primary goal was to make PR status checks more readable by getting rid of all the skipped steps due to new-version flow.

Actionifies complain. I would like to do the same with build and bottle but don't dare. Then they could move to brewkit which is the end goal to make it so anyone can host a pantry and use our brewkit to run builds (on their own hardware probs tho).

I dunno if the `new-version.yml` will run. I'm not sure if you can have secrets passed to workflow-calls

jhheider (approved):
Looks reasonable. We'll fix it if it breaks.

PR #1311, opened 2023-04-09
+libedit

mxcl:


jhheider:
~Think this already exists: https://github.com/teaxyz/pantry/blob/main/projects/thrysoee.dk/editline/package.yml~

Derp. you're clearly editing it. I return to my Sunday.

mxcl:
hmm no I didn't intend to edit it. I did a search and didn't find it so thought it wasn't there. Thanks and whoops.

PR #1313, opened 2023-04-10
+aspell.net

magnusviri:
`echo hoopy eastr | aspell -a`

Includes all dictionaries except aspell-nb-0.50.1-0 and aspell-is-0.51.1-0, which don't compile (and aren't part of the brew formula either).

AI assisted.

This could also be saved as gnu.org/aspell. I picked aspell.net because it seems you guys favor unique domains if they exist.

https://github.com/teaxyz/pantry/issues/315


magnusviri:
Strange the tests are failing. They worked on my computer.

jhheider:
it's the relocatibility; the tests are always done with a different $TEA_PREFIX than the build to ensure it doesn't hardcode the prefix. the above with `ASPELL_CONF` should work.

PR #1316, opened 2023-04-10
+sdl2

sanchitram1:
- +whisper.cpp
- base model & build
- added props
- fix-up tbin file
- +sdl2


jhheider:
I'm absolutely shocked SDL will build without packaging 12,000 things. Good job!

sanchitram1:
Both brew & its own docs mentioned requiring `lib*`, but it built perfectly with `pkg-config`, `autoconf`, and `automake`.  It's needed for more fun use cases of `whisper.cpp`, so i'm going to take this merge & see what happens!

jhheider:
my _guess_ is that you'll still need those libraries to render anything, but see how it goes.

PR #1318, opened 2023-04-11
+scons

### LINKED ISSUE #525: +scons (85/548)
Formulae: https://formulae.brew.sh/formula/scons

mxcl:
Closes #525

### DIFF
--- a/projects/scons.org/package.yml
+++ b/projects/scons.org/package.yml
@@ -15,6 +15,8 @@ provides:
   - bin/scons
 
 test:
+  dependencies:
+    tea.xyz/gx/cc: c99
   script: |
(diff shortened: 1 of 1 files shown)


mxcl:
> # Why doesn't SCons find my compiler/linker/etc.?
> 
> I can execute it just fine from the command line.
> A common problem for new users is that SCons can't seem to find a compiler, linker, or other utility that > they can run just fine from the command line. This is almost always because, by default, SCons does not use the same PATH environment variable that you use from the command line, so it can't find a program that has been installed in a "non-standard" location unless you tell it how. Here is the explanation from the SCons man page:

lol great

PR #1329, opened 2023-04-11
+blake2

### LINKED ISSUE #268: +libb2 (108/548)
Formulae: https://formulae.brew.sh/formula/libb2

This is blake2, there's also BLAKE3 :)

mxcl:
Closes #268


mxcl:
yeah no clue for these two different failures

jhheider:
interesting that it's both x86-64 though, especially with the error on linux.

jhheider:
> CONSIDER USING BLAKE3, faster than BLAKE2, see https://github.com/BLAKE3-team/BLAKE3

> BLAKE2 comes in two flavors:
> - BLAKE2b (or just BLAKE2) is optimized for 64-bit platforms—including NEON-enabled ARMs—and produces digests of any size between 1 and 64 bytes
> - BLAKE2s is optimized for 8- to 32-bit platforms and produces digests of any size between 1 and 32 bytes

jhheider:
And the last release was 4 years ago, and the issues are looking a little stale.

mxcl:
I guess we’ll hold off on this one until something needs it

PR #1334, opened 2023-04-11
+crates.io/git-delta

jasonmorganson:


projects/crates.io/git-delta/package.yml
```diff
@@ -0,0 +1,24 @@
...
+
+provides:
+  - bin/delta
+
+versions:
+  github: dandavison/delta/tags
+
+dependencies:
+  darwin:
+    zlib.net: ^1
```
mxcl, line 13: it only needs zlib or darwin? This seems unlikely.

jasonmorganson: In my testing zlib was only needed on darwin.

PR #1335, opened 2023-04-12
+crates.io/gitui

jasonmorganson:


projects/crates.io/gitui/package.yml
```diff
@@ -0,0 +1,26 @@
...
+  - bin/gitui
+
+versions:
+  github: extrawurst/gitui/tags
+  strip: /v/
+
+build:
+  dependencies:
+    perl.org: '*'
+    openssl.org: '*'
```
mxcl: almost certainly is a runtime dependency. Move to the root `dependencies:` node.

mxcl:
possibly it is building openssl itself rather than just the rust bindings:

https://github.com/teaxyz/pantry/actions/runs/4679118418/jobs/8288721015?pr=1335#step:7:437

which suggests it is not finding our openssl… :/

mxcl:
nice find. What we need is a “if this is a dependency and this is a dependency add this flag” system 🤔

PR #1339, opened 2023-04-12
Dark magic providers

mxcl:


jhheider (approved):
Looks good. Not worth doing a providers directory at the top level to avoid having to walk all the projects looking for providers? I guess we walk them all for other things in reasonable time, but I always worry about all the disk access.

mxcl:
Removed the deps part. If they have deps they need a full `package.yml` entry.

> Not worth doing a providers directory at the top level to avoid having to walk all the projects looking for providers? I guess we walk them all for other things in reasonable time, but I always worry about all the disk access.

We can build a cache layer when performance gets bad. I prefer to have these files in the namespaced directories.

mxcl:
the python entries are already packaged. I would delete the `package.yml`s in this PR too but I'm not sure of the consequences to this. Probs it just means the installed entries become stale.

I think our rules for these providers currently are that they cannot require any other deps. Thought that we may relax that over time.

Certainly I think at this time if the `pipx` command would require eg. `cc` to build we should package it.

jhheider:
> I think our rules for these providers currently are that they cannot require any other deps. Thought that we may relax that over time.

That seems reasonable.

PR #1343, opened 2023-04-12
Updated github.com/ggerganov/llama.cpp to 2023.04.11 8b67998

magnusviri:
And I'm copying quantize and python files to tbin, which should probably be bin... I haven't figured out how to use those tools yet.


mxcl:
> And I'm copying quantize and python files to tbin

these require a substantial virtual-env, which is why I didn't bundle them before. it made the linux pkg >2GB!

mxcl:
A bigger issue here is getting the model. I believe it’s on hugging face.

I'd prefer that we found an IPFS node that was healthy. hugging face maybe don't care about their bandwidth costs but I don't want tea to contribute.

PR #1344, opened 2023-04-12
+github.com/ggerganov/ggml

magnusviri:
This is required by github.com/ravenscroftj/turbopilot but I'm not sure if this is the best way to do it.


jhheider:
You can pull a specific commit as a tarball: https://github.com/teaxyz/pantry/blob/main/projects/github.com/jart/blink/package.yml

PR #1350, opened 2023-04-13
+git-lfs.com

magnusviri:
WIP. AI assisted. I'm working on ronn because I think that's required for the man page.

### DIFF
--- a/projects/git-lfs.com/package.yml
+++ b/projects/git-lfs.com/package.yml
@@ -1,0 +1,37 @@
+distributable:
+  url: https://github.com/git-lfs/git-lfs/releases/download/v{{version}}/git-lfs-v{{version}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: git-lfs/git-lfs/releases/tags
+
+
+build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+    go.dev: '*'
+#     ruby.org: '*'
+#     asciidoctor.org: '*'
+#     ronn.github.com: '*'
+  script: |
+    make --jobs {{ hw.concurrency }}
+    mkdir -p {{ prefix }}
+    mv bin {{ prefix }}/bin
+#     make man RONN={{prefix}}/bin/ronn || true # Ignore failure
+#     make test
+
+#   env:
+#     GIT_LFS_SHA: ''
+#     VERSION: ${{version}}
+
+
+
+provides:
+  - bin/git-lfs
+
+test:
+  script: |
+    git init
+    git lfs track test
+    git lfs version
--- a/projects/git-lfs.com/package.yml
+++ b/projects/git-lfs.com/package.yml
@@ -5,30 +5,30 @@ distributable:
 versions:
   github: git-lfs/git-lfs/releases/tags
 
+provides:
+  - bin/git-lfs
+
+dependencies:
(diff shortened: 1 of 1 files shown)


magnusviri:
https://github.com/teaxyz/pantry/issues/324

magnusviri:
I just learned how to use git reset --HARD and cherry-pick to remove commits that shouldn't be in the branch. So now this only contains git-lfs.com.

magnusviri:
I don't know how to fix this. I can replicate the Linux crash in Docker, but debugging isn't my strong point, and debugging on Linux is a complete mystery to me. I guess I need to generate a core dump and open it with gdb.

magnusviri:
I don't know what is going on. It works when in the build directory but not when installed. I suspect tea is changing the binary, but I'm not sure.

Here's the stack trace.

```
(gdb) run
Starting program: /root/.tea/git-lfs.com/v3.3.0/bin/git-lfs
warning: Error disabling address space randomization: Operation not permitted

Program received signal SIGSEGV, Segmentation fault.
0x0000ffff9f0c3520 in dl_main (phdr=<optimized out>, phnum=<optimized out>, user_entry=<optimized out>, auxv=<optimized out>) at rtld.c:1588
1588	rtld.c: No such file or directory.
(gdb) bt
#0  0x0000ffff9f0c3520 in dl_main (phdr=<optimized out>, phnum=<optimized out>, user_entry=<optimized out>, auxv=<optimized out>) at rtld.c:1588
#1  0x0000ffff9f0d497c in _dl_sysdep_start (start_argptr=start_argptr@entry=0xfffffbb6e410, dl_main=dl_main@entry=0xffff9f0c20c8 <dl_main>) at ../elf/dl-sysdep.c:253
#2  0x0000ffff9f0c1884 in _dl_start_final (arg=arg@entry=0xfffffbb6e410, info=info@entry=0xfffffbb6df98) at rtld.c:415
#3  0x0000ffff9f0c1b00 in _dl_start (arg=0xfffffbb6e410) at rtld.c:524
#4  0x0000ffff9f0c1048 in _start () from /lib/ld-linux-aarch64.so.1
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

Installed version crashes.

```
tea $ ls -al /root/.tea/git-lfs.com/v3.3.0/bin/git-lfs
-rwxr-xr-x 1 root root 10619184 Apr 24 00:14 /root/.tea/git-lfs.com/v3.3.0/bin/git-lfs
tea $ /root/.tea/git-lfs.com/v3.3.0/bin/git-lfs
Segmentation fault
```

Build dir version runs. It's slightly smaller (10619184 vs 10616832).

```
tea $ ls -al ./builds/git-lfs.com-3.3.0+linux/bin/git-lfs
-rwxr-xr-x 1 root root 10616832 Apr 24 00:13 ./builds/git-lfs.com-3.3.0+linux/bin/git-lfs
tea $ ./builds/git-lfs.com-3.3.0+linux/bin/git-lfs
git-lfs/3.3.0 (GitHub; linux arm64; go 1.20.3; git 3.3.0)
git lfs <command> [<args>]
```

jhheider:
That's possibly a consequence of `fix-elf.ts` in brewkit. `readelf -d /path/to/git-lfs` should show the RPATH and other linking data, and `ldd /path/to/git-lfs` should show if any of the libraries aren't being found. Very curious.

PR #1365, opened 2023-04-13
+crates.io/zellij

jasonmorganson:


projects/crates.io/zellij/package.yml
```diff
@@ -0,0 +1,20 @@
+distributable:
+  url: https://github.com/zellij-org/zellij/archive/refs/tags/v{{version}}.tar.gz
+  strip-components: 1
+
+provides:
+  - bin/zellij
+
+versions:
+  github: zellij-org/zellij/tags
```
mxcl: do `zellij-org/zellij` whih will use releases/tags which is better for 2 reasons:
1. we wont try to build random tags
2. we will build updates instantly

mxcl (changes requested):

projects/crates.io/zellij/package.yml
jasonmorganson: Got it

PR #1423, opened 2023-04-15
Provide helix runtime

jasonmorganson:
> set the HELIX_RUNTIME environment variable to point to the runtime files

See https://docs.helix-editor.com/install.html#configuring-helixs-runtime-files

### DIFF
--- a/projects/helix-editor.com/package.yml
+++ b/projects/helix-editor.com/package.yml
@@ -4,7 +4,6 @@ distributable:
 
 provides:
   - bin/hx
-  - runtime
 
 versions:
   github: helix-editor/helix/tags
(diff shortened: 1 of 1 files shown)


jhheider:
This looks odd. From the test run, it doesn't look like there's a binary or anything else called `runtime` in the root of the prefix: https://github.com/teaxyz/pantry/actions/runs/4707738330/jobs/8349841990?pr=1423

Do we need to copy files out of the srcdir? That would probably look like this:
```sh
mkdir -p "{{prefix}}"/share
cp -a runtime "{{prefix}}"/share
...

runtime:
  env:
    HELIX_RUNTIME: ${{prefix}}/share/runtime
```

jasonmorganson:
Thanks @jhheider. That is what's needed.

PR #1464, opened 2023-04-16
+openblas

mfts:
there might be an issue with the symlinking of the include/openblas because it's a nested folder with the same name


jhheider:
3/4 of the platforms should be enough to start. closes https://github.com/teaxyz/pantry/issues/290

PR #1549, opened 2023-04-19
+swiftlint

pakko972:


projects/github.com/realm/SwiftLint/package.yml
```diff
@@ -0,0 +1,25 @@
...
+platforms:
+  - darwin
+
+provides:
+  - bin/swiftlint
+
+warnings:
+  - vendored
+
+build:
```
mxcl, line 17: probs: `working-dir: build` would be the easiest way around these issues.

At some point I'll make using vendor compiled binaries more official. Though I would like the binaries to be GPG signed and few people do that.

mxcl:
ugh @jhheider is this file not there? https://github.com/teaxyz/pantry/actions/runs/4755578669/jobs/8449945551?pr=1549#step:8:41

jhheider:
```sh
$ ls /usr/lib/swift/
libswiftCreateML.dylib		libswiftRemoteMirror.dylib
$ swift --version
Apple Swift version 5.7.2 (swiftlang-5.7.2.135.5 clang-1400.0.29.51)
Target: x86_64-apple-darwin22.3.0
```

pakko972:
@mxcl @jhheider based on their `Package.swift`, SwiftLint needs macOS 12+.
Is there a way to specify this in the package.yml (maybe `platform`)?

jhheider:
we don't really have a method for platform restriction at this time. _however_, have you tried:
```sh
sed -i.bak -e 's/macOS(\.v12)/macOS(.v11)/' Package.swift
rm Package.swift.bak
```
to see if it works otherwise?

mxcl:
ah, well this would be the first package that has a macOS requirement we can’t work around. Fortunately adding platform restrictions for macOS is pretty easy relatively so makes it easy to spearhead this feature.

mxcl:
@jhheider is the intel mac system on macOS 11 then? Seems like we should keep our macOS builders in sync for platform. Fortunately it's easy to target older macOS by simply setting `MACOS_DEPLOYMENT_TARGET` which we do

mxcl:
@jhheider can you formalize `platforms:` so that here we don’t build excluded platforms, so people can specify that and get green checks? Depending on your workload

jhheider:
Omitting platforms is needed, but I don't think that would fix this PR, unless we omitted Mac altogether (since we "require" macOS 11 compliance).

You could try overriding the envvar to force macOS compliance; I believe we might have a package or two that does that. It should warrant a warning: key

mxcl:
> Omitting platforms is needed, but I don't think that would fix this PR, unless we omitted Mac altogether (since we "require" macOS 11 compliance).

We don’t support prior to macOS 11. We don’t require it of packages that require > 11. It's just until now we haven't had that.

jhheider:
I guess this is the only package.yml where we change the target: https://github.com/teaxyz/pantry/blob/2f050b71b0e41eb1d67c91640e48705827ddf922/projects/github.com/westes/flex/package.yml#L20-L30

In theory, we could bump the target in this package.yml, but some users might get caught out.

pakko972:
This trick won't work because the latest versions needs the `SwiftConcurrency` system library to be available, and this is not the case with macOS < 12.

This requirement has been introduced by [SwiftLint 0.49.0](https://github.com/realm/SwiftLint/releases/tag/0.49.0), meaning [SwiftLint 0.48.0](https://github.com/realm/SwiftLint/releases/tag/0.48.0) is compatible with macOS 11.

Would it be reasonable to install 0.48.0 for macOS 11 ?

mxcl:
> Would it be reasonable to install 0.48.0 for macOS 11 ?

That would progress us further, but I think we can install the latest macOS on the builders since targeting backwards is easy enough.

The problem is testing. Ideally we would have separate runners for testing or use one of the hacks to run macOS 11 in a docker image. eg. https://github.com/sickcodes/Docker-OSX and then be able to spin up test instances for whatever the package supports (eg. here we would only test against 12 and 13)

mxcl:
Apparently our builders are macOS 12 and macOS 13, so I don’t see how we aren't finding this library.

jhheider:
The file isn't present on either system, nor on my Mac Studio which is all the way current. It looks like it may require Xcode be installed, which we've actively resisted so we don't pollute the build machines.

mxcl:
We’re going to need Xcode. It'll pollute the build machines but I can't see an alternative, some shit just needs Xcode, eg. openjdk, which we must package.

I expect only small amounts of bad consequence. I just wanted to avoid it if poss.

jhheider:
I can add them this week.

jhheider:
Interestingly, it was building just fine, but not testing (on macOS 11; obviously). Put a trap in `get-platform.ts` so we can support things that won't test that far back.

PR #1583, opened 2023-04-20
	new file:   projects/php.net/package.yml

and-ri:


projects/php.net/package.yml
```diff
@@ -0,0 +1,38 @@
+distributable:
+  url: https://github.com/php/php-src/archive/refs/tags/php-{{ version }}.tar.gz
```
mxcl, line 2: the tagged tarballs typically do not contain `./configure` I'd looking for the official downloads.

Also recommend running `pkg build` locally before submitting as debug cycles will be way faster for you!

Finally: super glad to have you working on php. thank you.

PR #1599, opened 2023-04-20
+stable-diffusion-webui

mxcl:


jhheider:
This was merged, but never worked on linux. Does this package need to be mac only, @mxcl?

mxcl:
@jhheider yes, though we don’t have an official way to mark that yet right? We have added `platform: darwin` a few places, but nothing actually reads it.

jhheider:
Right, thus far we've had to do stuff like this:

https://github.com/teaxyz/pantry/blob/main/projects/freedesktop.org/libbsd/package.yml

mxcl:
like we made it so platforms that fail don't prevent bottles uploading months back right? so your concern is because it creates tickets to fix them or…?

Either way proper platform whitelisting would fix it.

jhheider:
We undid that a couple of weeks ago with: https://github.com/teaxyz/pantry/pull/1299/files

Since we moved the matrix inside each job, instead of outside, if any fail to build, all fail to bottle.

mxcl:
yeah ok that's a regression that wasn't intentional, but hopefully the new available system will fix that too?

jhheider:
Yep, a better fix.

jhheider:
I think as of brewkit 0.23.2 and attendant pantry changes, this is handled. building it with proper platform whitelisting now.

PR #1614, opened 2023-04-20
+mcmc-jags.sourceforge.io

magnusviri:
I need help. I don't understand this error. When I run `pkg build` I get this.

```
checking for gcc... gcc
`checking whether the C compiler works... no
configure: error: in `/Users/u0076374/Library/CloudStorage/Dropbox/Github/Tea/repos/pantry_mag/builds/mcmc-jags.sourceforge.io-4.3.2+darwin':
configure: error: C compiler cannot create executables
See `config.log' for more details
```

I'm wondering if I should run autoconf.

When I try to build with Linux (`pkg -L build`) I get this.

```
Download https://deno.land/x/cliffy@v0.25.7/ansi/mod.ts
error: error sending request for url (https://deno.land/x/cliffy@v0.25.7/flags/mod.ts): error trying to connect: dns error: failed to lookup address information: No address associated with hostname
    at file:///opt/tea/tea.xyz/brewkit/v0.22.0/libexec/sort.ts:3:28
```

### DIFF
--- a/projects/mcmc-jags.sourceforge.io/model.bug
+++ b/projects/mcmc-jags.sourceforge.io/model.bug
@@ -1,0 +1,7 @@
+data {
+  obs <- 1
+}
+model {
+  parameter ~ dunif(0,1)
+  obs ~ dbern(parameter)
+}
--- a/projects/mcmc-jags.sourceforge.io/package.yml
+++ b/projects/mcmc-jags.sourceforge.io/package.yml
@@ -3,16 +3,24 @@ distributable:
   strip-components: 1
 
 versions:
-  - 4.3.2
+  url: https://sourceforge.net/projects/mcmc-jags/files/JAGS/4.x/Source/
+  match: /JAGS-\d+\.\d+\.\d+\.tar\.gz/
+  strip:
+    - /^JAGS-/
+    - /\.tar\.gz$/
 
 build:
   dependencies:
     tea.xyz/gx/cc: c99
     gnu.org/gcc: '*'
     tea.xyz/gx/make: '*'
-  script: |
-    ./configure $ARGS
-    make --jobs {{ hw.concurrency }} install
+  script:
+    # - ./configure $ARGS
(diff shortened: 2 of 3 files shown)


magnusviri:
Hm, strange. My computer gets a different error than the GitHub runners.

The GitHub runners are erroring because tea's gcc doesn't include fortran. The [homebrew version of gcc](https://github.com/Homebrew/homebrew-core/blob/70339ad7a5d077833c43b33e99c7d1a00e25295a/Formula/gcc.rb) includes fortran.

magnusviri:
I don't get the GHA error on my computer (macOS and Linux). And I have the file that is missing on the GH runners.

```
+ jags script
/Users/builder/.tea/mcmc-jags.sourceforge.io/v4.3.2/bin/jags: line 39: /opt/mcmc-jags.sourceforge.io/v4.3.2/libexec/jags-terminal: No such file or directory
/Users/builder/.tea/mcmc-jags.sourceforge.io/v4.3.2/bin/jags: line 39: exec: /opt/mcmc-jags.sourceforge.io/v4.3.2/libexec/jags-terminal: cannot execute: No such file or directory
Error: Process completed with exit code 126.
```

jhheider:
Tests are run under a different prefix than build. So, this means there's a hard-coded path to the build directory in the output somewhere.

magnusviri:
How do I fix this?

jhheider:
Generally, you'd need to figure out where the hardcoded path is, then determine how to fix it up. _Usually_, they're hardcoded into shell/python scripts, so you can use `sed` to fix them, like https://github.com/teaxyz/pantry/blob/97933f2bf74249eeae1aa5494bedc1c90544d77c/projects/haskell.org/package.yml#L78-L81 . If it's baked into a compiled binary, it get much more difficult; openssl has this, for example, to fix its default certfile search path: https://github.com/teaxyz/pantry/blob/main/projects/openssl.org/x509_def.c.diff

jhheider:
#backlog

PR #1624, opened 2023-04-21
Re2c.org

and-ri:


projects/re2c.org/package.yml
```diff
@@ -0,0 +1,32 @@
+distributable:
+  url: https://github.com/skvadrik/re2c/archive/refs/tags/{{ version.raw }}.tar.gz
```
mxcl: use the official release tarballs: https://github.com/skvadrik/re2c/releases/download/3.0/re2c-3.0.tar.xz we prefer this since this is what the maintainer intends for us to use to build from source. We only use the autogenerated github sources are only used if necessary.

this also means you can drop autoconf/make and autogen.sh

projects/re2c.org/package.yml
```diff
@@ -0,0 +1,32 @@
...
+  env:
+    ARGS:
+      - --prefix="{{prefix}}"
+
+provides:
+  - bin/re2c
+  - man/man1/re2c.1
+
+test:
+  re2c --version
```
mxcl: do a `| grep {{version}}` for thoroughness

mxcl (changes requested):

and-ri:
Sorry about that, I'm still new to this👩🏻‍💻 I already changed the file but now something is wrong with the test

mxcl:
no need to apologize! your contribution is sincerely appreciated.

I imagine the output from rec2c is not expected. Let me try it locally…

mxcl:
```sh
$ re2c --version
re2c 3.0
```

yeah so we need to check for version.marketing, I'll push

mxcl:
Added https://github.com/teaxyz/pantry/pull/1624/commits/c87e40a703efc6f950e370ff3c2e7230a923539e

mxcl (approved):

and-ri:
Thank you for your help! Now I have become a little smarter 🤓

PR #1635, opened 2023-04-21
allow platform whitelist in yaml

jhheider:
needs https://github.com/teaxyz/brewkit/pull/87
and https://github.com/teaxyz/brewkit/pull/89

### DIFF
--- a/.github/workflows/bottle.yml
+++ b/.github/workflows/bottle.yml
@@ -10,6 +10,9 @@ on:
       platform:
         required: true
         type: string
+      projects:
+        required: false
+        type: string
     outputs:
       pr:
         description: 'The PR number'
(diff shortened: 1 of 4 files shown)


mxcl:
Seems like get-platform could determine this and shortcircuit all the other jobs…?

mxcl:
That'll require a brewkit change again tho, so up to you. It would be cleaner, but not required.

mxcl (approved):

jhheider:
Hm, usePantry.available uses Deno.build.os, so it's specific to the platform being run on. This is a good idea, but it involves a more complex check in cli.

jhheider:
But I have ideas.

jhheider:
Screenshot 2023-04-24 at 12 52 37
not bad.

PR #1637, opened 2023-04-22
gnu.org/gcc add fortran

magnusviri:
https://github.com/teaxyz/pantry/pull/1614 needs fortran. I added it to gcc, mimicking how homebrew does it.


jhheider:
very nice!

PR #1668, opened 2023-04-25
+k3d.io

branchv:
Adds [k3d](https://github.com/k3d-io/k3d)


jhheider:
thanks @branchvincent !

PR #1699, opened 2023-04-27
+yadm.io

branchv:
Adds [yadm](https://github.com/TheLocehiliosan/yadm)


jhheider:
thanks, @branchvincent !

PR #1706, opened 2023-04-27
gource + its deps

### LINKED ISSUE #562: +sdl2 (212/548)
Formulae: https://formulae.brew.sh/formula/sdl2

### LINKED ISSUE #590: +glew (272/548)
Formulae: https://formulae.brew.sh/formula/glew

mxcl:
Closes #590
Closes #562

### DIFF
--- a/projects/glew.sourceforge.io/package.yml
+++ b/projects/glew.sourceforge.io/package.yml
@@ -9,6 +9,9 @@ versions:
     - /tr title="/
     - /" class="folder/
 
+# it will work on other platforms, but we couldn't get it to work
+platforms: darwin/aarch64
+
 build:
   dependencies:
     tea.xyz/gx/cc: c99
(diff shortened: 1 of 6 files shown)


jhheider:
Looks like this uses mesa-glu > mesa on linux with homebrew for OpenGL; mesa-glu just deps mesa, and mesa has a ton of deps, but they look like we've already packaged them.

mxcl:
@jhheider not sure how you feel about this per-se, but I marked only glew as available `darwin/aarch64` which means it is impossible for `gource` to build on anything but the same.

Really that restriction should fall through. I fear though that practically that's just a nightmare of complexity we can easily avoid by manually propogating the YAML. So I'll do that for now.


---

Just being lazy, the failure for darwin/x86-64 looks tedious and pkging mesa looks tedious too, but don't want this work to go to waste.

jhheider:
Agree back to front. Mesa will be a pain (looked into it in depth yesterday). Yes, we could dive through deps for platforms, but better to be explicit.

mxcl:
needs brewkit 0.27.1 because we lost unzip somehow in the extract script

PR #1730, opened 2023-04-28
github.com/adah1972/libunibreak

### LINKED ISSUE #537: +libunibreak (146/548)
Formulae: https://formulae.brew.sh/formula/libunibreak

and-ri:
closed #537

### DIFF
--- a/projects/github.com/adah1972/libunibreak/package.yml
+++ b/projects/github.com/adah1972/libunibreak/package.yml
@@ -1,0 +1,25 @@
+distributable:
+  url: https://github.com/adah1972/libunibreak/releases/download/libunibreak_{{version.major}}_{{version.minor}}/libunibreak-{{version.marketing}}.tar.gz
+  strip-components: 1
+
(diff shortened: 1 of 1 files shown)


and-ri:
I can't successfully parse the version from github. Maybe there is some way?
Versions are named like '5_1' or '4_2-1'

jhheider:
> I can't successfully parse the version from github. Maybe there is some way? Versions are named like '5_1' or '4_2-1'

A recent change switched the default from releases to releases/tag. Explicitly using /releases should work, given their release names.

and-ri:
> > I can't successfully parse the version from github. Maybe there is some way? Versions are named like '5_1' or '4_2-1'
> 
> A recent change switched the default from releases to releases/tag. Explicitly using /releases should work, given their release names.

It turned out to be easier than I thought! Thank you 🙏 😊

jhheider:
When it's not easy, that's when we need to do more work. ;)

PR #1740, opened 2023-04-30
+tectonic-typesetting.github.io

branchv:
Adds [tectonic](https://github.com/tectonic-typesetting/tectonic), which needs a harfbuzz built with graphite (another option is allowing tectonic to use a vendored version)


jhheider:
Looks good. Finding the desired feature set for some libraries can be difficult; glad it was as simple as a flag change for harfbuzz.

PR #1828, opened 2023-05-04
+aws-sdk-cpp

### LINKED ISSUE #545: +aws-sdk-cpp (173/548)
Formulae: https://formulae.brew.sh/formula/aws-sdk-cpp

JrGoodle:
closes teaxyz/pantry#545

### DIFF
--- a/projects/github.com/aws/aws-sdk-cpp/package.yml
+++ b/projects/github.com/aws/aws-sdk-cpp/package.yml
@@ -7,6 +7,7 @@ versions:
 
 dependencies:
   curl.se: '*'
+  zlib.net: 1
 
 build:
   dependencies:
@@ -36,6 +37,8 @@ build:
       - -DENABLE_TESTING=OFF
 
 test:
+  dependencies:
+    tea.xyz/gx/cc: c99
   fixture: |
     #include <aws/core/Version.h>
     #include <iostream>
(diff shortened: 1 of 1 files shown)


JrGoodle:
seems like this is gonna be a fun one. there's a submodule that's not included in the tarball. and since there's no .git directory, it's not clear what commit of the submodule is needed. and it's recursive, so the submodule has submodules.

https://github.com/aws/aws-sdk-cpp/blob/main/.gitmodules
https://github.com/aws/aws-sdk-cpp/tree/main/crt
https://github.com/awslabs/aws-crt-cpp/blob/cb474daeeaf5c025bd3408103adf61b97b74e600/.gitmodules

I could imagine just cloning the repo instead of grabbing the tarball. or using the github api to figure out the submodule commit hash, then cloning the correct commit. or getting the commit and then getting a tarball, but you'd have to do that recursively. 

I searched for "submodule" in the pantry repo and didn't see any existing references, so I dunno if this case has come up yet.

JrGoodle:
I think the submodule issue here likely needs to be solved by: https://github.com/teaxyz/brewkit/issues/111

jhheider:
the fixes added work for me with the upcoming brewkit 0.31.0. recommend renaming to `aws.amazon.com/aws-sdk-cpp`.

jhheider:
there we go

mxcl:
Still marked as draft. Deliberate?

jhheider:
gonna go ahead and merge this. we can open for new work if needed.

JrGoodle:
sorry, got a bit sidetracked by work and life. thanks for the followup. I'll try to get to my other straggling PR drafts this week.

btw @jhheider packaging up the git repo as a tarball so it would "just work" with the rest of the brewkit code was pretty clever. when I was looking at the code that's where I got kinda overwhelmed by the changes that would have been needed without that trick.

PR #1831, opened 2023-05-04
httpd fix (bin/apachectl & directories)

and-ri:


and-ri:
I also wanted to make the server immediately ready to start after installation (change a couple of lines in httpd.conf)
```
port=8080

while lsof -i:$port >/dev/null 2>&1; do
      ((port++))
done
    
sed -i '' "s/Listen 8080/Listen $port/" {{prefix}}/etc/httpd/httpd.conf
sed -i.bak 's/#ServerName www.example.com:8080/ServerName localhost:'"$port"'/g' {{prefix}}/etc/httpd/httpd.conf
```
Everything is fine on darwin, but on linux httpd.conf seems to be located in a different path

jhheider:
I think that (or something very much like it) is needed to fix hard-coded path issues.

and-ri:
Oh my god, this is some black magic 🤯. On my mac and ubuntu vps, httpd is configured with the correct paths and works right after installation.

jhheider:
It's because our test installs it under a completely different prefix, specifically to detect this kind of hard-coded issue. So, it's built under `/opt`, then tested under `$HOME`. Has to be, because everyone's $TEA_PREFIX is likely to be completely different (~/.tea will include everyone's usernames).

jhheider:
This is an improvement; although httpd is _supposed_ to accept $APACHE_CONFDIR directly... Still, progress is progress.

PR #1836, opened 2023-05-04
+paulfitz.github.io/daff

magnusviri:
The source code is Haxe. I looked at packaging that up but it's a doozy. Because it's Haxe, the developer has published other versions, in this case, the python version is on Pypi. That's why I'm pointing to that source. I'm also hardcoding the version because the developer hasn't updated the Pypi version to the latest listed on GitHub.


mxcl:
If you mark them as ready when done then people on the team other than me will see it is ready and merge.

projects/paulfitz.github.io/daff/package.yml
```diff
@@ -0,0 +1,21 @@
+distributable:
+  url: https://files.pythonhosted.org/packages/0e/fc/82796c10545f3df9882566c79debac28b664e3a3a08fdb493ac3cc418709/daff-{{version}}.tar.gz
+  strip-components: 1
+
+versions:
+  - 1.3.46
+  #github: paulfitz/daff/tags
```
mxcl: add this, but set an `ignore` for versions you don't want.

projects/paulfitz.github.io/daff/package.yml
```diff
@@ -0,0 +1,21 @@
...
+
+provides:
+  - bin/daff
+
+dependencies:
+  python.org: '>=3.7'
```
mxcl: More likely ^3.7 since this will use v4 if and when it is released

mxcl (changes requested):

magnusviri:
Sorry to take so long to get to this. I've been pretty distracted lately. GHA looks slow. If it passes it's ready. I'll mark it as ready.

jhheider:
No apologies needed, @magnusviri . We respect your time spent. :)

PR #1838, opened 2023-05-05
+minizip

### LINKED ISSUE #549: +minizip (179/548)
Formulae: https://formulae.brew.sh/formula/minizip

JrGoodle:
closes #549

### DIFF
--- a/projects/zlib.net/minizip/package.yml
+++ b/projects/zlib.net/minizip/package.yml
@@ -6,6 +6,9 @@ versions:
   github: madler/zlib
   strip: /^v/
 
+# dependencies:
+#   zlib.net: '*'
+
 build:
   dependencies:
     gnu.org/autoconf: ^2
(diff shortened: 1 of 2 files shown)


JrGoodle:
I dunno if renaming zlib from `zlib.net` -> `zlib.net/zlib` is kosher, but minizip seems like it should be under that prefix too

jhheider:
Shouldn't be necessary, nor desirable. We have things like `rust-lang.org` and `rust-lang.org/cargo`, so it won't cause a conflict.

projects/zlib.net/minizip/package.yml
```diff
@@ -0,0 +1,40 @@
...
+        -e "s:-L\$(zlib_top_builddir):\$(zlib_top_builddir)/libz.a:" \
+        -e "s_-version-info 1:0:0 -l_:-version-info 1:0:0_" \
+        -e "s:libminizip.la -lz:libminizip.la:" \
+        Makefile.am
+      rm Makefile.am.bak
+    fi
+
+    autoreconf -fi
+    ./configure --prefix={{prefix}}
+    make install
```
jhheider: Looks like we run configure and make install twice. Not intended, right?

JrGoodle: I mostly just followed the homebrew formula: https://github.com/Homebrew/homebrew-core/blob/87b20470232f2fb238a297fbcfce08905e395ee7/Formula/minizip.rb

looks like the first one builds zlib, so I guess I should remove that part and use zlib as a dependency instead?

jhheider: ah, I see. yes, I would imagine we don't need a local build of zlib. I mean, I hope.

jhheider:
Looks great!

PR #1868, opened 2023-05-07
+exiftool

### LINKED ISSUE #732: +exiftool (520/548)
Formulae: https://formulae.brew.sh/formula/exiftool

JrGoodle:
closes #732


JrGoodle:
@jhheider seeing this error on CI:

```
Can't locate Image/ExifTool.pm in @INC (you may need to install the Image::ExifTool module) (@INC contains: /opt/exiftool.org/v12.60.0/libexec/lib /Library/Perl/5.30/darwin-thread-multi-2level /Library/Perl/5.30 /Network/Library/Perl/5.30/darwin-thread-multi-2level /Network/Library/Perl/5.30 /Library/Perl/Updates/5.30.3 /System/Library/Perl/5.30/darwin-thread-multi-2level /System/Library/Perl/5.30 /System/Library/Perl/Extras/5.30/darwin-thread-multi-2level /System/Library/Perl/Extras/5.30) at /Users/builder/.tea/exiftool.org/v12.60.0/bin/exiftool line 40.
[57](https://github.com/teaxyz/pantry/actions/runs/4908255524/jobs/8763909425?pr=1868#step:9:58)
BEGIN failed--compilation aborted at /Users/builder/.tea/exiftool.org/v12.60.0/bin/exiftool line 40.
```

The test works locally, so I'm guessing there's an issue with what I did here:

```
sed -i.bak -e 's|unshift @INC, $incDir;|unshift @INC, \"{{prefix}}/libexec/lib\";|' exiftool
```

Seems maybe similar to the other substituations with `_IMPORT_PREFIX`? Could you possibly explain exactly what's happening with the `_IMPORT_PREFIX` sed substitutions? I assume it must have to do with relocatability, but I don't quite grok exactly what's happening with that variable and all the relative path components.

jhheider:
Sure. Lots of packages embedded the build output directory in the package; oft times like "prefix=/opt/foo.bar/baz/v1.2.3/bin/baz". The test bed uses a different $TEA_PREFIX ($HOME/.tea, on various OS images), to ensure that installing to different prefixes work. Strongly necessary, since most installs will be at `/(home|Users)/$username/.tea/`. So, we often have to do some work to make it use its relative path. The common one for shell scripts is one like this:
```sh
sed -i.bak "s_{{prefix}}_\$(cd \$(dirname \$0)/.. \&\& pwd)_g" {{prefix}}/bin/*
```
to allow it locate its root based on its execution path. Different languages require different hacks. Other `perl` packages use `cpanmin.us`, such as https://github.com/teaxyz/pantry/blob/main/projects/github.com/linux-test-project/lcov/package.yml. That's probably the fix style you need.

jhheider:
woo! looks good!

PR #1872, opened 2023-05-07
qhull.org

### LINKED ISSUE #576: +qhull (252/548)
Formulae: https://formulae.brew.sh/formula/qhull

and-ri:
closed #576


and-ri:
Hm strange. Is this a bug or am I missing something?

jhheider:
Looks like it could be a bug in `fix-elf.ts`; this line in their build is suspcious:
```sh
-- Set runtime path of "/opt/qhull.org/v8.0.2/lib/libqhull_r.so.8.0.2" to "lib"
```
you might be able to do a `-DLIB_INSTALL_DIR="{{prefix}}"/lib`, based on https://github.com/qhull/qhull/blob/c2ef2209c28dc61ccfd22514971236587e820121/CMakeLists.txt#L124-L127

PR #1915, opened 2023-05-08
display-names

mxcl:
gui requirements /cc @ABevier @getneil @owenjames 

1. use this if set;
2. use the first bin item in the provides list;
3. otherwise, use the FQD

### DIFF
--- a/projects/agpt.co/package.yml
+++ b/projects/agpt.co/package.yml
@@ -2,6 +2,8 @@ distributable:
   url: https://github.com/Significant-Gravitas/Auto-GPT/archive/refs/tags/v{{version}}.tar.gz
   strip-components: 1
 
+display-name: Auto-GPT
+
 versions:
   github: Significant-Gravitas/Auto-GPT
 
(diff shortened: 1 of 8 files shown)


getneil:
linked this to kettle:168

mxcl:
based on our call today I think we won't be using this. sorry.

I should have said: don't implement anything until merged. Leaving open in case I change my mind. :D

getneil:
@mxcl noted.

mxcl:
@getneil k we're going to go with this now, please see the list of instructions at the top for how this should work.

we'll probs want to redo all the names

jhheider:
> gui requirements /cc @ABevier @getneil @owenjames
> 
> 1. use this if set;
> 2. use the first bin item in the provides list;
> 3. otherwise, use the FQD

Potentially, it might be worth considering:

> 3. fqd.match('/') && fqd.split('/').slice(-1)[0]
> 4. fqd.split('.').slice(-2)[0]

that should be right _enough_ for almost anything.

mxcl:
@ABevier I'll merge…? not sure how things are for this field

ABevier:
> @ABevier I'll merge…? not sure how things are for this field

go for it.  The GUI should be respecting this field in new builds.

PR #1927, opened 2023-05-09
sd-webui now has real releases

mxcl:


mxcl:
Note that this version will not show up as an update until we resolve how to rank v1 above v2023. We're working on that.

PR #1966, opened 2023-05-13
libsdl dependency

sanchitram1:
Added libsdl as a dependency for stream and command to work.


jhheider:
Need to rebuild after brewkit 0.31.0 finishes (for https://github.com/teaxyz/brewkit/commit/dfe754739f485cf094f340485ca55d8410eab456); I'll remove the redundant dep to trigger that.

Sadly, can't add a test for `stream` since it pulls from the mic.

PR #2000, opened 2023-05-19
+zsh-syntax-highlighting

sanchitram1:
1. Need to replace `/Users/sanch` with `$HOME`, but `$HOME` refers to the build directory.  What's the shorthand to refer to `$HOME`, or specifically, wherever the users places their `.zshrc` file?
2. In the test, running that source line to refresh the open terminal window results in this error below.  If I open a new terminal, it seems to have installed correctly.  
```bash
+ source /Users/sanch/.tea/github.com/zsh-users/zsh-syntax-highlighting/v0.7.1/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
+++ builtin alias -Lm '[^+]*'
/Users/sanch/.tea/github.com/zsh-users/zsh-syntax-highlighting/v0.7.1/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh: line 31: alias: -L: invalid option
alias: usage: alias [-p] [name[=value] ... ]
++ typeset zsh_highlight__aliases=
++ builtin unalias -m '[^+]*'
/Users/sanch/.tea/github.com/zsh-users/zsh-syntax-highlighting/v0.7.1/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh: line 36: unalias: -m: invalid option
unalias: usage: unalias [-a] name [name ...]
```

### DIFF
--- a/projects/github.com/zsh-users/zsh-syntax-highlighting/fixture.sh
+++ b/projects/github.com/zsh-users/zsh-syntax-highlighting/fixture.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+# Receive the directory as the first command-line argument
+directory="$1"
+
+# Check if the directory is provided
+if [ -z "$directory" ]; then
+  echo "Directory parameter is missing."
+  exit 1
+fi
+
+# Change to the specified directory
+cd "$directory"
+
+output=$(zsh tests/test-highlighting.zsh main)
+lines=()
+
+# Read the lines of the test
+while IFS= read -r line; do
+  lines+=("$line")
+done <<< "$output"
+
+# If line has not ok, with no TODO --> fail
+for line in "${lines[@]}"; do
+  if [[ $line =~ ^(.*\bnot ok\b)(.*)$ && ! $line =~ "#TODO" ]]; then
+    echo "Fail"
+    # Exit the loop if a failure is found (optional)
+    exit 1
+  fi
+done
+
+echo "Pass"
--- a/projects/github.com/zsh-users/zsh-syntax-highlighting/fixture.sh
+++ b/projects/github.com/zsh-users/zsh-syntax-highlighting/fixture.sh
@@ -10,21 +10,19 @@ if [ -z "$directory" ]; then
 fi
 
 # Change to the specified directory
-cd "$directory"
+cd "$directory" || { echo "Failed to change to the specified directory."; exit 1; }
 
+# Run the test & read the lines
 output=$(zsh tests/test-highlighting.zsh main)
 lines=()
-
-# Read the lines of the test
 while IFS= read -r line; do
   lines+=("$line")
 done <<< "$output"
 
-# If line has not ok, with no TODO --> fail
+# If line has 'not ok', with no '#TODO' --> fail & exit
 for line in "${lines[@]}"; do
   if [[ $line =~ ^(.*\bnot ok\b)(.*)$ && ! $line =~ "#TODO" ]]; then
     echo "Fail"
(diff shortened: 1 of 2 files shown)


sanchitram1:
How would you test something like this?  I was thinking of creating an executable in bin that repeats the instructions I print out after it builds, and then testing whether those instructions print correctly.

jhheider:
Hard to test interactive stuff. You could call the functions it provides (after `source`ing it) and see what they return, maybe?

sanchitram1:
They have some tests you can run, looking into those.  Tearing my hair out trying to figure it out right now thought :-)

sanchitram1:
Ok, so the package provides some example test-data for each highlighter, and a bunch of testing scripts to validate whether the highlighter behaves as expected.  

When you build the package, it doesn't include the testing data, so I pick one file at random, and put it in the main highlighter.  I also copy over the testing scripts.

The test just validates that the main highlighter correctly highlights the test data.  If it doesn't the test script outputs "not ok".  If it does, without a known "#TODO" -- that is a fail.

sanchitram1:
Missing executables error?

jhheider:
```yaml
provides:
  # Made an executable to give the user installation instructions, in case they forgot :-)
  - bin/zsh-syntax-highlighting
```
i don't see anything in the build script that makes the above executable.

sanchitram1:
No executable is provided by this package, so removed it from yaml.  Latest commit just cleans up the comments, and references pkgxdev/libpkgx#69 in the yaml file.

sanchitram1:
So the alternate approach we have:

Create .zshrc in {{prefix}}/etc that does:
```sh
for x in zshenv zshprofile zshrc zlogin; do
  test -f ~/$x && source ~/$x
done

source ./zsh-completions.zsh
export ZDOTDIR=~/.zsh
```

Won't that last line undo everything, since it'll use the actual value of $HOME and not the tea value?

jhheider:
> Won't that last line undo everything, since it'll use the actual value of $HOME and not the tea value?

Yup, that's the point. To source in the files provided by the package, then set everything back to "normal".

jhheider:
> ```shell
> source ./zsh-completions.zsh
> export ZDOTDIR=~/.zsh
> ```

Almost certainly needs to be:
```shell
source {{prefix}}/etc/zsh-completions.zsh
export ZDOTDIR=~
```

The only real problem will be if their normal ZDOTDIR isn't ~, but most people will install the file, then source it explicitly.

sanchitram1:
Right, but if we set everything back to "normal", then a new shell that you open up won't have syntax highlighting.  The line `source {{prefix}}/etc/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh` needs to be in the `.zshrc`, for the shell to know that it has to use it.

I thought the approach would be to set `ZDOTDIR` to the new `.zshrc` in {{prefix}}, which basically contains all your original z dot files, and also sources the `zsh-syntax-highlighting.zsh`.  The new `.zshrc` would not reset the value of ZDOTDIR in this case.

jhheider:
Hm. That might be right. It's all about expectation. I don't feel amazing about taking over ZDOTDIR, hence my first impulse to restore it.

sanchitram1:
I agree.  I think the way this project is currently packaged is consistent with the way you build it now -- you'd have to manually run the command to add the `source ...` to your `.zshrc`.

I was going to work on zsh-completions next, but it's an identical problem, since we'd have to add ` fpath=(path/to/zsh-completions/src $fpath)` to the `.zshrc`

PR #2126, opened 2023-06-07
+python2

jhheider:
@sanchitram1 how's your python2? That first test block could use some tests.

### DIFF
--- a/projects/python.org/package.yml
+++ b/projects/python.org/package.yml
@@ -201,30 +201,26 @@ test:
     tea.xyz/gx/cc: c99
     crates.io/semverator: '*'
   script:
-    # FIXME: python2 tests!
-    - |
-      if semverator satisfies '^2' {{ version }}; then
-        # some tests here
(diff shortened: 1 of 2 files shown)


mxcl:
Approved!

mxcl:
since we will eventually control the entire environment (including kernel thanks to containers) we will be able to gradually just build older and older things

security issues begin to occur, but we can flag things insecure and then prohibit you from automatically using them in production envs.

jhheider:
Got it passing at least a few tests. Let's make this happen.

mxcl:
lovely. Unfortunately it doesn’t just work due to the companions issue. I should see about fixing that.

sanchitram1:
Apologies for missing this. Do you still need any help with tests?

jhheider:
I never turn down better tests, but it has non-trivial tests which satisfies my minimum requirements.

Until it breaks, naturally.

PR #2132, opened 2023-06-08
fix baked in zsh search paths

jhheider:
fix for https://github.com/teaxyz/cli/actions/runs/5205491199/jobs/9391004477?pr=608 hopefully

trying to work around this:

```
module_path <S> <Z> (MODULE_PATH <S>)
An array (colon-separated list) of directories that zmodload searches for dynamically loadable modules. This
is initialized to a standard pathname, usually ‘/usr/local/lib/zsh/$ZSH_VERSION’. (The
‘/usr/local/lib’ part varies from installation to installation.) For security reasons, any value set in
the environment when the shell is started will be ignored.

These parameters only exist if the installation supports dynamic module loading.
```

### DIFF
--- a/projects/zsh.sourceforge.io/package.yml
+++ b/projects/zsh.sourceforge.io/package.yml
@@ -17,19 +17,25 @@ dependencies:
   invisible-island.net/ncurses: '*'
   pcre.org: '*'
 
+runtime:
+  env:
+    FPATH: ${{prefix}}/functions:$FPATH
+    MODULE_PATH: ${{prefix}}/lib/zsh/{{version.marketing}}
+
 build:
   dependencies:
     tea.xyz/gx/cc: c99
     tea.xyz/gx/make: '*'
-  script: |
-    SHARE="{{tea.prefix}}/zsh.sourceforge.io"
+  script:
+    # We need to be able to set MODULE_PATH to the correct location, so we have
+    # to break a small piece of zsh security
+    - run: |
+        sed -i.bak -e 's/^\(IPDEF8("MODULE_PATH",.*\)PM_DONTIMPORT|\(.*\)$/\1\2/' params.c
+        rm params.c.bak
+      working-directory: Src
 
-    ./configure $ARGS \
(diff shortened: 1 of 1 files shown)


jhheider:
@mxcl you ok with those `ZDOTDIR` shenanigans for fixing `zsh`? I daresay our existing package was somewhat broken without that.

mxcl:
this could break user’s configuration so we need another solution.

is the test sufficient to validate the functionality? If so I can have a crack at a source level fix.

jhheider:
The test should be sufficient. The breakage is because it bakes in the search paths at build time.

jhheider:
C++ fix turned out to be not terribly difficult. Just allow passing `MODULE_PATH` on the environment, which we handle using our `runtime` key. I don't *like* altering people's security models, but this seems like smart way to do it correctly.

projects/zsh.sourceforge.io/package.yml
```diff
@@ -17,19 +17,25 @@ dependencies:
   invisible-island.net/ncurses: '*'
   pcre.org: '*'
 
+runtime:
+  env:
+    FPATH: ${{prefix}}/functions:$FPATH
```
mxcl, line 22: We automatically append `$FPATH` (IIRC, it’s been a while since I was there).

jhheider: It didn't appear to get set unless I set it.

mxcl: seems like a bug then

jhheider: `FPATH` doesn't appear in cli or lib, so I think we're ok.

jhheider: Unless you mean we automatically convert any key to a ':'-list, which would be risky.

mxcl: I think we do automatically convert everything to a `:` list and you right: that's probs wrong.

Issue #571, opened 2023-03-13
+mecab-ipadic (241/548)


mfts:
Formulae: https://formulae.brew.sh/formula/mecab-ipadic

PR #2133, opened 2023-06-08
+mecab

### LINKED ISSUE #540: +mecab (160/548)
Formulae: https://formulae.brew.sh/formula/mecab

sanchitram1:
closes #540


projects/mecab/package.yml
```diff
@@ -0,0 +1,42 @@
+distributable:
+  url: https://deb.debian.org/debian/pool/main/m/mecab/mecab_{{version.raw}}.orig.tar.gz
+  # Canonical URL = https://drive.google.com/uc?export=download&id=0B4y35FiV1wh7cENtOXlicTFaRUE
+  # homepage = https://taku910.github.io/mecab/
+  strip-components: 1
+
+versions:
+  - 0.996
+  # Do not think they have versions on the site, homebrew picks it up from deb.debian.org
```
mxcl: @jhheider anywhere you can see we can scrape the versions?

sanchitram1: in the [translated web page](https://taku910.github.io/mecab/), there is a What's New section that has some versions.

sanchitram1:
dependency for #571

projects/mecab/package.yml
jhheider: http://ftp.debian.org/debian/pool/main/m/mecab/

jhheider: Package should likely be named taku910.github.io/mecab

mxcl: good catch, yeah this name is no good. We need an audit that detects such things imo.

mxcl:
Needs a couple changes thanks @sanchitram1

projects/mecab/package.yml
jhheider: shouldn't be hard. I envision a tri-partite test
- package name is parsable as a valid URL
- URL.host contains a period
- URL.host is resolvable via DNS

sanchitram1:
> http://ftp.debian.org/debian/pool/main/m/mecab/

@jhheider Apologizing in advance, I am terrible with regex & url scraping.  On your linked webpage, the tarball I want to fetch is [mecab_0.996.orig.tar.gz](http://ftp.debian.org/debian/pool/main/m/mecab/mecab_0.996.orig.tar.gz).  So, the regex to match in this case would be: 

```yaml
url: http://ftp.debian.org/debian/pool/main/m/mecab/
match: /mecab_\d+\.\d+\.orig\.tar\.gz
```

I am getting an error.  My theory is that tea expects major, minor, and patch, but the package only provides major & minor?

```sh
error: Uncaught Error: not-found: version: taku910.github.io/mecab
  if (!version) throw new Error(`not-found: version: ${utils.pkg.str(spec)}`)
                      ^
    at Object.resolve (file:///Users/sanch/.tea/tea.xyz/brewkit/v0.34.1/lib/usePantry.ts:41:23)
    at eventLoopTick (ext:core/01_core.js:182:11)
    at async file:///Users/sanch/.tea/tea.xyz/brewkit/v0.34.1/libexec/resolve.ts:15:15
````

**Comment by jhheider** [comment]
No, that part is fine. You need an array of `strip` regexes to remove everything that isn't version info. Probably

```yaml
strip:
  - /mecab_/
  - /\.orig\.tar\.gz/
```

sanchitram1:
I'd add the `relocatibility` commit to the examples @mfts.  I feel like I have seen it before somewhere.

jhheider:
Yeah, it's a common fixup for shell scripts.

PR #2134, opened 2023-06-08
+unxip

davdroman:


mxcl:
Could be our swift/xcode is too old to build this :/ @jhheider what we got again?

mxcl:
This is a case where we could just grab the homebrew bottles TBH. There's nothing non-relocatable about this.

davdroman:
> Could be our swift/xcode is too old to build this :/

Yeah I think Swift 5.7 is required, otherwise I just went for the bin download from the repo itself.

davdroman:
Interestingly, it doesn't even run because Swift Concurrency isn't available. So we might need to bump macOS/Swift anyway if we want this tool to pass validation at all.

mxcl:
Tests pass on darwin/aarch64 but not darwin/intel. Presumably because our test-runner is on an older macOS.

@jhheider we'll need solutions for this kind of thing eventually.

jhheider:
Yeah, that's going to be a headache. Not just platforms but versions.

mxcl:
for now let's just go with: keep macOS current and hope for the best. At the very least pls keep them on the same version of macOS.

jhheider:
excluded darwin+x86-64 for testing (since we use `macos-11`). I'll add an exception to `get-platform` for it eventually.

PR #2135, opened 2023-06-08
use `distributable.url` over `distributable.git`

mxcl:


davdroman:
@mxcl perhaps you might be interested in bottling and using this official tool for installing Swift on Linux https://github.com/swift-server/swiftly

Issue #725, opened 2023-03-13
+vault (508/548)


mfts:
Formulae: https://formulae.brew.sh/formula/vault

PR #2138, opened 2023-06-08
+vault

sanchitram1:
gox is a dependency according to homebrew, gonna look at packaging that up as well

however, this built without gox.

adds #725


sanchitram1:
Isn't this deprecated because of #2135 ?

Also, those repositories definitely exist.  Each of the actions were failing at different packages previously, and they all have a repo on GitHub.  I think the Makefile has a bootstrap section which calls this:
```make
# bootstrap the build by downloading additional tools needed to build
ci-bootstrap:
	@for tool in  $(EXTERNAL_TOOLS_CI) ; do \
		echo "Installing/Updating $$tool" ; \
		GO111MODULE=off $(GO_CMD) get -u $$tool; \
	done

# bootstrap the build by downloading additional tools that may be used by devs
bootstrap: ci-bootstrap
	go generate -tags tools tools/tools.go
```

jhheider:
i used the old syntax. hopefully the right syntax will work.

jhheider:
bam.

sanchitram1:
Another potential example for the wiki @mfts.  If you're getting the error `fatal: Not a git repository`, might have to try the `distributable.url git+...` method

PR #2155, opened 2023-06-12
lunarvim.org

### LINKED ISSUE #94: +lunarvim.org
https://www.lunarvim.org/
tea +python.org +rust-lang.org +neovim.io
LV_BRANCH='release-1.2/neovim-0.8' bash <(curl -s https://raw.githubusercontent.com/lunarvim/lunarvim/master/utils/installer/install.sh)

Installation of Rust dependencies: fd::fd-find, rg::ripgrep doesn't seem to work due to some unknown dependencies but the rest should work.

and-ri:
I'm not sure about test section... Does't found command to get lunarvim version
closes #94

### DIFF
--- a/projects/lunarvim.org/package.yml
+++ b/projects/lunarvim.org/package.yml
@@ -12,6 +12,7 @@ dependencies:
   pip.pypa.io: '*'
   nodejs.org: '*'
   rust-lang.org/cargo: '*'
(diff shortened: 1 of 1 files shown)


jhheider:
yes, that's odd for testing. but at least it doesn't segfault.

PR #2163, opened 2023-06-12
publish to new packages topic

ABevier:
We're going to start SNS->SQS for handling new package notifications instead of SQS directly.


.github/scripts/index-packages.ts
```diff
@@ -3,9 +3,13 @@
 import { usePantry } from "hooks"
 import * as ARGV from "./utils/args.ts"
 import { SQSClient, SendMessageCommand } from "npm:@aws-sdk/client-sqs@^3"
+import { SNSClient, PublishCommand } from "npm:@aws-sdk/client-sns"
```
jhheider: Probably smart to at least lock to `@^3` so a new major doesn't just snap this off.

jhheider (approved):
Works for me. Succinct.

PR #2168, opened 2023-06-13
+diff-so-fancy

davdroman:
Based off https://github.com/Homebrew/homebrew-core/blob/480317f32d9a2908002451092ac49b0d001541cd/Formula/diff-so-fancy.rb, but I need some help to make it work 🤔

### DIFF
--- a/projects/github.com/so-fancy/diff-so-fancy/package.yml
+++ b/projects/github.com/so-fancy/diff-so-fancy/package.yml
@@ -7,13 +7,12 @@ versions:
   strip: /^v/
 
 build:
-  script: |
-    chmod +x diff-so-fancy
-    mkdir -p "{{prefix}}/bin"
-    ln -sf "$PWD/diff-so-fancy" "{{prefix}}/bin"
+  script:
+    - chmod +x diff-so-fancy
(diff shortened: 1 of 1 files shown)


projects/github.com/so-fancy/diff-so-fancy/package.yml
```diff
@@ -0,0 +1,19 @@
...
+
+versions:
+  github: so-fancy/diff-so-fancy
+  strip: /^v/
+
+build:
+  script: |
+    chmod +x diff-so-fancy
+    mkdir -p "{{prefix}}/bin"
+    ln -sf "$PWD/diff-so-fancy" "{{prefix}}/bin"
```
davdroman: Something's definitely wrong with this but I'm not sure what it is.

jhheider: That's hardcoding the build $PWD in the symlink. You want `mv diff-so-fancy "{{prefix}}/bin/"`.

jhheider:
there you go! :)

davdroman:
@jhheider ah I see! I’d have never figured that out myself. Thanks for the help! :)

PR #2153, opened 2023-06-11
github.com/strukturag/libde265

and-ri:
closes #317

PR #2179, opened 2023-06-14
freetds.org

### LINKED ISSUE #224: +freetds (190/548)
Formulae: https://formulae.brew.sh/formula/freetds

and-ri:
closes #224


and-ri:
I added the `-i` flag to the `make` command to avoid the error `Global symbol "$prefix" requires explicit package name.`
same problem with #2153

PR #2180, opened 2023-06-14
videolan.org/x264

### LINKED ISSUE #269: +x264 (216/548)
Formulae: https://formulae.brew.sh/formula/x264

and-ri:
closes #269

### DIFF
--- a/projects/videolan.org/x264/package.yml
+++ b/projects/videolan.org/x264/package.yml
@@ -3,6 +3,7 @@ distributable:
   strip-components: 1
 
 versions:
+  #FIXME: every variant of x264 versions we can find has git commit shas in the version name
   - 0.164.3095 #+gitbaee400
 
 build:
--- a/projects/videolan.org/x264/package.yml
+++ b/projects/videolan.org/x264/package.yml
@@ -1,0 +1,38 @@
+distributable:
+  url: https://code.videolan.org/videolan/x264/-/archive/master/x264-master.tar.gz
+  strip-components: 1
+
+versions:
+  - 0.164
+
+dependencies:
+  gnu.org/gcc: '*'
+
+build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+    nasm.us: '*'
+  script: |
+    ./configure $ARGS
+    make --jobs {{ hw.concurrency }} install
+  env:
+    ARGS:
+      - --prefix="{{prefix}}"
+      - --disable-lsmash
+      - --disable-swscale
+      - --disable-ffms
+      - --enable-shared
+      - --enable-static
+      - --enable-strip
+
+provides:
+  - bin/x264
+
+test:
+  dependencies:
+      tea.xyz/gx/cc: c99
+  script: |
+    cc test.c -lx264 -o test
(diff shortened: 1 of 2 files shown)


projects/videolan.org/x264/package.yml
```diff
@@ -0,0 +1,36 @@
+distributable:
+  url: https://code.videolan.org/videolan/x264/-/archive/master/x264-master.tar.gz
```
mxcl: needs to be versioned.

projects/videolan.org/x264/package.yml
```diff
@@ -0,0 +1,36 @@
...
+  strip-components: 1
+
+versions:
+  - 0.164
```
mxcl: ideally there's a page we can scrape /cc @jhheider

mxcl (changes requested):

projects/videolan.org/x264/package.yml
jhheider: https://artifacts.videolan.org/x264/release-macos/ maybe. won't help us generate a URL though...

jhheider: Probably piggybacking on debian is the way to go: http://deb.debian.org/debian/pool/main/x/x264/

and-ri: Thanks @jhheider 
In the last commit, I added a versioned link. But version parsing fails due to extra text (e.g. +gitbaee400)

mxcl:
k so vlc don’t tag releases. Number of questions:

1. debian are doing periodic releases off master, where do they get the version? (`0.155.2917`)
2. what do brew do?
3. we typically do calver for this sort of thing, should we do that here?

projects/videolan.org/x264/package.yml
jhheider: Good point. Who writes software with no releases?

jhheider:
> 2\. what do brew do?

this:
```ruby
  # Cross-check the abbreviated commit hashes from the release filenames with
  # the latest commits in the `stable` Git branch:
  # https://code.videolan.org/videolan/x264/-/commits/stable
  livecheck do
    url "https://artifacts.videolan.org/x264/release-macos-arm64/"
    regex(%r{href=.*?x264[._-](r\d+)[._-]([\da-z]+)/?["' >]}i)
    strategy :page_match do |page, regex|
      # Match the version and abbreviated commit hash in filenames
      matches = page.scan(regex)

      # Fetch the `stable` Git branch Atom feed
      stable_page_data = Homebrew::Livecheck::Strategy.page_content("https://code.videolan.org/videolan/x264/-/commits/stable?format=atom")
      next if stable_page_data[:content].blank?

      # Extract commit hashes from the feed content
      commit_hashes = stable_page_data[:content].scan(%r{/commit/([\da-z]+)}i).flatten
      next if commit_hashes.blank?

      # Only keep versions with a matching commit hash in the `stable` branch
      matches.map do |match|
        release_hash = match[1]
        commit_in_stable = commit_hashes.any? do |commit_hash|
          commit_hash.start_with?(release_hash)
        end

        match[0] if commit_in_stable
      end
    end
  end
```
that seems pretty awful.

PR #2193, opened 2023-06-16
pin deno to 1.34.2 because 1.34.3 broke

ABevier:


jhheider (approved):
I defer to you fully. ;)

PR #2264, opened 2023-06-21
fix(npx)

### LINKED ISSUE #2262: ❌ build issues: npmjs.com=9.7.2
Running log of build failures for npmjs.com=9.7.2

### LINKED ISSUE #2263: ❌ build issues: npmjs.com=9.7.1
Running log of build failures for npmjs.com=9.7.1

jhheider:
nodejs.org>=20.3 has started using a new filesystem call that Docker Desktop 4.20.1 (the newest at time of writing) doesn't support.

We're not alone.

https://github.com/nodejs/node/issues/48444

### DIFF
--- a/projects/npmjs.com/package.yml
+++ b/projects/npmjs.com/package.yml
@@ -10,7 +10,10 @@ provides:
   - bin/npx
 
 dependencies:
-  nodejs.org: '>=14'
+  # 20.3.x causes a error on linux+aarch64 (technically docker)
+  # sh: 1: tldr: Text file busy
+  # https://github.com/nodejs/node/issues/48444
+  nodejs.org: '>=14<20.3'
 
 build:
   dependencies:


projects/npmjs.com/npx-shim
```diff
@@ -1,10 +1,12 @@
...
+set -x
+
 # or npx fails ※
 # FIXME a patch to npm here then upstream would be preferable
 mkdir -p ~/.local/lib
 
-d="$(cd "$(dirname $0)" && pwd)"
-"$d"/npm exec "$@"
+d="$(cd "$(dirname $0)"/../tbin && pwd)"
+"$d"/npx "$@"
```
mxcl: npx is just an alias for `npm exec`. It seems like calling that directly is safer. Dunno.

mxcl:
fixing what specifically?

jhheider:
> fixing what specifically?

hopefully this: https://github.com/teaxyz/pantry/actions/runs/5337582095

projects/npmjs.com/npx-shim
jhheider: true, but it looks like it sets some env. just trying to see what might be the cause of https://github.com/teaxyz/pantry/actions/runs/5337582095

jhheider:
`Text file busy` is supposed to mean someone is editing the `.text` segment of the binary while it's running. That makes little sense. I wonder if the newest node is the issue.

projects/npmjs.com/package.yml
```diff
@@ -10,7 +10,10 @@ provides:
   - bin/npx
 
 dependencies:
-  nodejs.org: '>=14'
+  # 20.3.x causes a error on linux+aarch64 (technically docker)
+  # sh: 1: tldr: Text file busy
+  # https://github.com/nodejs/node/issues/48444
+  nodejs.org: '>=14<20.3'
```
mxcl, line 16: we kinda need an override at a different level for this sort of thing.

an env which is “never use this version of node”, because doing this is wrong to the greater ecosystem. Not to mention we may forget to undo it.

jhheider: This is true.

jhheider: Possibly it's smarter just to override for testing, since we _know_ we're in docker there?

PR #2281, opened 2023-06-23
git-lfs.github.com

### LINKED ISSUE #324: +git-lfs (461/548)
Formulae: https://formulae.brew.sh/formula/git-lfs

and-ri:
closes #324


projects/git-lfs.guthub.com/package.yml
```diff
@@ -0,0 +1,39 @@
...
+  
+  script:
+    - make
+    - make man RONN={{deps.rtomayko.github.io/ronn.prefix}}/bin/ronn
+    - mkdir -p {{prefix}}
+    - mv ./bin/ {{prefix}}/
+    - mv ./man/ {{prefix}}/
+
+  env:
+    GIT_LFS_SHA:
```
jhheider: is this intentionally blank? it probably warrants a comment, if so.

projects/git-lfs.github.com/package.yml
```diff
@@ -0,0 +1,46 @@
...
+versions:
+  github: git-lfs/git-lfs
+
+dependencies:
+  git-scm.org: '*'
+  
+build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
```
jhheider: Are either of these being used?

projects/git-lfs.github.com/package.yml
```diff
@@ -0,0 +1,46 @@
...
+    gnu.org/wget: '*'
+  
+  script:
+    - wget $DOWNLOAD_URL
+    - $UNARCHIVE
```
jhheider: Are we unable to use https://github.com/git-lfs/git-lfs#from-source to build it? if so, it should have the 
```yaml
warnings:
  - vendored
```
block at the top level.

and-ri: This is the only way to avoid the "error: git-lfs died on signal 11" error when testing. I couldn't think of a better solution :)

PR #2306, opened 2023-06-27
open-mpi.org

### LINKED ISSUE #349: +open-mpi (245/548)
Formulae: https://formulae.brew.sh/formula/open-mpi

and-ri:
closes #349


projects/open-mpi.org/package.yml
```diff
@@ -0,0 +1,73 @@
...
+    - ./configure $ARGS
+    - make --jobs {{ hw.concurrency }} all
+    - make --jobs {{ hw.concurrency }} check
+    - make --jobs {{ hw.concurrency }} install
+    - install {{prefix}}/lib/*.mod {{prefix}}/include/
+    - sed -i.bak 's|{{prefix}}|${pcfiledir}/../..|g' {{prefix}}/lib/pkgconfig/*.pc
+    
+  env:
+    darwin:
+      MACOSX_DEPLOYMENT_TARGET: "$(sw_vers -productVersion)"
```
jhheider: This should be automatically 10.6, so we support versions we intend (i.e. you shouldn't have to set it). does it not build unless set?

projects/open-mpi.org/package.yml
```diff
@@ -0,0 +1,73 @@
...
+  strip-components: 1
+
+versions:
+  github: open-mpi/ompi/tags
+
+dependencies:
+  open-mpi.org/hwloc: '*'
+  libevent.org: '*'
+  linux:
+    gnu.org/gcc: '*'
```
jhheider: Is this really a run-time and not a build-time dep? And doesn't darwin need a compiler as well?

and-ri: image
the homebrew formula states that gcc is needed for gfortran
hmm i must have excluded it for darwin during one of my attempts to assemble it and forgot to return it 😄

jhheider: fun. weird that it'd be runtime.

and-ri: If I return gcc for darwin I get:
```
configure:6591: checking whether the C compiler works
configure:6613: gcc -w   -Wl,-rpath,/Users/andrejrabcenko/.tea conftest.c  >&5
ld: library not found for -lSystem
collect2: error: ld returned 1 exit status
```

jhheider:
well, that was painful.

PR #2314, opened 2023-06-28
hdfgroup.org/HDF5

### LINKED ISSUE #307: +hdf5 (182/548)
Formulae: https://formulae.brew.sh/formula/hdf5

and-ri:
closes #307

### DIFF
--- a/projects/hdfgroup.org/HDF5/package.yml
+++ b/projects/hdfgroup.org/HDF5/package.yml
@@ -11,6 +11,10 @@ dependencies:
   linux:
     zlib.net: '*'
 
+runtime:
+  env:
+      HDF5_ROOT: '{{prefix}}'
+
 build:
   dependencies:
     tea.xyz/gx/make: '*'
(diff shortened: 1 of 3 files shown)


projects/hdfgroup.org/HDF5/package.yml
```diff
@@ -0,0 +1,72 @@
+distributable:
+  url: https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-{{version.marketing}}/hdf5-{{version}}/src/hdf5-{{version}}-2.tar.bz2
+  strip-components: 2
+
+versions:
+  - 1.14.1
```
and-ri, line 6: I'm not sure about the versions. On the page https://support.hdfgroup.org/ftp/HDF5/releases/, there is no list of all versions, only folders with {{version.marketing}}. And the latest version has an additional `-2` in the file name (`1.14.1-2.tar.bz2`), while the others don't (`1.14.0.tar.bz2`, `1.13.3.tar.bz2`).

There is only one version available on GitHub (`hdf5-1_14_1-2`).

jhheider: Yeah, they don't make it easy. Might be the best we can do for now.

PR #2330, opened 2023-06-30
remove index packages

ABevier:
There is no need to have GHA trigger package indexing.  There is now an S3 trigger whenever a `versions.txt` file is added to `dist.tea.xyz` so we can remove this.


jhheider (approved):
deleting code is the sublimest pleasure

PR #2334, opened 2023-06-30
github.com/anholt/libepoxy

### LINKED ISSUE #378: +libepoxy (346/548)
Formulae: https://formulae.brew.sh/formula/libepoxy

and-ri:
closes #378


jhheider:
looks like this will need freeglut, which will need mesa, which will suck.

and-ri:
> looks like this will need freeglut, which will need mesa, which will suck.

Yes, I'm already working on mesa packaging. This is a real monster

PR #2340, opened 2023-07-01
+nim-lang.org

ivabus:
Init at v1.6.14

### DIFF
--- a/projects/nim-lang.org/package.yml
+++ b/projects/nim-lang.org/package.yml
@@ -2,9 +2,8 @@ distributable:
   url: https://nim-lang.org/download/nim-{{ version }}.tar.xz
   strip-components: 1
 
-# FIXME: test more versions
 versions:
-  - v1.6.14
+  github: nim-lang/Nim/tags
 
 provides:
   - bin/nim
--- a/projects/nim-lang.org/package.yml
+++ b/projects/nim-lang.org/package.yml
@@ -20,7 +20,7 @@ provides:
 
 
 dependencies:
-  llvm.org: '*'
+  gnu.org/gcc: '*'
   pcre.org: '*'
   openssl.org: '*'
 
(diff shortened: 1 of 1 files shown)


ivabus:
Oh, they are waiting for gcc on linux, fix in the way

ivabus:
Ready to merge, review pls

projects/nim-lang.org/package.yml
```diff
@@ -0,0 +1,46 @@
...
+dependencies:
+  gnu.org/gcc: '*'
+  pcre.org: '*'
+  openssl.org: '*'
+
+build:
+  dependencies:
+    gnu.org/gcc: '*'
+    pcre.org: '*'
+    openssl.org: '*'
```
mxcl: if these are the same then you don't need to specify them again in the `build` node, however are these actually also runtime deps?

projects/nim-lang.org/package.yml
```diff
@@ -0,0 +1,46 @@
+distributable:
+  url: https://nim-lang.org/download/nim-{{ version }}.tar.xz
+  strip-components: 1
+
+# FIXME: test more versions
+versions:
+  - v1.6.14
```
mxcl: is there are URL that we can scrape? /cc @jhheider

mxcl (changes requested):

projects/nim-lang.org/package.yml
ivabus: Yes, these are actually runtime deps. Nim compiles to C++, so it needs C++ compiler to run code (LLVM doesn't work properly when compiling Nim, so I turned on gcc, which works fine).

projects/nim-lang.org/package.yml
ivabus: I think we can use GitHub tags. (Commited)

jhheider:
Looking good, @ivabus ; sorry for the delay in addressing.

ivabus:
Yay!

PR #2344, opened 2023-07-02
+gnu.org/findutils

### LINKED ISSUE #695: +findutils (454/548)
Formulae: https://formulae.brew.sh/formula/findutils

and-ri:
closes #695


projects/gnu.org/findutils/package.yml
```diff
@@ -0,0 +1,50 @@
...
+    darwin:
+      ARGS:
+        - --program-prefix=g
+
+provides:
+  darwin:
+    - bin/gfind
+    - bin/glocate
+    - bin/gupdatedb
+    - bin/gxargs
```
jhheider: I'd skip the prefixing on darwin; we have omitted this for other gnu tools, since tea allows controlling what you get.

and-ri: > I'd skip the prefixing on darwin; we have omitted this for other gnu tools, since tea allows controlling what you get.

Oh, okay. I will change :)

PR #2346, opened 2023-07-02
fix(taskfile.dev) Build from local sources

MaxWinterstein:
Change to desired installation - from downloaded sources

Mostly inspired by https://github.com/Homebrew/homebrew-core/blob/master/Formula/go-task.rb


mxcl (approved):

mxcl:
Thank you!

PR #2441, opened 2023-07-07
imagemagick new version & deps

and-ri:


and-ri:
Hmm... does this compile with hardcoded paths?

PR #2549, opened 2023-07-10
+ghostscript

### LINKED ISSUE #248: +ghostscript (158/548)
Formulae: https://formulae.brew.sh/formula/ghostscript

mxcl:
Closes #248

### DIFF
--- a/projects/ghostscript.com/fixture.ps
+++ b/projects/ghostscript.com/fixture.ps
@@ -1,0 +1,6 @@
+%!PS
+/Courier
+20 selectfont
+72 500 moveto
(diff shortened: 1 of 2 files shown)


mxcl:
relocatability issues

mxcl:
I tried to make a setting for brewkit to install at a tmp prefix then mv for relocatability testing but this is complicated since libtea itself has opinions on prefixes. So in the end I just renamed the project locally, built, renamed back again and tested. That worked fine.

Maybe brewkit could do the same actually for the `--relocate` flag.

PR #2576, opened 2023-07-14
+swift-outdated

davdroman:


davdroman:
I believe darwin+x86-64 is failing for a similar reason to https://github.com/teaxyz/pantry/pull/1549#issuecomment-1519081770.

jhheider:
Both builders should have Xcode 14.1 currently.

jhheider:
Hm, I take that back. looks like x86-64 is on xcode 13.1 currently. Let me see about that.

jhheider:
Looks good.

PR #1210, opened 2023-04-02
Sourceware.org/valgrind

magnusviri:
This is Linux only. I basically try to make a formula instead of installing. I keep hoping it will be easy. But...

I have no idea why it gets this error (in docker):

```
gcc -DHAVE_CONFIG_H -I. -I..  -I.. -I../include -I../include -I../VEX/pub -I../VEX/pub -DVGA_arm64=1 -DVGO_linux=1 -DVGP_arm64_linux=1 -DVGPV_arm64_linux_vanilla=1  -I../coregrind -DVG_LIBDIR="\"/root/.tea/sourceware.org/valgrind/v3.20.0/libexec/valgrind"\" -DVG_PLATFORM="\"arm64-linux\""   -m64 -O2 -g -Wall -Wmissing-prototypes -Wshadow -Wpointer-arith -Wstrict-prototypes -Wmissing-declarations -Wcast-align -Wcast-qual -Wwrite-strings -Wempty-body -Wformat -Wformat-security -Wignored-qualifiers -Wenum-conversion -finline-functions -fno-stack-protector -fno-strict-aliasing -fno-builtin -Wno-cast-align -Wno-self-assign -Wno-tautological-compare  -DENABLE_LINUX_TICKET_LOCK -w  -c -o libcoregrind_arm64_linux_a-m_stacktrace.o `test -f 'm_stacktrace.c' || echo './'`m_stacktrace.c
m_machine.c:486:4: error: __builtin_longjmp is not supported for the current target
   VG_MINIMAL_LONGJMP(env_unsup_insn);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../include/pub_tool_libcsetjmp.h:134:35: note: expanded from macro 'VG_MINIMAL_LONGJMP'
#define VG_MINIMAL_LONGJMP(_env)  __builtin_longjmp((_env),1)
                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~
m_machine.c:1819:10: error: __builtin_setjmp is not supported for the current target
     if (VG_MINIMAL_SETJMP(env_unsup_insn))
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../include/pub_tool_libcsetjmp.h:133:44: note: expanded from macro 'VG_MINIMAL_SETJMP'
#define VG_MINIMAL_SETJMP(_env)   ((UWord)(__builtin_setjmp((_env))))
[...]

PR #2578, opened 2023-07-14
valgrind.org

and-ri:


jhheider:
see #1210 for a discussion of the some of the relocation issues in valgrind

PR #2590, opened 2023-07-15
eyrie.org/eagle/podlators

and-ri:


and-ri:
hmm... I don't understand, is it something with pearl? On my ubuntu all it's ok

jhheider:
Have you tried using `cpanmin.us` to install `ExtUtils::MakeMaker`? That'd be my next guess.

and-ri:
> Have you tried using `cpanmin.us` to install `ExtUtils::MakeMaker`? That'd be my next guess.

Oh thanks for the idea. I very rarely come across Pearl and don't know much about it, to be honest. 😄

jhheider:
It was like magic. In the 90s. Nowadays, it feels a little path-complicated like python and ruby do for library resolution.

PR #2601, opened 2023-07-17
+tea-package-builder

jhheider:
@ArionThinker does this look right to you? I know a few of those linux deps aren't done yet (I saw nss and dbus in the list), but this is a start.

### DIFF
--- a/projects/github.com/ArionThinker/tea-package-builder/package.yml
+++ b/projects/github.com/ArionThinker/tea-package-builder/package.yml
@@ -1,0 +1,31 @@
+distributable:
+  url: https://github.com/ArionThinker/tea-package-builder/archive/refs/tags/{{version}}.tar.gz
+  strip-components: 1
+
+versions:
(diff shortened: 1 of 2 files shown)


mxcl:
like ideally I think the pkg could just vend a web server like sd-webui then the pkg would be relatively tiny.

However I doubt there's a trivial way to make both an electron app and a vendable web-server entrypoint.

A long time ago I proposed we could provide electron as a package and thus if we distribute electron apps they would be relatively tiny. Still would be cool to explore that.

jhheider:
I think it links a single binary on linux, so it might not help there. Though, if we did, we could just soft-link the framework into the `.app` on darwin, for deduplication savings (as long as versioning were preserved.

jhheider:
This is 380MB packaged; not our largest package, but we could save at least a couple hundred meg per via that method.

and-ri:
Oh wow 🔥 
I didn't think it could be packaged. I am very glad 😃

PR #2639, opened 2023-07-21
docker.com/cli

### LINKED ISSUE #277: +docker (333/548)
Formulae: https://formulae.brew.sh/formula/docker

and-ri:
closes #277

### DIFF
--- a/projects/docker.com/cli/package.yml
+++ b/projects/docker.com/cli/package.yml
@@ -37,18 +37,18 @@ provides:
   - bin/docker
 test:
   script:
-    - colima start $ARGS
+    - colima stop
+    - colima start $ARGS test
     - docker --version | grep {{version}}
     - docker context list
-    - docker context use colima
+    - docker context use colima-test
     - docker info
     - colima stop
   env:
     ARGS:
-      - --cpu 2
-      - --memory 1
-      - --disk 1
-      - --very-verbose #debug
+      - --memory 2
+      - --disk 10
+      - --verbose
     aarch64:
       ARGS:
         - --arch aarch64
--- a/projects/docker.com/cli/package.yml
+++ b/projects/docker.com/cli/package.yml
@@ -46,6 +46,7 @@ test:
   env:
     ARGS:
       - --runtime docker
+      - --verbose
     aarch64:
       ARGS:
         - --arch aarch64
--- a/projects/docker.com/cli/package.yml
+++ b/projects/docker.com/cli/package.yml
@@ -8,6 +8,8 @@ platforms:
   - darwin
 dependencies:
   github.com/abiosoft/colima: '*'
+runtime:
+  DOCKER_HOST: "unix://${HOME}/.colima/default/docker.sock"
 build:
   dependencies:
     go.dev: '*'
@@ -37,7 +39,7 @@ provides:
   - bin/docker
 test:
   script:
-    - docker --version | grep {{version}}
     - colima start
+    - docker --version | grep {{version}}
     - docker info
     - colima stop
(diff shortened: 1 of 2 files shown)


and-ri:
Hmm, I don’t understand what’s wrong with colima ... I checked it on several macbooks and everything starts on them 🧐

jhheider:
Any chance it's conflicting with the running docker instance on the builders?

and-ri:
> Any chance it's conflicting with the running docker instance on the builders?

Oh, I didn’t even think about the fact that the docker might already be running on the builder 🫣

and-ri:
I'm not sure if we can just leave the version check in that case?
And I do not fully understand what is meant by "stub the binaries" here https://github.com/teaxyz/pantry/issues/277#issuecomment-1643900009

jhheider:
In this instance, I suspect he means write a shim that will call docker desktop binaries, if present, otherwise use tea-provided colima in their place (not yet packaged).

mxcl:
instead of stubbing we should probs just document it. I've started a thing where I add a `README.md` to the gui which we can put things like this.

projects/docker.com/cli/package.yml
```diff
@@ -0,0 +1,40 @@
+distributable:
+  url: https://github.com/docker/cli/archive/v{{version}}.tar.gz
+  strip-components: 1
+versions:
+  github: docker/cli/tags
+platforms:
+  # colima -> only darwin
+  - darwin
```
mxcl: we should build for linux and make the dependency darwin only?

and well, the dep isn't a dep really. It's a `companion` at best.

companions cannot be platform restricted tho.

will do for now to have it as a dep. but yeah we should build for linux if poss.

and-ri:
> instead of stubbing we should probs just document it. I've started a thing where I add a `README.md` to the gui which we can put things like this.

The ability to add a `README.md` would be great
The README file will be placed right next to the package.yml, am I understanding correctly?

mxcl:
> The README file will be placed right next to the package.yml, am I understanding correctly?

yes, if you take a look you can see a few I've added

projects/docker.com/cli/package.yml
```diff
@@ -0,0 +1,38 @@
+distributable:
+  url: https://github.com/docker/cli/archive/v{{version}}.tar.gz
+  strip-components: 1
+versions:
+  github: docker/cli/tags
+dependencies:
+  darwin:
+    github.com/abiosoft/colima: '*'
```
mxcl: this should be removed pls

there is no mandatory dep

projects/docker.com/cli/package.yml
```diff
@@ -0,0 +1,38 @@
...
+          go-md2man -in="$md" -out="{{prefix}}/man/man${section}/$(basename "$md" .md)"
+        done
+  env:
+    GOPATH: $PWD
+    GO111MODULE: auto
+    ARGS:
+      - -trimpath
+      - -o={{prefix}}/bin/docker
+provides:
+  - bin/docker
```
mxcl, line 32: doesn't provide more?

mxcl (changes requested):

projects/docker.com/cli/package.yml
and-ri: 👌🏻

projects/docker.com/cli/package.yml
and-ri: only `docker`
image

projects/docker.com/cli/package.yml
```diff
@@ -0,0 +1,35 @@
...
+test:
+  script:
+    - docker --version | grep {{version}}
```
mxcl, line 35: ideally we'd start colima and test it properly, but this is fine for now

mxcl (approved):

PR #2649, opened 2023-07-23
github.com/adegtyarev/streebog: init at 0.13

ivabus:
Asking for a review, I don't really know how to make the build stage look better (but working)


ivabus:
wtf, i can't build it on linux by hands...

jhheider:
So, adding `strip-components: 1` to the `distributable:` will remove the need for that `cd`.

My guess is the linux build needs this fix: https://patch-diff.githubusercontent.com/raw/adegtyarev/streebog/pull/32.patch. But that repo hasn't had a commit in almost 4 years, and a release in 4.5. Is this still alive?

ivabus:
Ok, I’ll try the patch and strip-components. Repo is not alive, but it contains realisation of hashing algorithm, that may need to be used in some specific scripts. I don't know if tea needs this package.

ivabus:
Found out some weird behaviour in `make` which results in building `gost3411-2012` on darwin and `gost3411` on linux

EDIT: idk what is going on, now it builds with correct names without changing anything

jhheider:
woo! looks good.

PR #2677, opened 2023-07-27
Added symfony-cli

ianwijma:
Let me know if I could improve the package.yml.

Related project: https://github.com/symfony-cli/symfony-cli


jhheider:
Made a few tweaks. All platforms is the same as omitting the platforms list.

PR #2678, opened 2023-07-27
icu.unicode.org

and-ri:
for #2661 linux support

### DIFF
--- a/projects/icu.unicode.org/hello
+++ b/projects/icu.unicode.org/hello
@@ -1,0 +1,1 @@
+hello\nworld\n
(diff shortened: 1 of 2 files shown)


mxcl:
isn't this already added as [unicode.org](https://github.com/teaxyz/pantry/blob/main/projects/unicode.org/package.yml) ?

and-ri:
> isn't this already added as [unicode.org](https://github.com/teaxyz/pantry/blob/main/projects/unicode.org/package.yml) ?

Oh right. The absence of "ICU" in the title misled me 😞

mxcl:
Typically if a domain only has one pkg we just use the domain rather than some uglier alternative.

Originally we were strict on paths and urls and stuff, but over time we relaxed it as the name is not meant to be the homepage as much as a unique identifier.

PR #2681, opened 2023-07-28
+micromamba

mxcl:


mxcl:
linux won’t work until https://github.com/mamba-org/mamba/issues/2650 is fixed

PR #2700, opened 2023-07-30
plantuml.com

### LINKED ISSUE #731: +plantuml (519/548)
Formulae: https://formulae.brew.sh/formula/plantuml

and-ri:
closes #731


and-ri:
hmm... strange that this only happens on darwin aarch64

jhheider:
Looks like it's installing harfbuzz v5, not v8. Not sure _why_ though.

and-ri:
`error: Uncaught Error: cannot intersect: >=4<6 && ^8`
hmm... does that mean it only sees versions 4 to 6?

jhheider:
https://github.com/teaxyz/pantry/blob/577ffe507465e5b26411c39c951177e5e0b2c382/projects/gnome.org/pango/package.yml#L12 almost certainly. If pango passes tests with '>=4', then we should be good to unrestrict that.

and-ri:
wow 😮

Issue #279, opened 2023-02-10
+nss (177/548)


mfts:
Formulae: https://formulae.brew.sh/formula/nss

PR #2722, opened 2023-08-02
chromedriver.chromium.org

### LINKED ISSUE #2711: +chromedriver
https://github.com/macports/macports-ports/blob/master/www/chromedriver/Portfile

Auto-GPT needs it. So pls pkg this and then make it a dependency of Auto-GPT.

and-ri:
closes #2711


and-ri:
require #279 for linux

PR #2730, opened 2023-08-02
jmeter.apache.org

### LINKED ISSUE #698: +jmeter (457/548)
Formulae: https://formulae.brew.sh/formula/jmeter

and-ri:
closes #698


mxcl:
can we prefer eg. apache.org/jmeter for such things.

I know it's not strictly the homepage, but that's not the primary reason for our namespacing scheme.

I'm trying to keep the root namespace a little clearer.

and-ri:
> can we prefer eg. apache.org/jmeter for such things.
> 
> I know it's not strictly the homepage, but that's not the primary reason for our namespacing scheme.
> 
> I'm trying to keep the root namespace a little clearer.

Yes sure

projects/jmeter.apache.org/package.yml
```diff
@@ -11,6 +11,9 @@ warnings:
   - vendored
 dependencies:
   openjdk.org: '*'
+runtime:
+  env:
+    JAVA_HOME: "{{prefix}}"
```
mxcl: This seems wrong, the user can set this, we are overriding it.

and-ri: Hmm 🤔 

perhaps it would be correct to check the existence of a variable and assign a value if it does not exist?

mxcl: why is it set? Add comments pls

and-ri: JAVA_HOME just resolve problem 
```
The operation couldn’t be completed. Unable to locate a Java Runtime.
Please visit http://www.java.com/ for information on installing Java.
````
https://github.com/teaxyz/pantry/actions/runs/5743363132/job/15567693928

and-ri: Or maybe we can add JAVA_HOME in `test` section and add `README.md` with information that JAVA_HOME must be set in env?

jhheider: Shouldn't `JAVA_HOME` be `${{deps.openjdk.org.prefix}}`, and exported from that package? Per my limited recollection of configuring Java. @ABevier was a Javanista: any thoughts on this?

and-ri: Yes, the final version is {{deps.openjdk.org.prefix}}

https://github.com/teaxyz/pantry/commit/70957af471f223d8cb60abf22867b1e6cd0e2a04

jhheider: Right. But shouldn't that be in the runtime of `openjdk.org`, not packages that depend on it?

and-ri: > Right. But shouldn't that be in the runtime of `openjdk.org`, not packages that depend on it?

Yes, I had thoughts of doing this too. I made changes to this pull request
https://github.com/teaxyz/pantry/pull/2727

PR #2746, opened 2023-08-05
+khronos.org/opencl-headers

magnusviri:
I totally copied glslang. I don't know what the cmake flags do. Can someone way more knowledgeable about this help?

### DIFF
--- a/projects/khronos.org/opencl-headers/package.yml
+++ b/projects/khronos.org/opencl-headers/package.yml
@@ -1,0 +1,43 @@
+distributable:
+  url: https://github.com/KhronosGroup/OpenCL-Headers/archive/refs/tags/v{{version.raw}}.tar.gz
+  strip-components: 1
+
+versions:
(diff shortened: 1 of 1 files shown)


jhheider:
It looks like there's only two custom headers in `CmakeLists.txt`:
```cmake
option(OPENCL_HEADERS_BUILD_TESTING "Enable support for OpenCL C headers testing." OFF)
option(OPENCL_HEADERS_BUILD_CXX_TESTS "Enable support for OpenCL C headers testing in C++ mode." ON)
```

Looks like the [homebrew forumula](https://github.com/Homebrew/homebrew-core/blob/6d08f69a9ed2356ada55f3771de29585335dfa3c/Formula/opencl-headers.rb) doesn't have anything special set. So, you shouldn't need much more than a prefix.

magnusviri:
Ok, I think it's ready.

PR #2751, opened 2023-08-06
mozilla.org/nss

### LINKED ISSUE #279: +nss (177/548)

and-ri:
closes #279

### DIFF
--- a/projects/mozilla.org/nspr/package.yml
+++ b/projects/mozilla.org/nspr/package.yml
@@ -11,10 +11,15 @@ build:
   dependencies:
     tea.xyz/gx/cc: c99
     tea.xyz/gx/make: '*'
-  script: |
-    ./nspr/configure $ARGS
-    make --jobs {{ hw.concurrency }}
-    make --jobs {{ hw.concurrency }} install
+  script:
+    - ./nspr/configure $ARGS
+    - make --jobs {{ hw.concurrency }}
+    - make --jobs {{ hw.concurrency }} install
+    - run: |
+        mv nspr/* .
+        rmdir nspr
+        ln -s . nspr
+      working-directory: ${{prefix}}/include
   env:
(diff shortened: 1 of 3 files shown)


and-ri:
I'm having trouble building this for darwin/aarch64.
I managed to make some progress using llvm clang but still run into the error `aes-armv8.c:14:2: error: "Compiler option is invalid"`
My mac architecture is x86-64 and I can't reproduce the error locally.
Perhaps you have some ideas?

jhheider:
Something makes me this `-march=armv8-a -mfpu=crypto-neon-fp-armv8` isn't right.

For `gnutls` we had to do this: https://github.com/teaxyz/pantry/blob/577ffe507465e5b26411c39c951177e5e0b2c382/projects/gnutls.org/package.yml#L28-L34

Possibly force-settings arch/cpu/fpu options is the right move. Interesting that it doesn't build out of the box on aarch64, since Apple doesn't make any x86_64 devices any more.

jhheider:
Homebrew's recipe is against 3.87. Maybe see if that builds? This appears to be a mirror of it.

If that works, the release notes might be instructive for building later releases (versions: can be an array).

jhheider:
homebrew's comments make me think that firefox did some seriously non-standard stuff here. might have to branch out for other formulae for clues.

and-ri:
@jhheider maybe we can package it for linux only for now? This commit: https://github.com/teaxyz/pantry/pull/2751/commits/a211990bf9c1d2379149e83303c2286acbbbfbca

[chromedriver](https://github.com/teaxyz/pantry/pull/2722) needs nss on linux

what do you think?

jhheider:
Releasing for Darwin for now with a comment seems fine.

PR #2760, opened 2023-08-07
cairographics.org/pycairo

### LINKED ISSUE #514: +py3cairo (52/548)
Formulae: https://formulae.brew.sh/formula/py3cairo

and-ri:
closes #514

### DIFF
--- a/projects/cairographics.org/pycairo/package.yml
+++ b/projects/cairographics.org/pycairo/package.yml
@@ -15,6 +15,8 @@ build:
   script:
     - python3 setup.py install $ARGS
     - mv {{prefix}}/lib/python{{deps.python.org.version.marketing}}/site-packages/*/* {{prefix}}/lib/python{{deps.python.org.version.marketing}}/site-packages/
+    - run: ln -s python{{deps.python.org.version.marketing}} python{{deps.python.org.version.major}}
+      working-directory: "{{prefix}}/lib"
   env:
     ARGS:
       - --verbose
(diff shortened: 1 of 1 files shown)


projects/cairographics.org/pycairo/package.yml
```diff
@@ -0,0 +1,26 @@
...
+  github: pygobject/pycairo
+dependencies:
+  cairographics.org: '*'
+runtime:
+  env:
+    PYTHONPATH: "{{prefix}}/lib/python3.11/site-packages:$PYTHONPATH"
+build:
+  dependencies:
+    freedesktop.org/pkg-config: '*'
+    python.org: ^3.11
```
jhheider: If you're going to lock the directory via `PYTHONPATH`, then I think you probably need to lock the version as well, to `~3.11`. Possibly that could instead be worked around with symlinks? `ln -s python{{version.marketing}} python{{version.major}}`?

and-ri: I did with {{deps.python.org.version.marketing}}
So will it be ok?

projects/cairographics.org/pycairo/package.yml
```diff
@@ -0,0 +1,28 @@
+distributable:
+  url: https://github.com/pygobject/pycairo/releases/download/v{{version}}/pycairo-{{version}}.tar.gz
+  strip-components: 1
+versions:
+  github: pygobject/pycairo
+dependencies:
+  cairographics.org: '*'
+runtime:
+  env:
+    PYTHONPATH: "{{prefix}}/lib/python{{deps.python.org.version.marketing}}/site-packages:$PYTHONPATH"
```
jhheider: Actually, this probably won't work. Since it doesn't have a runtime dep on python. So, we'd either have to hardcode it to 3 (which is probably what we need to do), or something else that probably won't work.

and-ri: Yes, you are right. I did not think about the fact that Python is not in runtime

PR #2763, opened 2023-08-08
fix display name for rust

ABevier:


jhheider:
Ha. This might be worth adding to audit.ts in brewkit.

PR #2778, opened 2023-08-09
+buf.build — all things protobuf

### LINKED ISSUE #2779: +buf.build
https://buf.build

kevinmichaelchen:
https://buf.build/

Resolves https://github.com/teaxyz/pantry/issues/2779

### DIFF
--- a/projects/buf.build/package.yml
+++ b/projects/buf.build/package.yml
@@ -10,11 +10,17 @@ provides:
 
 build:
   script:
-    - make godeps build
+    - make build install
+    - mkdir -p "{{ prefix }}"/bin
+    - mv ~/.cache/buf/$PLATFORM/gobin/buf "{{ prefix }}"/bin
   dependencies:
     go.dev: ^1.20
     gnu.org/make: '*'
   env:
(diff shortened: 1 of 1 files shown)


projects/buf.build/package.yml
```diff
@@ -0,0 +1,33 @@
...
+  github: bufbuild/buf
+
+provides:
+  - bin/buf
+
+build:
+  script:
+    - make build install
+    - mkdir -p "{{ prefix }}"/bin
+    - mv ~/.cache/buf/$PLATFORM/gobin/buf "{{ prefix }}"/bin
```
jhheider: We should really try to avoid letting this pollute a user's $HOME. https://buf.build/docs/installation#from-source only talks about building from github sources, not local sources.

I suspect our logic should look something like https://github.com/teaxyz/pantry/blob/main/projects/gitlab.com/gitlab-org/cli/package.yml.

kevinmichaelchen: Thanks for the feedback, @jhheider!

jhheider: No problem!

PR #2856, opened 2023-08-12
+turso — deploy your databases to the edge

kevinmichaelchen:
https://turso.tech/

### DIFF
--- a/projects/turso.tech/package.yml
+++ b/projects/turso.tech/package.yml
@@ -3,7 +3,8 @@ distributable:
   strip-components: 1
 
 versions:
-  github: tursodatabase/turso-cli
+  github: tursodatabase/turso-cli/tags
+  strip: /^v/
 
 provides:
   - bin/turso
@@ -23,7 +24,6 @@ build:
     LDFLAGS:
       - -s
       - -w
-      - -X main.version={{version}}
       - -X main.debugMode=false
     linux:
       # or segmentation fault
--- a/projects/turso.tech/package.yml
+++ b/projects/turso.tech/package.yml
@@ -10,8 +10,9 @@ provides:
 
 build:
   script: |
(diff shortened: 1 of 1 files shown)


projects/turso.tech/package.yml
```diff
@@ -0,0 +1,37 @@
...
+
+versions:
+  github: tursodatabase/homebrew-tap
+
+provides:
+  - bin/turso
+
+build:
+  script: |
+    git clone https://github.com/tursodatabase/turso-cli
```
jhheider: Instead of this, either you should make the `distributable.url: https://github.com/tursodatabase/turso-cli/archive/refs/tags/v{{version}}.tar.gz` (and use the same github for the version), or make the `distributable: { url: git+https://github.com/tursodatabase/turso-cli, ref: v{{version}} }`, if it _strictly requires_ being built from a git checkout (rare, but it exists).

kevinmichaelchen: trying the former approach (see https://github.com/teaxyz/pantry/pull/2856/commits/7228003f2fed515c97857ba573edbbaad805d4a2).

```
no versions parsed. Re-run with DEBUG=1 to see output.
{ versions: [] }
error: Uncaught Error: not-found: version: turso.tech
```

Maybe it's a "Tags vs Releases" thing? The Turso repo [has Tags but not Releases](https://github.com/tursodatabase/turso-cli/tags).

jhheider: Looks like setting the version depends on passing `-tags=prod` to the build command.

jhheider:
If you're happy with the above fixes, looks like this is ready, @kevinmichaelchen .

kevinmichaelchen:
@jhheider Thanks for the fixes! LGTM.

PR #2931, opened 2023-08-16
netlib.org/lapack

and-ri:


projects/netlib.org/lapack/package.yml
```diff
@@ -0,0 +1,33 @@
+distributable:
+  url: https://github.com/Reference-LAPACK/lapack/archive/v{{version}}.tar.gz
+  strip-components: 1
+versions:
+  github: Reference-LAPACK/lapack
+dependencies:
+  gnu.org/gcc: '*' # for gfortran
```
jhheider: This isn't right, I don't think. It doesn't have any runtime deps, since, in theory, any fortran compiler could be used, right?

PR #2974, opened 2023-08-20
+dynein -- DynamoDB CLI

kevinmichaelchen:
DynamoDB CLI written in Rust.

https://github.com/awslabs/dynein

### DIFF
--- a/projects/github.com/awslabs/dynein/package.yml
+++ b/projects/github.com/awslabs/dynein/package.yml
@@ -9,14 +9,13 @@ versions:
   github: awslabs/dynein
   strip: /v/
 
-dependencies:
-  linux:
-    openssl.org: '*'
-
 build:
   dependencies:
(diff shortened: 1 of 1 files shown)


kevinmichaelchen:
Linux builds failing due to missing openssl

```
  thread 'main' panicked at '

  Could not find directory of OpenSSL installation, and this `-sys` crate cannot
  proceed without this knowledge. If OpenSSL is installed and this crate had
  trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the
  compilation process.

  Make sure you also have the development packages of openssl installed.
  For example, `libssl-dev` on Ubuntu or `openssl-devel` on Fedora.
```

Maybe fixed by
```yaml
dependencies:
  linux:
    openssl.org: '*'
```

jhheider:
looks like it depends on `openssl.org: ^1.1`. lots of stuff won't use v3 yet.

PR #2991, opened 2023-08-21
man-db.gitlab.io/man-db

### LINKED ISSUE #2980: +man
We need `man` for various reasons.

* there's https://man-db.gitlab.io/man-db/
* but I wonder what the `man` that eg. macOS pkgs is? I cannot find it. The above is newer and has more deps and more deps is tedious.

and-ri:
closes #2980


and-ri:
Hmmm... The path to the configuration file is hardcoded in the binaries

and-ri:
Maybe to make such stubs? But this seems like a very dirty solution.

```
cat <<EOF > man
  #!/bin/bash
  
  config_file="\$(dirname \$0)/../etc/man_db.conf"
  
  \$(dirname \$0)/../libexec/bin/man --config-file="\$config_file" "\$@"
EOF
```

jhheider:
Will it work off an envvar? The stub can work if needed.

jhheider:
looks great

PR #3017, opened 2023-08-22
introduce github.com/mxcl/swift-sh

Nelyus:
Allows swift script file to declare dependencies

### DIFF
--- a/projects/github.com/mxcl/swift-sh/package.yml
+++ b/projects/github.com/mxcl/swift-sh/package.yml
@@ -1,0 +1,33 @@
+distributable:
+  url: "https://github.com/mxcl/swift-sh/archive/{{version}}.tar.gz"
+  # this obscure option dictates how many components are removed from the
+  # extracted tarball.
+  strip-components: 1
+
+versions:
+  github: mxcl/swift-sh # reads github release tags
+
+platforms:
+  - darwin
+  # - linux/x86-64
+
+dependencies:
+  # swift.org: 5
+
+build:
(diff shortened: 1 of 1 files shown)


Nelyus:
here is a couple links that might be useful

- the repo on GitHub  
https://github.com/mxcl/swift-sh
- The Homebrew formula page  
https://formulae.brew.sh/formula/swift-sh#default
- The Homebrew formula definition  
https://github.com/Homebrew/homebrew-core/blob/e7227eab9c783d94557a5bbdcc17bd27e889c014/Formula/s/swift-sh.rb

Nelyus:
I don't understand the failures

jhheider:
GitHub actions breakdown. They happen surprisingly often.

jhheider:
the new one is more interesting. x86-64 is having a different linking issue than aarch64. that warrants more research.

mxcl:
lol I guess I can probably help with this one.

jhheider:
i don't know, man. pretty complex stuff.

mxcl:
Sounds like either Xcode isn't installed (but I'm sure it is now right?) or `libswiftCore.dylib` is in a weird place. We'll have to dive into brewkit I think.

jhheider:
it's possible that my fix for file.dylib paths in binaries should have been to stop it from changing them to @rpath/file.dylib instead of changing them to @rpath/pkg/ver/lib/file.dylib

Nelyus:
🥳

PR #3062, opened 2023-08-24
SpacetimeDB

kevinmichaelchen:
> A database and server combined into one.
>
> It is a relational database system that lets you upload your application logic directly into the database by way of **fancy stored procedures called "modules."**
> 
> Instead of deploying a web or game server that sits in between your clients and your database, your clients connect directly to the database and execute your application logic inside the database itself. You can write all of your permission and authorization logic right inside your module just as you would in a normal server.
> 
> This means that you can write your entire application in a single language, Rust, and **deploy it as a single binary**. **No more microservices, no more containers, no more Kubernetes, no more Docker, no more VMs, no more DevOps, no more infrastructure, no more ops, no more servers.**

https://github.com/clockworklabs/SpacetimeDB
https://spacetimedb.com/

### DIFF
--- a/projects/spacetimedb.com/package.yml
+++ b/projects/spacetimedb.com/package.yml
@@ -6,7 +6,7 @@ provides:
   - bin/spacetime
 
 versions:
-  - 0.0.0
+  - 2023.8.12
   # FIXME once there has been an official release
   # See https://github.com/orgs/teaxyz/discussions/321
   # github: clockworklabs/SpacetimeDB
@@ -23,6 +23,7 @@ build:
     cmake.org: ^3
     perl.org: "*"
     tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
     freedesktop.org/pkg-config: ^0.29
   env:
     linux:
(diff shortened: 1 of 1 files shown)


kevinmichaelchen:
Still in beta, so I'm trying to do something similar for what we did with [**projects/surrealdb.com/package.yml**](https://github.com/teaxyz/pantry/blob/main/projects/surrealdb.com/package.yml).

**UPDATE**: As of today, it doesn't look like `-beta` suffixes are supported yet. See related [discussion](https://github.com/orgs/teaxyz/discussions/321#discussioncomment-4721555).

kevinmichaelchen:
They [vendor](https://github.com/clockworklabs/SpacetimeDB/blob/master/Cargo.toml#L172) v0.10 of the `openssl` crate.

[perl and C compiler are build deps](https://github.com/sfackler/rust-openssl/blob/openssl-v0.10.56/openssl/src/lib.rs#L11-L15).

projects/spacetimedb.com/package.yml
```diff
@@ -0,0 +1,37 @@
+distributable:
+  url: https://github.com/clockworklabs/SpacetimeDB/archive/refs/tags/v0.6.1-beta.tar.gz
+  strip-components: 1
+
+provides:
+  - bin/spacetime
+
+versions:
+  - 0.0.0
```
jhheider: This should probably be calver. Tea treats all calver as less than non-calver semver.

kevinmichaelchen:
Finally 😅

PR #3092, opened 2023-08-25
+stego -- steganographic swiss army knife

kevinmichaelchen:
https://github.com/ajmwagar/stego
https://crates.io/crates/stego

There are no tags or releases, so we're building a specific Git commit SHA.

### DIFF
--- a/projects/crates.io/stego/package.yml
+++ b/projects/crates.io/stego/package.yml
@@ -1,16 +1,15 @@
 distributable:
-  url: https://github.com/ajmwagar/stego/archive/d9d5911f4d2d141fea74936f235a74bf03961c71.zip
+  url: https://github.com/ajmwagar/stego/archive/d9d5911f4d2d141fea74936f235a74bf03961c71.tar.gz
   strip-components: 1
 
 provides:
   - bin/stego
 
 versions:
-  - 0.0.0
+  - 2019.10.22
   # FIXME once there has been an official release
 
 build:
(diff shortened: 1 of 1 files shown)


projects/crates.io/stego/package.yml
```diff
@@ -0,0 +1,22 @@
+distributable:
+  url: https://github.com/ajmwagar/stego/archive/d9d5911f4d2d141fea74936f235a74bf03961c71.zip
+  strip-components: 1
+
+provides:
+  - bin/stego
+
+versions:
+  - 0.0.0
```
jhheider: Current convention is that this be the calver of the commit date (2023.08.25, for example).

projects/crates.io/stego/package.yml
```diff
@@ -0,0 +1,22 @@
+distributable:
+  url: https://github.com/ajmwagar/stego/archive/d9d5911f4d2d141fea74936f235a74bf03961c71.zip
```
jhheider: Prefer .tar.gz so strip-components works

projects/crates.io/stego/package.yml
```diff
@@ -0,0 +1,22 @@
...
+
+provides:
+  - bin/stego
+
+versions:
+  - 0.0.0
+  # FIXME once there has been an official release
+
+build:
+  working-directory: stego-d9d5911f4d2d141fea74936f235a74bf03961c71
```
jhheider: This won't be necessary with a stripped tar.gz download.

jhheider (changes requested):

PR #3093, opened 2023-08-26
+huniq -- Filter out duplicates on the command line

kevinmichaelchen:
https://github.com/koraa/huniq
https://crates.io/crates/huniq

Filter out duplicates on the command line. Replacement for `sort | uniq` optimized for speed (10x faster) when sorting is not needed.


kevinmichaelchen:
References a specific Git commit hash as mentioned in https://github.com/teaxyz/pantry/issues/3081

jhheider:
Same remarks as stego for best practices.

PR #3095, opened 2023-08-26
cocogitto -- Conventional Commits toolbox

kevinmichaelchen:
https://github.com/cocogitto/cocogitto
https://docs.cocogitto.io/


kevinmichaelchen:
I guess URL could also be https://docs.cocogitto.io/

jhheider:
Probably just call it cocogitto.io at that point.

PR #3104, opened 2023-08-26
+steampipe -- Query cloud resources with SQL (AWS, Azure, GCP and more)

kevinmichaelchen:
https://github.com/turbot/steampipe
https://steampipe.io/

Use SQL to instantly query your cloud services (AWS, Azure, GCP and more). Open source CLI. No DB required.


jhheider:
Probably need to sudo -u nobody on Linux/x86-64.

jhheider:
Hm. Well, you may just have to skip the test if `id` says you're 0.

PR #3122, opened 2023-08-26
+dog -- A command-line DNS client.

kevinmichaelchen:
https://github.com/ogham/dog
https://dns.lookup.dog/

> A command-line DNS client.

### DIFF
--- a/projects/dns.lookup.dog/package.yml
+++ b/projects/dns.lookup.dog/package.yml
@@ -13,7 +13,8 @@ build:
   dependencies:
     rust-lang.org: '>=1.56'
     rust-lang.org/cargo: '*'
-    cmake.org: ^3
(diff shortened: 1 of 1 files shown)


kevinmichaelchen:
i guess openssl also needs to be declared a runtime dep?

projects/dns.lookup.dog/package.yml
```diff
@@ -0,0 +1,27 @@
...
+    rust-lang.org: '>=1.56'
+    rust-lang.org/cargo: '*'
+    freedesktop.org/pkg-config: '*'
+    openssl.org: ^1.1
+  script: |
+    cargo install --locked --path . --root {{prefix}}
+
+test:
+  dependencies:
+    openssl.org: ^1.1
```
jhheider: This should be a runtime dep, right?

PR #3129, opened 2023-08-26
+check-if-email-exists -- Check if an email address exists without sending any email

kevinmichaelchen:
https://github.com/reacherhq/check-if-email-exists
https://reacher.email/

**Open-Source Email Verification API**

> Check if an email address exists without sending any email, written in Rust. Comes with a ⚙️ HTTP backend.

### DIFF
--- a/projects/reacher.email/check-if-email-exists-cli/package.yml
+++ b/projects/reacher.email/check-if-email-exists-cli/package.yml
@@ -17,6 +17,8 @@ build:
   dependencies:
     rust-lang.org: '>=1.65'
     rust-lang.org/cargo: '*'
+    perl.org: '*'
+    tea.xyz/gx/make: '*'
   script:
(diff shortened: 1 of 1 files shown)


projects/reacher.email/check-if-email-exists-cli/package.yml
```diff
@@ -0,0 +1,29 @@
...
+    perl.org: '*'
+    tea.xyz/gx/make: '*'
+  script:
+    cargo install --path . --root {{prefix}}
+
+test:
+  script:
+    - which check_if_email_exists
+    # The CLI version is misaligned with the repo releases/tags
+    # - test "$(check_if_email_exists --version)" = "check-if-email-exists-cli 0.8.32"
```
jhheider: There's probably a way to fix this. I'd grep the sources for that version, then probably use sed to alter it during build.

PR #3163, opened 2023-08-29
+htslib.org

magnusviri:
I don't know how to use this. I just packaged it up following their install instructions and looking at the Brew formula.


magnusviri:
It looks like it is still failing. `dyld: Library not loaded: @rpath/sourceware.org/bzip2/v1/lib/libbz2.dylib` The [install page](https://github.com/samtools/htslib/blob/develop/INSTALL) says it needs libbz2 and liblzma. I probably should've added those before I committed.

PR #3186, opened 2023-09-01
+hurl.dev -- Run and test HTTP Requests

kevinmichaelchen:
https://hurl.dev/
https://github.com/Orange-OpenSource/hurl


kevinmichaelchen:
Tested adding `--locked` on macOS and it works.

Will push that up soon.

PR #3187, opened 2023-09-01
sourceforge.net/projects/libtirpc

and-ri:


jhheider:
not quite sure how to tell it to use the libc linkage it finds...

jhheider:
there we go.

PR #3249, opened 2023-09-06
+open-interpreter

sanchitram1:
https://github.com/KillianLucas/open-interpreter/

pretty neat imo


mxcl:
duped https://github.com/teaxyz/pantry/blob/bffdde2291d9a864e33b415ffb11e808449f54f1/projects/openinterpreter.com/package.yml#L4

I didn't PR so possibly that was part of the reason we didn't realise we had already added it. My bad

jhheider:
WE WERE JUST SO EXICTED

PR #3319, opened 2023-09-14
docker-clean -- Bash script to clean Docker containers, images, volumes, and networks

kevinmichaelchen:
https://github.com/ZZROTDesign/docker-clean

### DIFF
--- a/projects/github.com/ZZROTDesign/docker-clean/package.yml
+++ b/projects/github.com/ZZROTDesign/docker-clean/package.yml
@@ -1,0 +1,22 @@
+distributable:
+  url: https://github.com/ZZROTDesign/docker-clean/archive/refs/tags/v{{version}}.tar.gz
+  strip-components: 1
+
+versions:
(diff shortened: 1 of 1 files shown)


projects/github.com/ZZROTDesign/docker-clean/package.yml
```diff
@@ -0,0 +1,22 @@
...
+  strip-components: 1
+
+versions:
+  github: ZZROTDesign/docker-clean
+
+provides:
+  - bin/docker-clean
+
+dependencies:
+  gnu.org/bash: '*'
```
kevinmichaelchen, line 12: @jhheider I think any shell would work

jhheider: depending on bash is pretty tried-and-true to prevent issues, if you _need_ to depend on a shell.

kevinmichaelchen:
10m timeout - i guess github is just busy 😆

PR #3323, opened 2023-09-15
+goreleaser -- Deliver Go binaries as fast and easily as possible

kevinmichaelchen:
https://github.com/goreleaser/goreleaser
https://goreleaser.com/


jhheider:
Probably simpler to just name it `goreleaser.com`.

kevinmichaelchen:
@jhheider yeah wasn't sure what to do - fwiw they've got some other repos/tools as well like chglog and nfpm

https://github.com/goreleaser

I'll defer to you on this one

kevinmichaelchen:
oh i guess they can always just add more nested dirs ... yeah, what you're saying makes more sense

jhheider:
I'd say _in general_ if the project is the domain, that's sufficient. For justification by precedent, I'll select:
```
postgresql.org
postgresql.org/libpq
```
Fortunately, our namespacing gives us excellent flexibility here.

PR #3334, opened 2023-09-16
gnome.org/pygobject

### LINKED ISSUE #520: +pygobject3 (73/548)
Formulae: https://formulae.brew.sh/formula/pygobject3

and-ri:
closes #520

### DIFF
--- a/projects/gnome.org/PyGObject/package.yml
+++ b/projects/gnome.org/PyGObject/package.yml
@@ -2,22 +2,30 @@ distributable:
   url: https://download.gnome.org/sources/pygobject/{{version.marketing}}/pygobject-{{version}}.tar.xz
   strip-components: 1
 versions:
-  gitlab: gitlab.gnome.org:GNOME/pygobject/tags
+  url: https://download.gnome.org/sources/pygobject/cache.json
+  match: /pygobject-\d+\.\d+\.\d+\.tar\.xz/
+  strip:
+    - /^pygobject-/
+    - /\.tar\.xz/
 dependencies:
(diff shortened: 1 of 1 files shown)


projects/gnome.org/PyGObject/package.yml
```diff
@@ -0,0 +1,40 @@
...
+  match: /pygobject-\d+\.\d+\.\d+\.tar\.xz/
+  strip:
+    - /^pygobject-/
+    - /\.tar\.xz/
+dependencies:
+  gnome.org/gobject-introspection: '*'
+  cairographics.org/pycairo: '*'
+runtime:
+  env:
+    PYTHONPATH: "{{prefix}}/lib/python3.11/site-packages:$PYTHONPATH"
```
jhheider: I think using a specific minor could be dangerous in the future. We should be doing:
```yaml
- run: ln -s python{{deps.python.org.version.marketing}} python{{deps.python.org.version.major}}
```
so we can use `version.major` in these envvar.

projects/gnome.org/PyGObject/package.yml
```diff
@@ -0,0 +1,40 @@
...
+build:
+  dependencies:
+    tea.xyz/gx/cc: c99
+    tea.xyz/gx/make: '*'
+    mesonbuild.com: '*'
+    ninja-build.org: '*'
+    freedesktop.org/pkg-config: '*'
+    python.org: ~3.11
```
jhheider: although if we leave it hard-locked to 3.11 (which probably isn't right), it won't matter.

PR #3362, opened 2023-09-19
tcl-lang.org

### LINKED ISSUE #601: +tcl-tk (298/548)
Formulae: https://formulae.brew.sh/formula/tcl-tk

and-ri:
closes #601

### DIFF
--- a/projects/tcl-lang.org/package.yml
+++ b/projects/tcl-lang.org/package.yml
@@ -80,7 +80,6 @@ build:
       - --includedir={{prefix}}/include/tcl-tk
       - --mandir={{prefix}}/share/man
       - --enable-threads
-      - --enable-64bit
     TCL_PACKAGE_PATH: "{{prefix}}/lib"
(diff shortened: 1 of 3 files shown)


and-ri:
config.log:
```
***
configure:5074: gcc -E  conftest.c
conftest.c:27:10: fatal error: 'net/errno.h' file not found
#include <net/errno.h>
         ^~~~~~~~~~~~~
1 error generated.
configure:5080: $? = 1
***
```

Hmm, why isn't it using `sys/errno.h` 🤔

jhheider:
The warnings above that line claim it can't figure out 64bit for the arch. I wonder if there's something more we need to do for darwin+aarch64.

jhheider:
Is it the same or different source as https://github.com/teaxyz/pantry/blob/main/projects/tcl.tk/tcl/package.yml? That one worked enough to build, though this is clearly more involved.

PR #3376, opened 2023-09-20
fix path (rapidjson)

and-ri:


projects/rapidjson.org/package.yml
```diff
@@ -20,6 +20,8 @@ build:
         CMakeLists.txt
     rm CMakeLists.txt.bak
     make . install
+    sed -i.bak "s|$TEA_PREFIX|\$ENV{TEA_PREFIX}|g" {{prefix}}/lib/cmake/RapidJSON/RapidJSONConfig.cmake
```
jhheider: Consider
`    sed -i.bak "s:$(tea --prefix):\$\{_IMPORT_PREFIX\}/../../../..:g" "{{prefix}}"/lib/cmake/fizz/fizz-targets.cmake`
as not relying on the envvar (though both should be fine).

in theory, https://github.com/teaxyz/brewkit/blob/97fae5a4cff96e0c351e81e0e101f3c5767cb507/libexec/fixup.ts#L73C1-L86C2 should be fixing this. if it's not, we might need to revisit it. it might be that rapidjson hasn't rebuild recently, so I'll rebuild it and see if that's the issue.

and-ri:
Hmm... No, this option doesn't work correctly.

```
CMake Error in src/arrow/CMakeLists.txt:
  Imported target "rapidjson::rapidjson" includes non-existent path

    "/../../../../rapidjson.org/v1.1.0/include"
```

jhheider:
hm. looks like the IMPORT_PREFIX didn't make it into the cmake.

and-ri:
Actually rebuilding the package helped! Thank you 👍🏻

PR #3386, opened 2023-09-21
apache.org/arrow

### LINKED ISSUE #578: +apache-arrow (256/548)
Formulae: https://formulae.brew.sh/formula/apache-arrow

and-ri:
closes #578


jhheider:
we're getting to the stuff with max deps now. maybe we should have sorted the list topographically rather than by popularity.

and-ri:
> we're getting to the stuff with max deps now. maybe we should have sorted the list topographically rather than by popularity.

Do you mean this list (#99)?

I try to do everything in order, but many packages require dependencies that are not in this list. That's why I try to first add something that doesn't need a lot of missing deps.

jhheider:
Yup. We tried to get the dependencies in there too, but I think we missed some system-dependent ones.

PR #3405, opened 2023-09-23
remove gx pkgs, deps on brekit v0.46

mxcl:


jhheider:
this build issue can be "fixed" by making `pkg-build spec[]` > `loop { pkg-build $0 }`.

mxcl:
I don't think we really want this to build to be honest lol

jhheider:
no, we don't want to wait 60 hours for it to build. but we _could_ allow `pkg build foo bar` to work even if they have conflicting `build.dependencies[]` entries.

mxcl:
yeah I think the original thinking was PRs would contain only pkgs that need each other.

And that is still a thing we should handle, though it's less common than this sort of thing.

PR #3429, opened 2023-09-25
xkbcommon.org

and-ri:
xkbcommon => https://github.com/teaxyz/pantry/issues/542

### DIFF
--- a/projects/xkbcommon.org/package.yml
+++ b/projects/xkbcommon.org/package.yml
@@ -9,9 +9,6 @@ dependencies:
   x.org/xcb: '*'
   freedesktop.org/XKeyboardConfig: '*'
   gnome.org/libxml2: '*'
-runtime:
-  env:
-    XKB_CONFIG_ROOT: "{{prefix}}/share/X11"
 build:
   dependencies:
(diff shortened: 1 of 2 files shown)


projects/xkbcommon.org/package.yml
```diff
@@ -9,6 +9,10 @@ dependencies:
   x.org/xcb: '*'
   freedesktop.org/XKeyboardConfig: '*'
   gnome.org/libxml2: '*'
+runtime:
+  env:
+    XDG_CONFIG_HOME: "{{deps.freedesktop.org/XKeyboardConfig.prefix}}/share/X11"
+    XKB_CONFIG_EXTRA_PATH: "{{deps.freedesktop.org/XKeyboardConfig.prefix}}/share/locale"
```
jhheider: Isn't is likely that these should be defined on XKeyboardConfig?

and-ri: sure 👍🏻

and-ri: https://github.com/teaxyz/pantry/pull/3430

PR #3445, opened 2023-09-27
chore: Do Go builds work?

kevinmichaelchen:


kevinmichaelchen:
@jhheider I forgot to `rm $FIXTURE.go` -- should we cleanup after we build?

jhheider:
shouldn't be necessary. the testbeds are transient.

PR #3465, opened 2023-09-30
+github.com/containers/skopeo

### LINKED ISSUE #2695: +skopeo
Formulae: https://formulae.brew.sh/formula/skopeo

Skopeo is from the same family as Podman, which is already packaged by tea.

I believe having skopeo packaged by tea would be specially helpful, since the original project doesn't distribute binaries (and that's why I created https://github.com/felipecrs/skopeo-bin).

felipecrs:
Things to note:

1. This skopeo is statically linked, different than [Homebrew's](https://github.com/Homebrew/homebrew-core/blob/f82bfa8133fa7c84a368e5f9cb972fbce9c730f4/Formula/s/skopeo.rb) version.
2. Skopeo's team itself does not encourage static linking, but it is possible and [documented](https://github.com/containers/skopeo/blob/2d1ae1fdb3d88347eec8a17d4415a964ce86d54f/install.md#building-a-static-binary).
3. This applies https://github.com/containers/skopeo/pull/2014 to make the Skopeo binary truly self-contained and portable.
4. This "bottle" is much inspired by my own project [skopeo-bin](https://github.com/felipecrs/skopeo-bin), which aims to provide portable binaries for Skopeo.

Closes #2695

### DIFF
--- a/projects/github.com/containers/skopeo/package.yml
+++ b/projects/github.com/containers/skopeo/package.yml
@@ -14,12 +14,13 @@ dependencies:
 build:
   dependencies:
     go.dev: "^1.18"
-    github.com/cpuguy83/go-md2man: "*"
     tea.xyz/gx/make: "*"
     curl.se: "*"
     git-scm.org: "*"
+  env:
+    DISABLE_DOCS: 1 # disables the generation of man pages with go-md2man
   script: |
-    # bundle default-policy.json into the binary to provide a working out-of-the-box experience
+    # bundles default-policy.json into the binary to provide a working out-of-the-box experience
     # https://github.com/containers/skopeo/pull/2014
     curl -fsSL https://github.com/containers/skopeo/pull/2014.diff | git apply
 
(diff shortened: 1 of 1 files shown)


jhheider:
CI is down for a bit as we're doing some stuff with the runners, @felipecrs . Sorry for the inconvenience!

felipecrs:
> CI is down for a bit as we're doing some stuff with the runners, @felipecrs . Sorry for the inconvenience!

Ops, got it. My bad.

jhheider:
oh, no worries. i was being stealthy. ;)

felipecrs:
The failure is exactly what the PR I apply with curl and git solves. That's weird.

I'm going to take a look.

jhheider:
I believe git doesn't have a non-zero exit code if `git apply` fails. `patch` is recommended for that reason.

felipecrs:
Got it. Thanks for the hint. Let with me.

PR #3479, opened 2023-10-02
+github.com/regclient/regclient

felipecrs:
The regclient project provides these three binaries, but they are not dependent on each other. The author even distributes one docker image for each for each of them, so I think we should do the same.

Something I noted however is that VSC_REF is tagged with dirty, apparently because:

```console
$ git status --porcelain
?? props/
?? tea.yaml
?? xyz.tea.build.sh
```

Any ideas on that?

### DIFF
--- a/projects/github.com/regclient/regclient/package.yml
+++ b/projects/github.com/regclient/regclient/package.yml
@@ -1,0 +1,31 @@
+distributable:
+  url: git+https://github.com/regclient/regclient
+  ref: v{{ version }}
+
+versions:
+  github: regclient/regclient/releases/tags
+
+provides:
+  - bin/regctl
+  - bin/regsync
+  - bin/regbot
+
+dependencies:
+  curl.se/ca-certs: '*'
+
+build:
+  dependencies:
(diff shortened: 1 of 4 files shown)


felipecrs:
Failure in CI seems very related to https://github.com/pkgxdev/brewkit/issues/176.

jhheider:
do they need to be git checkouts? my guess is that branch of brewkit isn't doing a `mkdir -p` like it's supposed to.

felipecrs:
> do they need to be git checkouts? my guess is that branch of brewkit isn't doing a `mkdir -p` like it's supposed to.

Otherwise version won't be set automatically: https://github.com/regclient/regclient/blob/9eb65476834ca155ac69a198576089d7fde4c241/Makefile#L12C1-L12C8

But there is no rush on this.

jhheider:
once brewkit 0.46.2 builds with https://github.com/pkgxdev/brewkit/commit/390d8a87fa8f33324d398d6ea9a70a047357dace it should be fixed

PR #3511, opened 2023-10-03
+opendev.org/opendev/git-review

felipecrs:
This is an invaluable tool for people working with Gerrit, like myself. However, relying on distributions packages is pretty bad because they are very outdated.

Also, installing git-review is normally a pain because it otherwise requires Python and Pip to be installed already.

Refs:

- https://opendev.org/opendev/git-review
- https://github.com/Homebrew/homebrew-core/blob/d1e8148ba17630d07cb7480ca50b7c514dd9e683/Formula/g/git-review.rb

Some questions:

- This package requires `requests`. I believe it also requires CA certs then. Should we add pkgx's ca-certs as dependency and set `REQUESTS_CA_BUNDLE` to it? It will connect to Gerrit instances from everywhere.
- I noticed this in the logs:

```
/tmp/d5ff659d/xyz.tea.python.build/pip-build-env-hs7clopt/overlay/lib/python3.12/site-packages/setuptools/command/develop.py:40: EasyInstallDeprecationWarning: easy_install command is deprecated.
  !!

          ********************************************************************************
          Please avoid running ``setup.py`` and ``easy_install``.
          Instead, use pypa/build, pypa/installer or other
          standards-based tools.

          See https://github.com/pypa/setuptools/issues/917 for details.
          ********************************************************************************

  !!
```


felipecrs:
Thank you!

PR #3518, opened 2023-10-03
fix(ruby)

### LINKED ISSUE #3516: Gem native extension errors on Ruby 3.1.3
I am having issues building gem native extensions on Ruby v3.1.3. I've traced the issue down to rubygems not being available inside the gem's `extconf.rb`.

Causing the following error [on line 30 of extconf.rb](https://github.com/sass/sassc-ruby/blob/master/ext/extconf.rb#L30) when it hits `Gem::Platform`, but rubygems is not loaded.

Failing here: https://github.com/sass/sassc-ruby/blob/master/ext/extconf.rb#L30

```ruby
Installing sassc 2.4.0 with native extensions
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /Users/peter/code/ruby313_test/vendor/bundle/ruby/3.1.0/gems/sassc-2.4.0/ext
/Users/peter/.pkgx/ruby-lang.org/v3.1.3/bin/ruby -I/Users/peter/.pkgx/rubygems.org/v3.4.20/lib extconf.rb
*** extconf.rb failed ***
Could not create Makefile due to some reason, probably lack of necessary
libraries and/or headers.  Check the mkmf.log file for more details.  You may
need configuration options.
``` 

I'm not sure what the root cause is in `pkgx`, but I suspect it might have to do with the following comment:
https://github.com/pkgxdev/pantry/blob/main/projects/ruby-lang.org/package.yml#L50-L55

I've also tested on Ruby v3.2.2 and it doesn't have the same issue.

I've attached [ruby313_test.tar.gz](https://github.com/pkgxdev/pantry/files/12795185/ruby313_test.tar.gz) with a `.ruby-version` and `Gemfile` that demonstrates the bug.

```
$ tar xzvf ruby313_test.tar.gz
$ cd ruby313_test
$ dev
$ bundle
```

jhheider:


mxcl:
what are we fixing? the linked ticket? the creator of that ticket claims it is ok now, is it not?

jhheider:
i think i was stymied trying to get an older version (3.1.3) to re-build with our current configuration; I haven't come back to it, and this work is probably stale at best and incorrect at worst.

mxcl:
This I guess? https://github.com/pkgxdev/pantry/issues/3517

jhheider:
quite so

PR #3530, opened 2023-10-04
llvm & gcc symlinks for eg cc

mxcl:


projects/gnu.org/gcc/package.yml
```diff
@@ -64,14 +64,6 @@ build:
       if: linux
       working-directory: ${{prefix}}
 
-    # weird ass files it makes
-    - run: |
-        if test "{{hw.platform}}+{{hw.arch}}" = "darwin+aarch64"; then
```
mxcl: I can very much see them in the installed bottle I have tho.

PR #3537, opened 2023-10-05
`git foo` will automatically install and run `git-foo`

mxcl:


mxcl:
@jhheider maybe take a once over of this in case I am doing something insanely stupid

jhheider:
doesn't _seem_ crazy...

PR #3545, opened 2023-10-06
qt.io/qt5

### LINKED ISSUE #542: +qt@5 (168/548)
Formulae: https://formulae.brew.sh/formula/qt@5

and-ri:
closes #542


mxcl:
should this just be `qt`? Why is it versioned?

and-ri:
> should this just be `qt`? Why is it versioned?

It's just that `qt@5' is often found in other packages. I thought maybe it would be more convenient

But I have already changed to qt 👌🏻

mxcl:
other package manager don’t implicitly support versioning. We avoid this sort of thing. We have a couple of versioned packages and some stuff like yarn has two packages.

I dunno maybe I'm making the wrong call here @jhheider thoughts?

jhheider:
Certain enormous frameworks seem to be done this way, presumably because people like hanging onto old majors. It _shouldn't be necessary_ with `pkgx`.

Other things that often follow this protocol for which we have not done it:
- php
- node
- postgres

mxcl:
it's the first Qt we have added so we can cope if we must rename it.

jhheider:
```
dyld[99827]: Library not loaded: @rpath/libzstd.1.dylib
  Referenced from: <E3661C4D-CA6D-3C5D-93A6-465E94A03042> /opt/facebook.com/zstd/v1.5.5/bin/zstd
```
can confirm that zstd seems to have the wrong path. let me see if i can fix that.

jhheider:
that solved _some_ of the issues. that version tagging issue needs a DEFINE... somehow.

jhheider:
well, that's a lot closer, i guess.

and-ri:
image
We don't have the option to specify executables separately for platforms? Because they are present for Linux

and-ri:
image
Eh, how I don't like these cases when locally it is built successfully, but not here

mxcl:
> We don't have the option to specify executables separately for platforms? Because they are present for Linux

We should have that for sure. Although I have used Qt before and it seems odd that these are not on mac.

jhheider:
We definitely already do: https://github.com/pkgxdev/pantry/blob/main/projects/cmake.org/package.yml

jhheider:
```
qxkbcommon.cpp:279:16: error: use of undeclared identifier 'XKB_KEY_dead_longsolidusoverlay'; did you mean 'Qt::Key_Dead_Longsolidusoverlay'?
```
that just seems like:
- it's not finding xkbcommon.org
- or a version conflict

and-ri:
I just want to clarify. Does the error say that it is necessary to add `gnu.org/make` to the dependencies in the test?

```
+ /github/home/.pkgx/pkgx.sh/v1.0.2/bin/pkgx -- make
Warning: `make` is not an explicit dependency!
× unexpected error
│ NotFound
│     at default (file:///home/builder/actions-runner/_work/pantry/pantry/builds/pkgx.sh-1.0.2+linux/src/utils/execve.ts:21:15)
│     at default (file:///home/builder/actions-runner/_work/pantry/pantry/builds/pkgx.sh-1.0.2+linux/src/modes/x.ts:39:3)
│     at eventLoopTick (ext:core/01_core.js:183:11)
│     at async default (file:///home/builder/actions-runner/_work/pantry/pantry/builds/pkgx.sh-1.0.2+linux/src/app.ts:43:11)
│     at async file:///home/builder/actions-runner/_work/pantry/pantry/builds/pkgx.sh-1.0.2+linux/entrypoint.ts:59:3
╰─➤ https://github.com/pkgxdev/pkgx/issues/new
Error: Process completed with exit code 1.
```

jhheider:
> I just want to clarify. Does the error say that it is necessary to add `gnu.org/make` to the dependencies in the test?

That error says make was used in the test, but isn't a dependency; therefore, if qt5 requires make for normal operation, it should add it. If it's just needed for the test, then it need not.

and-ri:
I think it is needed only for this test
```
  script:
    - qmake hello.pro
    - make # <========
    - run: otool -l hello
      if: darwin
    - run: ldd hello
      if: linux
    - ./hello
   ```

Then I'll add `gnu.org/make` to the test section

jhheider:
> Then I'll add `gnu.org/make` to the test section

shouldn't be needed. the warning is just so _you_ know that the test is passing since `make` is present, but if make is normally needed, then you'll have to add it.

that said, that subsequent error is worrying; if it's because make isn't loaded, then maybe adding it will fix it. worth a test; too bad qt takes _so long_ to build.

jhheider:
oooh, looks like you might have it!

mxcl (approved):
Amazing effort

jhheider:
Yes, great work, @ArionThinker !

PR #3549, opened 2023-10-06
fix(kubectx): build the package 

aslafy-z:
I understand this package is vendored Instead of rebuilt. I wanna give a go to mutate it to the recommended pattern.
This is draft.


aslafy-z:
I've just noticed that these files are indeed copied, but in their bash versions, not the go ones. Since the rewritten Go version hasn't reached stability yet, I'm closing this pull request.

PR #3566, opened 2023-10-07
github.com/helmfile/vals

aslafy-z:
package for https://github.com/helmfile/vals cli tool

### DIFF
--- a/projects/github.com/helmfile/vals/package.yml
+++ b/projects/github.com/helmfile/vals/package.yml
@@ -26,6 +26,6 @@ build:
 
 test:
   - vals --help | grep "vals is a Helm-like configuration"
-  # Some versions of vals print had no command to print the version
+  # Some versions of vals had no command to print the version
   - echo $(vals version || true) | grep "v{{version}}" || \
     vals --help | grep -v "Print vals version"
--- a/projects/github.com/helmfile/vals/package.yml
+++ b/projects/github.com/helmfile/vals/package.yml
@@ -16,7 +16,6 @@ build:
    - mkdir -p "{{ prefix }}"/bin
    - go build -trimpath -ldflags="$LDFLAGS" -o "{{ prefix }}"/bin/vals ./cmd/vals
   env:
-    CGO_ENABLED: 0
     LDFLAGS:
       - -w
       - -s
--- a/projects/github.com/helmfile/vals/package.yml
+++ b/projects/github.com/helmfile/vals/package.yml
@@ -25,7 +25,7 @@ build:
       - -X main.version=v{{version}}
 
 test:
-  - vals | grep "vals is a Helm-like configuration"
+  - vals --help | grep "vals is a Helm-like configuration"
   # Some versions of vals print had no command to print the version
   - echo $(vals version || true) | grep "v{{version}}" || \
-    vals | grep -v "Print vals version"
+    vals --help | grep -v "Print vals version"
--- a/projects/github.com/helmfile/vals/package.yml
+++ b/projects/github.com/helmfile/vals/package.yml
@@ -25,7 +25,7 @@ build:
       - -X main.version=v{{version}}
 
 test:
-  - vals --help | grep "vals is a Helm-like configuration"
+  - vals --help 2>&1 | grep "vals is a Helm-like configuration"
   # Some versions of vals had no command to print the version
   - echo $(vals version || true) | grep "v{{version}}" || \
-    vals --help | grep -v "Print vals version"
+    vals --help 2>&1 | grep -v "Print vals version"
--- a/projects/github.com/helmfile/vals/package.yml
+++ b/projects/github.com/helmfile/vals/package.yml
@@ -1,0 +1,31 @@
+distributable:
+  url: https://github.com/helmfile/vals/archive/refs/tags/v{{version}}.tar.gz
+  strip-components: 1
+
+display-name: pinniped
+
+versions:
+  github: helmfile/vals/releases/tags
+
+provides:
+  - bin/vals
+
+build:
+  dependencies:
+    go.dev: '*'
+  script:
+   - go mod download
+   - mkdir -p "{{ prefix }}"/bin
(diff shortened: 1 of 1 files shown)


projects/github.com/helmfile/vals/package.yml
```diff
@@ -0,0 +1,31 @@
...
+    CGO_ENABLED: 0
+    LDFLAGS:
+      - -w
+      - -s
+      - -X main.version=v{{version}}
+
+test:
+  - vals --help | grep "vals is a Helm-like configuration"
+  # Some versions of vals had no command to print the version
+  - echo $(vals version || true) | grep "v{{version}}" || \
```
jhheider, line 30: likely need a `vals version 2>&1 || true...` here.

aslafy-z: It looks like `vals --help` outputs to stderr. Updated

projects/github.com/helmfile/vals/package.yml
```diff
@@ -0,0 +1,31 @@
+distributable:
+  url: https://github.com/helmfile/vals/archive/refs/tags/v{{version}}.tar.gz
+  strip-components: 1
+
+display-name: pinniped
```
jhheider: this is from the prior PR, right? ;)

aslafy-z: Hehe nice catch thanks!

aslafy-z: I removed it, I guess by default the last component of the package path is used, right ?

aslafy-z: For the pinniped PKG, I mixed the sample https://github.com/pkgxdev/brewkit/blob/main/share/TEMPLATE.pkg.yml and some other packages manifests. However, this field is not present in the sample. It is used somewhere ?

jhheider: > I removed it, I guess by default the last component of the package path is used, right ?

yup.

> However, this field is not present in the sample. It is used somewhere ?

initially it was for the ossapp gui, but it is also used in short-form commands. we should add it to the template.

projects/github.com/helmfile/vals/package.yml
```diff
@@ -0,0 +1,28 @@
...
+    go.dev: '*'
+  script:
+   - go mod download
+   - mkdir -p "{{ prefix }}"/bin
+   - go build -trimpath -ldflags="$LDFLAGS" -o "{{ prefix }}"/bin/vals ./cmd/vals
+  env:
+    LDFLAGS:
+      - -w
+      - -s
+      - -X main.version=v{{version}}
```
jhheider, line 22: probably need to add
```yaml
    linux:
      ARGS:
        - -buildmode=pie
```

if using cgo, to prevent segfaults.

aslafy-z: Any recommendation? Should we prefer building with cgo or without? In their makefile, vals have cgo enabled.

projects/github.com/helmfile/vals/package.yml
aslafy-z: https://github.com/pkgxdev/brewkit/issues/183

projects/github.com/helmfile/vals/package.yml
aslafy-z: Mmh, I was looking to the Makefile, however, goreleaser builds the production releases

https://github.com/helmfile/vals/blob/57ae35b137ae8b95b979b262df81433eb349b84a/.goreleaser.yml#L6

CGO is disabled

jhheider: there are some [interesting arguments against it](https://dave.cheney.net/tag/cgo), but if it works, it works. ;)

projects/github.com/helmfile/vals/package.yml
```diff
@@ -0,0 +1,31 @@
...
+    go.dev: '*'
+  script:
+   - go mod download
+   - mkdir -p "{{ prefix }}"/bin
+   - go build $GOARGS -trimpath -ldflags="$LDFLAGS" -o "{{ prefix }}"/bin/vals ./cmd/vals
+  env:
+    LDFLAGS:
+      - -w
+      - -s
+      - -X main.version=v{{version}}
```
aslafy-z, line 22:
```suggestion
  dependencies:
    go.dev: '*'
    curl.se: '*'
    stedolan.github.io/jq: '*'
    
  script:
    - curl -fsS "https://api.github.com/repos/helmfile/vals/git/tags/v{{version}}" | jq '.object.sha' > .commit_sha
    - go mod download
    - mkdir -p "{{ prefix }}"/bin
    - go build $GOARGS -trimpath -ldflags="$LDFLAGS" -o "{{ prefix }}"/bin/vals ./cmd/vals
  env:
    LDFLAGS:
      - -w
      - -s
      - -X main.version=v{{version}} -X main.commit=$(cat .commit_sha)
```

`vals version` shows the commit sha https://github.com/helmfile/vals/blob/57ae35b137ae8b95b979b262df81433eb349b84a/.goreleaser.yml#L8. What do you think of fetching it from GitHub?

jhheider: that'd be the usual way to do it. i dislike things requiring commit sha, but using the git+https distributable and git-scm.org build dep will do it.

jhheider: your curl > jq is elegant for letting us keep using the tarball source.

aslafy-z: That was the point of implementing it like that, yup.
What are the available moustache variables? 
Tags are fetched so maybe the associated Sha can be interpolated too?

jhheider: we don't have a moustache for it. in theory, it would be something returnable, but only for certain types of versions, so it'd be empty a lot of the time. for better or worse, many things are not on github.

PR #3570, opened 2023-10-08
fix(gcc).symlinks

jhheider:
I believe these should be the binutils versions

### DIFF
--- a/projects/gnu.org/gcc/package.yml
+++ b/projects/gnu.org/gcc/package.yml
@@ -7,8 +7,7 @@ versions:
   strip: /^releases\/gcc-/
 
 dependencies:
-  linux:
-    gnu.org/binutils: '*' # linker needs `as`
+  gnu.org/binutils: '*' # linker needs `as`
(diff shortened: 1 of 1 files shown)


projects/gnu.org/gcc/package.yml
```diff
@@ -69,9 +68,9 @@ build:
     # means we're building with the exact same version FIXME
     - run: |
         ln -sf gcc cc
-        ln -sf gcc-ar ar
-        ln -sf gcc-nm nm
-        ln -sf gcc-ranlib ranlib
+        ln -sf ../../../binutils/v\*/bin/ar ar
```
mxcl, line 71: Probs should be a shim to ensure the correct versions are used. But let’s merge for now to see if this works for you or not. (specifically `pkgx +gnu.org/binutils ar "$@"` *will use* the current binutils rather than change the version).

The question becomes: who do gnu provide these if we shouldn't use them?

jhheider: It's a really great question I have no good answer for.

PR #3583, opened 2023-10-09
+jetporch.com

tannevaled:


jhheider (commented):
Package name should be `jetporch.com`. What are the openssl/zlib comments indicating? If either is used during the build, they should be in the build dependencies, at a minimum. And they likely don't specifically require the very newest versions; generally ^major is sufficient.

tannevaled:
> Package name should be `jetporch.com`. What are the openssl/zlib comments indicating? If either is used during the build, they should be in the build dependencies, at a minimum. And they likely don't specifically require the very newest versions; generally ^major is sufficient.

when trying to link dynamicaly the test failed on debian regarding the openssl library version. i linked statically to make it work.

jhheider:
Static linking is generally good on rust programs. Does it really require openssl^3 or would openssl: '*' work? in general, the most permissive choice possible is best.

tannevaled:
> Static linking is generally good on rust programs. Does it really require openssl^3 or would openssl: '*' work? in general, the most permissive choice possible is best.

with static linking il build with the default version.

PR #3600, opened 2023-10-10
Add firebase-tools to npmjs.com/provider.yml

svenjacobs:
As mentioned [here](https://github.com/orgs/pkgxdev/discussions/793).

### DIFF
--- a/projects/npmjs.com/provider.yml
+++ b/projects/npmjs.com/provider.yml
@@ -8,7 +8,9 @@ cmds:
   - vercel
   - changelogithub
   - get-port
+  - firebase
 args:
   chalk: npx --yes --package=chalk-cli --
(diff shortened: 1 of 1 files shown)


mxcl:
note that this doesn’t work great RN. It will be suggested for command-not-found, but typing `x` doesn't run it.

It does work `pkgx firebase foo`, but that's it.

If you want firebase to be a tool that can be added to environments, it needs to be fully packaged. For node packages this is quite easy, there's a few examples in the pantry you can riff off.

Longer term I would like to make “providers” more proper, but I haven't had the ideas about that yet. I will at least fix the `x` bug.

svenjacobs:
> If you want firebase to be a tool that can be added to environments, it needs to be fully packaged.

Thanks, I will work on that!

PR #3602, opened 2023-10-10
+supabase.com/cli

j1philli:


j1philli:
oh shoot, this is pulling pre release builds of supabase. will fix that up

Edit:
doesn't seem like supabase/cli makes this easy and pantry versioning is basic so probably okay to just leave it the way it is unless there are objections

jhheider:
yeah, their versions aren't helpful for determining pre-releases vs releases. hopefully fine. thanks for the contribution!

PR #3604, opened 2023-10-10
Package google.com/firebase-tools

svenjacobs:
A follow up to #3600, now the right way.
Hope everything is right because this is my first package.

### DIFF
--- a/projects/google.com/firebase-tools/package.yml
+++ b/projects/google.com/firebase-tools/package.yml
@@ -8,17 +8,17 @@ versions:
   github: firebase/firebase-tools
 
 dependencies:
-  nodejs.org: '*'
+  nodejs.org: "*"
 
 build:
   dependencies:
-    npmjs.com: '*'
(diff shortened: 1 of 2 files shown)


svenjacobs:
Hm, something is not right. The tests worked on my machine.

mxcl:
yeah agreed, this looks like it should work

jhheider:
Look at the bottle extraction: https://github.com/pkgxdev/pantry/actions/runs/6471355430/job/17569763324?pr=3604

it's missing most of the files it's likely to need, i believe. it may require `--install-links`, which as i recall is badly named, and keeps changing from default to non-default behavior. we have a variety of npm packages in the pantry, and no real agreement between them which flags are used :(

svenjacobs:
Thanks @jhheider, that did the trick!

jhheider:
thanks for the contribution, @svenjacobs !

PR #3653, opened 2023-10-13
+gitsign -- Keyless Git signing with Sigstore

spencergilbert:
[gitsign](https://github.com/sigstore/gitsign)

### DIFF
--- a/projects/sigstore.dev/gitsign/package.yml
+++ b/projects/sigstore.dev/gitsign/package.yml
@@ -5,22 +5,19 @@ distributable:
 versions:
   github: sigstore/gitsign
 
-dependecies:
-  git-scm.org: ^2
-
 build:
   dependencies:
     go.dev: ^1.21
   script:
     - go mod download
     - mkdir -p "{{ prefix }}"/bin
-    - go build -v -trimpath -ldflags="$LDFLAGS" -o "$BUILDLOC"
+    - go build -v -trimpath -ldflags="$LDFLAGS" -o "{{prefix}}"/bin/gitsign
+    - go build -v -trimpath -ldflags="$LDFLAGS" -o "{{prefix}}"/bin/gitsign-credential-cache ./cmd/gitsign-credential-cache
   env:
     GOPROXY: https://proxy.golang.org,direct
     GOSUMDB: sum.golang.org
     GO111MODULE: on
     CGO_ENABLED: 0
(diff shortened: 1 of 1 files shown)


jhheider:
Looks like it needs a top-level:
```yaml
dependecies:
  git-scm.org: ^2
```

spencergilbert:
Fixing! I was also wondering if an ["optional" binary](https://github.com/sigstore/gitsign/tree/main/cmd/gitsign-credential-cache) should be packaged separately or included here?

EDIT: Also confused why including the dependency just in the test section doesn't work :thinking:

jhheider:
If it's from the same source, I'd include it.

jhheider:
You copied in my typo ;)

`depende*n*cies`

spencergilbert:
It turns out spelling is quite difficult :sweat_smile: thanks for the hand-holding!

jhheider:
no problem. we _love_ our contributors.

spencergilbert:
Ah, I'm definitely thinking of git as an outside of `pkgx` tool - thanks again

jhheider:
If we do everything right, everything we package should be able to run with our packaged code. Not 100%, but reeeeeeeeally close.

jhheider:
_et voila!_
```shell
$ pkgx --sync gitsign --help
 ✓  ~/.pkgx/sigstore.dev/gitsign/v0.7.1
Keyless Git signing with Sigstore!

Usage:
  gitsign [flags]
  gitsign [command]

Available Commands:
  attest      add attestations to Git objects
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  show        Show source predicate information
  verify      Verify a commit
  version     print Gitsign version

Flags:
  -a, --armor               create ascii armored output
  -b, --detach-sign         make a detached signature
  -h, --help                help for gitsign
      --include-certs int   -3 is the same as -2, but omits issuer when cert has Authority Information Access extension. -2 includes all certs except root. -1 includes all certs. 0 includes no certs. 1 includes leaf cert. >1 includes n from the leaf. Default -2. (default -2)
  -u, --local-user string   use USER-ID to sign
  -s, --sign                make a signature
      --status-fd int       write special status strings to the file descriptor n. (default -1)
  -v, --verify              verify a signature
      --version             print Gitsign version

Use "gitsign [command] --help" for more information about a command.
```

PR #3659, opened 2023-10-14
+oapi-codegen -- Go OpenAPI client & server generator

mycrEEpy:
Adding [oapi-codegen](https://github.com/deepmap/oapi-codegen)


jhheider:
Looks great. Thanks!

PR #3666, opened 2023-10-15
+operator-sdk -- SDK for building Kubernetes applications

mycrEEpy:
Adding [operator-sdk](https://sdk.operatorframework.io)

### DIFF
--- a/projects/operatorframework.io/operator-sdk/package.yml
+++ b/projects/operatorframework.io/operator-sdk/package.yml
@@ -1,8 +1,8 @@
 # https://sdk.operatorframework.io/
 
-distributable: ~
-warnings:
-  - vendored
+distributable:
+  url: git+https://github.com/operator-framework/operator-sdk
+  ref: v{{version}}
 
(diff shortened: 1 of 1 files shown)


jhheider:
Is there a good reason not to build from source? Looks like it's a simple `go` program:

https://sdk.operatorframework.io/docs/contribution-guidelines/developer-guide/

mycrEEpy:
If it's more desirable to build from source I can change it, just thought since the releases are coming with pre-built binaries, why not use them.

jhheider:
it's a valid question; we generally build our own binaries so we can link against our libraries and checksum/sign our bottles with (reasonable) confidence. we'd prefer having signatures on vendored binaries, but those aren't widely available. so, vendored binaries are a second-best option where possible.

mycrEEpy:
@jhheider i've changed it to build from source

jhheider:
very nice! thanks for the contribution, @mycrEEpy !

PR #3667, opened 2023-10-15
feat(kpt.dev): add kpt package to pantry

vkumbhar94:


projects/kpt.dev/package.yml
```diff
@@ -0,0 +1,31 @@
+# https://kpt.dev/
+
+distributable: ~
+warnings:
+  - vendored
+
+versions:
+  github: kptdev/kpt/tags
```
vkumbhar94, line 8: I couldn't make it to install latest beta version. How can I @jhheider @mxcl?

```
pkg build kpt.dev@v1.0.0-beta.46
error: Uncaught Error: invalid semver range: @v1.0.0-beta.46
      const err = () => new Error(`invalid semver range: ${input}`)
                        ^
    at err (https://deno.land/x/libpkgx@v0.14.1/src/utils/semver.ts:137:25)
    at new Range (https://deno.land/x/libpkgx@v0.14.1/src/utils/semver.ts:196:49)
    at Module.parse (https://deno.land/x/libpkgx@v0.14.1/src/utils/pkg.ts:11:22)
    at file:///Users/vaibhavkumbhar/.pkgx/pkgx.sh/brewkit/v0.49.0/libexec/find.ts:8:43
```

jhheider: at present, we only take semver versions without build or pre-release identifiers. we haven't yet had a clamor to add this, but it could be possible.

jhheider:
Is there a good reason not to build from source, since it's just a `go` binary?

https://kpt.dev/installation/kpt-cli?id=source

projects/kpt.dev/package.yml
mxcl: We haven't yet decided how to handle betas since we haven't had a good example of one. Is this a good example of one we should support? @vkumbhar94

vkumbhar94: Sounds like a good example as in the last stable version 0.39.3 was released around 2 yrs ago.
Since then v1.X.X beta versions; last beta released just 5 days ago.
I am just thinking out loud that it is going through some major redesign so they haven't yet called out a stable v1, they are just continuing beta.
But if we look at it from the user's point of view, there would be several features made available in beta since the last 2 yrs. Thus one may tend to use the beta version rather than using stable version released a few yrs back.

vkumbhar94:
Building from source brings dep on Go, whereas to install and use kpt one shouldn't require golang to be available on their machine.

Moreover, `go install` puts a compiled binary in GOROOT/bin which may lead to several kpt binaries depending on different go versions installed.

IMHO, we shouldn't make it too heavy to use kpt. Using precompiled go binaries works well as far as my experience unless they are dependent on some linux libraries to be available at runtime.

jhheider:
> Building from source brings dep on Go, whereas to install and use kpt one shouldn't require golang to be available on their machine.

only as a build dependency, not for a user to run it.

> Moreover, `go install` puts a compiled binary in GOROOT/bin which may lead to several kpt binaries depending on different go versions installed.

that won't be true of our built product. look at [goreleaser](https://github.com/pkgxdev/pantry/blob/main/projects/goreleaser.com/package.yml) as one example.

> IMHO, we shouldn't make it too heavy to use kpt. Using precompiled go binaries works well as far as my experience unless they are dependent on some linux libraries to be available at runtime.

i agree that pre-compiled binaries can be just fine, but unless we're unable to build them ourselves (reliably) for some reason, we generally build and distribute our binaries to increase supply-chain control.

vkumbhar94:
Thanks @jhheider I was about to work today (on weekend).

jhheider:
hope i didn't steal your thunder. ;) i know people are busy so i like to help out if it looks like that might be welcome.

PR #3669, opened 2023-10-15
+crystal-lang.org

elbywan:
Adds a recipe for the [crystal](https://crystal-lang.org/) language and the [shards](https://github.com/crystal-lang/shards) package manager. 

This is my first time contributing, so sorry if I missed anything 🙇.

### DIFF
--- a/projects/crystal-lang.org/package.yml
+++ b/projects/crystal-lang.org/package.yml
@@ -32,6 +32,7 @@ dependencies:
 runtime:
   env:
     CRYSTAL_PATH: ${{prefix}}/lib:$CRYSTAL_PATH
+    CRYSTAL_LIBRARY_PATH: $LD_LIBRARY_PATH
 
 build:
   dependencies:
@@ -76,3 +77,4 @@ test:
     - cp $FIXTURE test.cr
     - crystal build -o test test.cr
     - test "$(./test)" = "{{version}}"
+    - test "$(crystal eval 'puts {{Crystal::VERSION}}')" = "{{version}}"
--- a/projects/crystal-lang.org/package.yml
+++ b/projects/crystal-lang.org/package.yml
@@ -68,7 +68,4 @@ build:
 provides:
   - bin/crystal
 
-test:
-  env:
-    CRYSTAL_LINK_FLAGS: -Wl,-rpath,{{pkgx.prefix}}
-  script: test $(crystal eval 'puts {{Crystal::VERSION}}') = {{version}}
+test: test $(crystal eval 'puts {{Crystal::VERSION}}') = {{version}}
--- a/projects/crystal-lang.org/package.yml
+++ b/projects/crystal-lang.org/package.yml
@@ -68,4 +68,11 @@ build:
 provides:
   - bin/crystal
 
-test: test $(crystal eval 'puts {{Crystal::VERSION}}') = {{version}}
+test:
+  fixture: puts {{Crystal::VERSION}}
+  env:
+    CRYSTAL_LINK_FLAGS: -Wl,-rpath,{{pkgx.prefix}}
+  script:
+    - cp $FIXTURE test.cr
+    - crystal build -o test test.cr
+    - test "$(./test)" = "{{version}}"
--- a/projects/crystal-lang.org/package.yml
+++ b/projects/crystal-lang.org/package.yml
@@ -59,6 +59,10 @@ build:
       - FLAGS=--no-debug
       - interpreter=true
       - CRYSTAL_CONFIG_PATH=../lib
+    linux:
+      CC: clang
+      CXX: clang++
+      LD: clang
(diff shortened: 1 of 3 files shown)


projects/crystal-lang.org/package.yml
```diff
@@ -0,0 +1,47 @@
+distributable:
+  url: https://github.com/crystal-lang/crystal/archive/refs/tags/{{ version }}.tar.gz
+  strip-components: 1
+
+versions:
+  github: crystal-lang/crystal
+
+# Crystal is needed to build itself…
+# So we cannot build it from source.
```
jhheider, line 9: since we're downloading a binary, can we use that to build from source ourselves?

jhheider: https://crystal-lang.org/install/from_sources/

projects/crystal-lang.org/package.yml
```diff
@@ -0,0 +1,47 @@
...
+  dependencies:
+    curl.se: "*"
+  working-directory: ${{prefix}}
+  script: |
+    curl -Lfo crystal.tar.gz "https://github.com/crystal-lang/crystal/releases/download/{{version}}/crystal-{{version}}-1-$PLATFORM.tar.gz"
+    tar --strip-components=1 -xf crystal.tar.gz
+    rm crystal.tar.gz
+  env:
+    darwin/aarch64: { PLATFORM: darwin-universal }
+    darwin/x86-64: { PLATFORM: darwin-universal }
```
jhheider: these can be combined to a `darwin:` key.

projects/crystal-lang.org/package.yml
elbywan: > since we're downloading a binary, can we use that to build from source ourselves?

But since we need to download the binary anyway, what would be the point to download and then compile it from source? It would produce it twice.

> https://crystal-lang.org/install/from_sources/

It is stated there:

> [Install the latest Crystal release](https://crystal-lang.org/install). To compile Crystal, you need Crystal :).

projects/crystal-lang.org/package.yml
elbywan: 👍

projects/crystal-lang.org/package.yml
jhheider: Right. That's how we build rust, and a few others I believe. They have bootstrapping built into their build process. Used to do it for Haskell but their build systems are in constant flux and it got too unstable.

In general we prefer to build our own packages, unless unavoidable.

elbywan: Ah I understand, there is a [repository](https://github.com/crystal-lang/bootstrap-script) to bootstrap the compiler but it seems pretty outdated :(.

https://github.com/crystal-lang/bootstrap-script/issues/2
https://forum.crystal-lang.org/t/segfault-while-bootstraping-from-ruby/6073/4

jhheider: Should be able to bootstrap it from itself.

Unless you want to try and get linux+aarch64 working.

elbywan: > Should be able to bootstrap it from itself.

Allright I'll try to do that when I get some free time 👍.

jhheider:
that was a bit of a pain. as predicted. thanks for doing this, @elbywan !

mxcl:
```sh
$ pkgx crystal eval 'puts "Hello pkgx!"'

ld: warning: search path '$LD_LIBRARY_PATH' not found
dyld[46561]: Library not loaded: @rpath/pcre.org/v2/v10.42.0/lib/libpcre2-8.0.dylib
  Referenced from: <B71A33B2-5360-384D-BB27-FDE3EC592899> /Users/mxcl/.cache/crystal/crystal-run-eval.tmp
  Reason: no LC_RPATH's found
Program received and didn't handle signal ABRT (6)
```

Doesn’t work for me.

jhheider:
Yes, there's a fixme comment. Build works fine with the shim, but eval compiles but can't be passed linker flags (that I could find).

mxcl:
k… well that's a shame since that was going to be my tweet demonstrating crystal.

jhheider:
I'll take another look.

mxcl:
> I'll take another look.

No jacob you got enough on man. I'm sure it's a bug report to crystal or major `package.yml` hackery. Just makes it so there isn't a zero dep tweet.

jhheider:
> > I'll take another look.
> 
> No jacob you got enough on man. I'm sure it's a bug report to crystal or major `package.yml` hackery. Just makes it so there isn't a zero dep tweet.

like that would stop me. 6d90ced542b7522bb76ba2f65426982ad679cecd / 0d68a9ea22fc1b3519f4dc60b2178348d794aa8f look like they do it.

PR #3679, opened 2023-10-16
+opendev.org/openstack/python-openstackclient

tannevaled:


projects/opendev.org/openstack/python-openstackclient/package.yml
```diff
@@ -0,0 +1,35 @@
...
+  - bin/openstack-inventory
+
+test:
+  env:
+    TERM: vt100
+  script: |
+    openstack help server list &&
+    test $? = 0 &&
+    openstack-inventory --help &&
+    test $? = 0
```
jhheider: Unless I'm missing something, this is just:
```yaml
  script:
    - openstack help server list
    - openstack-inventory --help
```

right? our test suite fails if any exit code is non-zero.

tannevaled: sure, this is too easy ;)

Issue #265, opened 2023-02-10
+fb303 (327/548)


mfts:
Formulae: https://formulae.brew.sh/formula/fb303
Source: https://github.com/facebook/fb303

PR #3684, opened 2023-10-16
facebook.com/mvfst

and-ri:
mvfst => #265


jhheider:
worth trying `test.env.linux.LD: clang` since i think direct calls to ld.lld are weird.

and-ri:
> worth trying `test.env.linux.LD: clang` since i think direct calls to ld.lld are weird.

Actually I already did it:
image
It's strange that it calls `ld.lld`

jhheider:
[might need a cmake define...](https://stackoverflow.com/questions/1867745/cmake-use-a-custom-linker)

jhheider:
iirc, `__aarch64_cas8_acq_rel` is usually adding `-latomic`.

PR #3696, opened 2023-10-18
GitHub.com/ggerganov/llama.cpp update

magnusviri:
This has 2 commits. One commit switches from a hardcoded version to a github version. Llama.cpp sometimes updates several times a day. Today there have been 7 releases. I have no idea if you want to bottle all of those versions. Maybe it could just bottle up every 10th version or something?

The second commit changes the order of stuff in the llama.cpp shell script. It checks if --fetch is not specified and if --model or -m is specified and if yes, then it runs llama.cpp. Otherwise, it downloads the model and runs llama.cpp with that. [OpenLLaMA-3Bv2.ggml.f16.bin](https://huggingface.co/SlyEcho/open_llama_3b_v2_ggml/resolve/main/open-llama-3b-v2-f16.bin) does not work with the latest version of llama.cpp. The error is:

```
gguf_init_from_file: invalid magic number 67676a74
error loading model: llama_model_loader: failed to load model from /Users/james/.local/share/models/OpenLLaMA/OpenLLaMA-3Bv2.ggml.f16.bin

llama_load_model_from_file: failed to load model
llama_init_from_gpt_params: error: failed to load model '/Users/james/.local/share/models/OpenLLaMA/OpenLLaMA-3Bv2.ggml.f16.bin'
main: error: unable to load model
```

If I specify my own model file it works. I can update the model downloaded by llama-fetch but if I did so, I'd want to change it so that it offers a choice of model files and lets the user pick. For example, here's 3 model files I would offer.

- [dolphin-llama2-7b.Q4_0.gguf](https://huggingface.co/TheBloke/Dolphin-Llama2-7B-GGUF/resolve/main/dolphin-llama2-7b.Q4_0.gguf)
- [dolphin-2.1-mistral-7b.Q4_0.gguf](https://huggingface.co/TheBloke/dolphin-2.1-mistral-7B-GGUF/resolve/main/dolphin-2.1-mistral-7b.Q4_0.gguf)
- [llama-2-7b-chat.Q4_0.gguf](https://huggingface.co/TheBloke/Llama-2-7b-Chat-GGUF/resolve/main/llama-2-7b-chat.Q4_0.gguf)

Ideally, I'd create a tool that used Hugging Face's API's and I'd offer models dynamically. But I know I wont ever do it, I've already over committed my time.


mxcl:
Thanks James,

> The second commit changes the order of stuff in the llama.cpp shell script. It checks if --fetch is not specified and if --model or -m is specified and if yes, then it runs llama.cpp. Otherwise, it downloads the model and runs llama.cpp with that.

Why this change tho?

magnusviri:
> > The second commit changes the order of stuff in the llama.cpp shell script. It checks if --fetch is not specified and if --model or -m is specified and if yes, then it runs llama.cpp. Otherwise, it downloads the model and runs llama.cpp with that.
> 
> Why this change tho?

I deleted the OpenLlama model because it doesn't work and I didn't want it to redownload it. So if I specify the model it wont redownload it. I figured this might be desirable behavior. I think the real solution is to have some separate model downloader. I've kind of talked about this before, but I don't see me working on it any time soon because I'm already behind on other projects.

mxcl:
k so it's currently broken. Unfortunately I don’t see how the user is supposed to figure out what they need to do with these changes. I will need to go in and figure out a solution on top of this PR I think.

mxcl:
They release too often, more than one a day shows complete disregard for users and packagers. Here they just release every commit to `main`! When there is no distinction between “push commit” and “release version” it doesn’t count as releases.

I think we should consider a “build only” package that builds the tool on the user’s computer for these sort of scenarios. Dunno, just spitballing here.

jhheider:
> I think we should consider a “build only” package that builds the tool on the user’s computer for these sort of scenarios.

That'd allow for `we have that package, but not that version bottled; would you like to try and build it locally?` response from the resolver.

projects/github.com/ggerganov/llama.cpp/package.yml
```diff
@@ -32,6 +34,7 @@ build:
       make --jobs {{hw.concurrency}}
       install -D main {{prefix}}/libexec/llama.cpp
       install -D props/llama.cpp {{prefix}}/bin/llama.cpp
+      install -D ggml-metal.metal {{prefix}}/libexec/
```
mxcl: how is this consumed?

mxcl: does it need to be alongside the llama.cpp executable?

mxcl:
none of the provided models work I get:

```
llama_init_from_gpt_params: error: failed to create context with model '/Users/mxcl/.local/share/llama.cpp/llama-2-7b-chat.Q4_0.gguf'
main: error: unable to load model
```

this is with these parameters:

```sh
exec "$D"/bin/llama.cpp \
    --model "$MODEL_DIR/$MODEL_FILENAME" \
    --ctx-size 512 \
    --batch-size 1024 \
    --n-predict 256 \
    --keep 48 \
    --repeat_penalty 1.0 \
    --color \
    --interactive \
    --reverse-prompt "User:" \
    --file "$D"/share/prompts/chat-with-bob.txt
```

le sigh. I'll check the docs, but @magnusviri you know the parameters required?

projects/github.com/ggerganov/llama.cpp/package.yml
magnusviri: Yes. When it's not there it gives this error:

```
llama_new_context_with_model: n_ctx      = 512
llama_new_context_with_model: freq_base  = 10000.0
llama_new_context_with_model: freq_scale = 1
llama_new_context_with_model: kv self size  =   64.00 MB
ggml_metal_init: allocating
ggml_metal_init: found device: Apple M1
ggml_metal_init: picking default device: Apple M1
ggml_metal_init: default.metallib not found, loading from source
ggml_metal_init: loading '(null)'
ggml_metal_init: error: Error Domain=NSCocoaErrorDomain Code=258 "The file name is invalid."
llama_new_context_with_model: ggml_metal_init() failed
llama_init_from_gpt_params: error: failed to create context with model '/Users/u0076374/.cache/lm-studio/models/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/mistral-7b-instruct-v0.1.Q6_K.gguf'
main: error: unable to load model
```

mxcl: cool thanks

magnusviri:
> They release too often, more than one a day shows complete disregard for users and packagers. Here they just release every commit to `main`! When there is no distinction between “push commit” and “release version” it doesn’t count as releases.
> 
> I think we should consider a “build only” package that builds the tool on the user’s computer for these sort of scenarios. Dunno, just spitballing here.

I think it's a great idea and a great way to support it. I think the llama.cpp community is so bleeding edge that there's a trail of dead bodies behind them. They are constantly breaking backwards compatibility and adding crazy stuff. E.g. they just added [LLaVA support](https://github.com/ggerganov/llama.cpp/pull/3436). This is actually why I want a model downloader that is separate because the space moves too fast. 2 months ago is a lifetime ago.

Another thing that would make local building cool is if the user could specify build flags. E.g., llama.cpp has MPI support but only with a build flag. It has extra dependencies also (either MPICH or OpenMPI).

mxcl:
> That'd allow for we have that package, but not that version bottled; would you like to try and build it locally? response from the resolver.

Having shit build automatically in situations where it is necessary is really nice. Brew does that, but that's because bottles came later. We can certainly install brewkit and do it. The build environments are pretty isolated at this point.

magnusviri:
> le sigh. I'll check the docs, but @magnusviri you know the parameters required?

I hate saying this: Works for me?

```
llama.cpp -p "What is the airspeed velocity of an unladen sparrow?" -m /Users/james/AI/models/llama-2-7b-chat.Q4_0.gguf
...
What is the airspeed velocity of an unladen sparrow?
 Einzeln 2022-10-19 (Wednesday) 06:34 (UTC+03:00)

In the movie Monty Python and the Holy Grail, a character named King Arthur asks another character, "What is the airspeed velocity of an unladen sparrow?" to which the other character responds, "What do you mean? An unladen sparrow? Ah, it's got no feathers, now, has it?"
...
```

magnusviri:
You know, my version of llama.cpp is a few days old. Maybe they've broken compatibility with the models I posted already.

magnusviri:
> You know, my version of llama.cpp is a few days old. Maybe they've broken compatibility with the models I posted already.

I just built 2041 and it works too. This version is supposed to have LLaVA support, so I'm going to go try to find out what that is and see if I can get it to work.

mxcl:
hmm ok, must be something I did wrong. Thanks

magnusviri:
Did you try a different model? My experience is this stuff is fragile.

mxcl:
k figured it out, my ggml file was in the wrong place

mxcl:
I worry about building every release without decent QA. We can have decent QA if we download a model, but downloading every time is rude. We'd want to persist it. However our tests don’t persist anything by design and reassign HOME even.

PR #3708, opened 2023-10-18
+edgedb.com

isworking:
Adds a recipe for the [EdgeDB CLI](https://edgedb.com).

It's my first contribution to PKGX, so let me know if I missed something.

### DIFF
--- a/projects/edgedb.com/package.yml
+++ b/projects/edgedb.com/package.yml
@@ -14,7 +14,9 @@ build:
   env:
     RUSTFLAGS: -A warnings -C debuginfo=0
   script: |
-    cargo install --path . --root {{prefix}}
+    mv build.rs build.rs.bak
+    rm Cargo.lock
+    cargo install --path . --locked --root {{prefix}}
 
 test:
   test "$(edgedb --version)" = "EdgeDB CLI {{version.raw}}"
--- a/projects/edgedb.com/package.yml
+++ b/projects/edgedb.com/package.yml
@@ -1,0 +1,23 @@
+distributable:
+  strip-components: 1
+  url: https://github.com/edgedb/edgedb-cli/archive/refs/tags/v{{version.raw}}.tar.gz
+
+versions:
+  github: edgedb/edgedb-cli/tags
+  strip: /v/
+
+build:
+  dependencies:
+    rust-lang.org: ^1.61
+    rust-lang.org/cargo: ^0
+    perl.org: ^5
+  env:
+    RUSTFLAGS: -A warnings -C debuginfo=0
+  script: |
+    cargo install --path . --root {{prefix}}
+
+test:
+  test "$(edgedb --version)" = "EdgeDB CLI {{version.raw}}"
+
+provides:
+  - bin/edgedb
--- a/projects/edgedb.com/package.yml
+++ b/projects/edgedb.com/package.yml
@@ -20,9 +20,7 @@ build:
   script: |
     # Required to avoid conflicts
     mv build.rs build.rs.bak
-    # Required because of proc-macro2 version conflict
-    rm Cargo.lock
-    cargo install --path . --locked --root {{prefix}}
+    cargo install --path . --root {{prefix}}
 
 test:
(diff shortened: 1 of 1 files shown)


jhheider:
looks like it might have a build dependency on `perl.org: ^5`.

isworking:
> looks like it might have a build dependency on `perl.org: ^5`.

I'm trying to build it first with the latest version (to see if it works). If it does, I'll pin the version and do a new commit.

jhheider:
`perl.org: ^5` is the latest version (5.38.0 at the moment).

projects/edgedb.com/package.yml
```diff
@@ -0,0 +1,27 @@
...
+  strip-components: 1
+  url: https://github.com/edgedb/edgedb-cli/archive/refs/tags/v{{version.raw}}.tar.gz
+
+versions:
+  github: edgedb/edgedb-cli/tags
+  strip: /v/
+  
+platforms:
+  - darwin
+  - linux
```
jhheider: this can be omitted, since it builds for all platforms.

projects/edgedb.com/package.yml
```diff
@@ -0,0 +1,27 @@
...
+
+build:
+  dependencies:
+    rust-lang.org: "1.73.0"
+    rust-lang.org/cargo: "0.74.0"
+    perl.org: "*"
```
jhheider: the specific versions are likely too restrictive. Rust is likely `^1.65`, iirc, and cargo and perl should probably be ^0 and ^5 unless we learn better.

projects/edgedb.com/package.yml
```diff
@@ -0,0 +1,27 @@
...
+  script: |
+    mv build.rs build.rs.bak
```
jhheider, line 22: if this is necessary and non-harmful, it really needs a comment explaining it.

projects/edgedb.com/package.yml
```diff
@@ -0,0 +1,27 @@
...
+    rm Cargo.lock
```
jhheider: this can be accomplished by just omitting the `--locked` arg on the next line.

projects/edgedb.com/package.yml
```diff
@@ -0,0 +1,27 @@
...
+    RUSTFLAGS="-C debuginfo=0 -A warnings" cargo install --locked --path . --root {{prefix}}
```
jhheider: these rust flags are better moved to an env key:
```yaml
  env:
    # comment explaining the flags needed.
    RUSTFLAGS:
      - -C debuginfo=0
      - -A warnings
```

jhheider (commented):
looks great! just some style fixes.

isworking:
Sorry, I mistakenly forced push main to this branch

isworking:
@jhheider It's passing all the checks now. Would you give me more feedbacks?

jhheider:
The only other change I'd make is instead of removing Cargo.lock and using --locked, you just need to remove the --locked flag (and then you don't have to remove anything) to get around version lock issues.

isworking:
> The only other change I'd make is instead of removing Cargo.lock and using --locked, you just need to remove the --locked flag (and then you don't have to remove anything) to get around version lock issues.

I've applied these changes and it passes all the checks (and also builds successfully).

jhheider:
looks great! thanks for this, @Rajdeep-TG !

isworking:
I don't know why but the build for `darwin+aarch64` failed (maybe because of some network issue or something).

jhheider:
```json
},
  ok: false,
  redirected: false,
  status: 502,
  statusText: "Bad Gateway",
  url: "https://api.github.com/graphql"
},
```
yep. it'll recover on re-run.

isworking:
Thanks for quick response. Another thing, when will it be available on PKGX?

jhheider:
each platform will be available as soon as the `upload` step succeeds. very soon, in other words.

jhheider:
now: https://github.com/pkgxdev/pantry/actions/runs/6577934184

isworking:
> now: https://github.com/pkgxdev/pantry/actions/runs/6577934184

Thanks!

PR #3724, opened 2023-10-19
+fluentci.io

tsirysndr:
[fluentci.io](https://fluentci.io) is a CI/CD tool based on [Dagger](https://dagger.io/) and [Deno](https://deno.com)

### DIFF
--- a/projects/fluentci.io/package.yml
+++ b/projects/fluentci.io/package.yml
@@ -24,6 +24,7 @@ build:
 provides:
   - bin/fluentci
 
-test: |
-    fluentci --version
-    test "$(fluentci --version)" = "fluentci {{ version }}"
+test:
+  - fluentci --version
+  # contains shell escapes; can't use test
+  - fluentci --version | grep "fluentci {{ version }}"
--- a/projects/fluentci.io/package.yml
+++ b/projects/fluentci.io/package.yml
@@ -9,21 +9,21 @@ versions:
 
 dependencies:
   dagger.io: ^0.8.8
+  deno.land: ^1.36
   charm.sh/glow: ^1.5.1
 
 build:
   dependencies:
     deno.land: ^1.37
-  working-directory: ${{prefix}}/bin
   script: |
-    deno compile -A --unstable --import-map="$SRCROOT"/import_map.json --output fluentci "$SRCROOT"/main.ts
+    pkgx --sync
+    deno compile -A --unstable --import-map="$SRCROOT"/import_map.json --output fluentci ./main.ts
     mkdir -p "{{ prefix }}"/bin
     cp fluentci "{{ prefix }}"/bin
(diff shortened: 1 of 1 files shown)


projects/fluentci.io/package.yml
```diff
@@ -0,0 +1,29 @@
...
+build:
+  dependencies:
+    deno.land: ^1.37
+  working-directory: ${{prefix}}/bin
+  script: |
+    deno compile -A --unstable --import-map="$SRCROOT"/import_map.json --output fluentci "$SRCROOT"/main.ts
+    mkdir -p "{{ prefix }}"/bin
+    cp fluentci "{{ prefix }}"/bin
+  # ^^^^ yes, this is a hack, but it's the only way to get the test to run, and it works
+  # the test fail if we don't copy the binary to the prefix, I don't know why
```
jhheider: what about dropping the working-directory and just doing:
```yaml
  script:
    - deno compile -A --unstable --import-map=import_map.json --output fluentci main.ts
    - mkdir -p "{{ prefix }}"/bin
    - install fluentci "{{ prefix }}"/bin
```

projects/fluentci.io/package.yml
```diff
@@ -0,0 +1,29 @@
...
+  url: https://github.com/fluentci-io/fluentci/archive/refs/tags/v{{ version }}.tar.gz
+  strip-components: 1
+
+display-name: fluentci
+
+versions:
+  github: fluentci-io/fluentci
+
+dependencies:
+  dagger.io: ^0.8.8
```
jhheider, line 11: does it have a runtime dep on dagger? but not deno?

projects/fluentci.io/package.yml
tsirysndr: I think I've found the problem, we also need to skip this package [here ](https://github.com/pkgxdev/brewkit/blob/a9c140eec89c8a9db429c5c2ffb2c2b03198bbf3/share/brewkit/fix-elf.ts#L26)

projects/fluentci.io/package.yml
tsirysndr: I just added, Deno was missing
I got this message: `pkgx: cyclic dep: deno.land: fluentci.io`, isn't that a problem?

jhheider: Well, it doesn't need the build dep with the runtime dep. I assume that's why it thinks the dep is cyclic. But a cyclic dep isn't a game stopper, just requires bootstrapping.

projects/fluentci.io/package.yml
tsirysndr: I created a PR for this : https://github.com/pkgxdev/brewkit/pull/189

jhheider:
finally got it!

PR #3726, opened 2023-10-19
apache.org/zookeeper

### LINKED ISSUE #595: +zookeeper (282/548)
Formulae: https://formulae.brew.sh/formula/zookeeper

and-ri:
closes #595

### DIFF
--- a/projects/apache.org/zookeeper/package.yml
+++ b/projects/apache.org/zookeeper/package.yml
@@ -1,6 +1,3 @@
-# distributable:
-#   url: https://archive.apache.org/dist/zookeeper/zookeeper-{{version}}/apache-zookeeper-{{version}}.tar.gz
-#   strip-components: 1
 versions:
   url: https://archive.apache.org/dist/zookeeper/
   match: /zookeeper-\d+\.\d+\.\d+/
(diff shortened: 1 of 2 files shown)


projects/apache.org/zookeeper/package.yml
```diff
@@ -0,0 +1,78 @@
...
+    gnu.org/libtool: '*'
+    maven.apache.org: '*'
+    freedesktop.org/pkg-config: '*'
+    curl.se: '*'
+    linux:
+      gnu.org/gcc: '*'
+  # Malformed input or input contains unmappable characters: ~/pantry/builds/apache.org???zookeeper-3.9.1+linux
+  working-directory: ../zookeeper
+  script:
+    - curl -L "https://archive.apache.org/dist/zookeeper/zookeeper-{{version}}/apache-zookeeper-{{version}}.tar.gz" | tar -xz --strip-components=1
```
jhheider: this seems pretty messy (and could easily mess up changes in brewkit in the future); can we just fix whatever part of their build is complaining? had to do something similar for folly:

https://github.com/pkgxdev/pantry/blob/59c5d75d836d7cd8bf08a13e0f3073fec1c5ae4c/projects/facebook.com/folly/package.yml#L30-L41

Regardless, I'd be more inclined to do something like:
- include the distributable
- cd ..
- if test -e zookeeper; rm -rf zookeeper; done
- mv "apache.org*zookeeper" zookeeper

and-ri:
I'm not sure if version checking is enough for the test.
But I can’t catch the problem of starting zkServer in CI...
Locally the server runs on Linux and Darwin

and-ri:
Perhaps a default zoo.cfg should be created so that the user can start the server right away without additional actions? I'm not sure, that's why I'm asking 😇

jhheider:
that's not a bad idea, especially if they ship with reasonable defaults.

PR #3733, opened 2023-10-20
+mkdocs.org/mkdocs -- Project documentation with Markdown

CrystalMethod:


mxcl:
If mkdocs.org is primarily for mkdocs the tool then we prefer to drop the subfolder `./mkdocs/`

jhheider:
thanks for this, @CrystalMethod !

PR #3741, opened 2023-10-20
+luarocks.org

### LINKED ISSUE #647: +luarocks (376/548)
Formulae: https://formulae.brew.sh/formula/luarocks

isworking:
Adds a recipe for [LuaRocks](https://luarocks.org).
closes #647 

(It's a working version of #2315)
Open for suggestions

### DIFF
--- a/projects/luarocks.org/package.yml
+++ b/projects/luarocks.org/package.yml
@@ -1,0 +1,40 @@
+distributable:
+  url: https://github.com/luarocks/luarocks/archive/refs/tags/v{{version.raw}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: luarocks/luarocks/tags
+  strip: /v/
+
+dependencies:
+  lua.org: '*'
+  info-zip.org/unzip: '*'
+
+runtime:
+  env:
+    LUAROCKS_ROOT: "{{prefix}}"
+    LUA_ROOT: "{{deps.lua.org.prefix}}"
+
+build:
+  dependencies:
+    gnu.org/make: ^4
+    gnu.org/sed: ^4
+
+  script: |
+    ./configure $ARGS
+    make --jobs {{ hw.concurrency }} install
+  env:
+    ARGS:
+      - --prefix="{{prefix}}"
+      - --sysconfdir="{{prefix}}/etc"
+      - --rocks-tree="{{prefix}}"
+      - --force-config
+      - --disable-incdir-check
+
+provides:
+  - bin/luarocks
+  - bin/luarocks-admin
+
+test:
+  script: |
+    test "$(luarocks --version | awk 'NR==1 {print $NF}')" = "{{version.raw}}"
--- a/projects/luarocks.org/package.yml
+++ b/projects/luarocks.org/package.yml
@@ -10,19 +10,20 @@ dependencies:
   lua.org: '*'
   info-zip.org/unzip: '*'
 
-runtime:
-  env:
-    LUA_ROOT: "{{deps.lua.org.prefix}}"
-
 build:
   dependencies:
     gnu.org/make: ^4
     gnu.org/sed: ^4
-
-  script: |
-    ./configure $ARGS
-    make --jobs {{ hw.concurrency }} install
-    sed -i.bak "s_{{prefix}}_\$(cd \$(dirname \$0)/.. \&\& pwd)_g"
+  script:
+    - ./configure $ARGS
+    - make --jobs {{ hw.concurrency }} install
+    - run: |
+        fix-shebangs.ts luarocks-admin luarocks
+        sed -i.bak \
+            -e 's|\[\[{{prefix}}|debug.getinfo(1).source:match("@?(.*/)") .. \[\[..|g' \
+            luarocks-admin luarocks
+        rm luarocks-admin.bak luarocks.bak
+      working-directory: ${{prefix}}/bin
   env:
     ARGS:
       - --prefix="{{prefix}}"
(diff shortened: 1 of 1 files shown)


projects/luarocks.org/package.yml
```diff
@@ -0,0 +1,40 @@
...
+  strip: /v/
+
+dependencies:
+  lua.org: '*'
+  info-zip.org/unzip: '*'
+
+runtime:
+  env:
+    LUAROCKS_ROOT: "{{prefix}}"
+    LUA_ROOT: "{{deps.lua.org.prefix}}"
```
jhheider: this line should almost certainly be on the lua.org recipe.

isworking: So, is this correct? (I'm new to Lua, so Idk much about it)
What change should I do?

jhheider: well, if these aren't required, then omit them. if the second one is required, I'd put it in `projects/lua.org/package.yml`. it all depends on what luarocks needs to function.

isworking: > well, if these aren't required, then omit them. if the second one is required, I'd put it in `projects/lua.org/package.yml`. it all depends on what luarocks needs to function.

I'll omit them and see if it builds.

isworking: It works with both of them omitted.

jhheider:
`/opt/lua.org/v5.4.6/bin/lua: bad interpreter`

looks like _hopefully_ you just need to fix up some shebangs (might be as simple as `fix-shebangs.ts {{prefix}}/bin/*`). if there's more complicated hard-coded paths, you can do things like `sed -i.bak "s_{{prefix}}_\$(cd \$(dirname \$0)/.. \&\& pwd)_g"` if it's shell scripts. there are other tricks in the pantry.

isworking:
> `/opt/lua.org/v5.4.6/bin/lua: bad interpreter`
> 
> looks like _hopefully_ you just need to fix up some shebangs (might be as simple as `fix-shebangs.ts {{prefix}}/bin/*`). if there's more complicated hard-coded paths, you can do things like `sed -i.bak "s_{{prefix}}_\$(cd \$(dirname \$0)/.. \&\& pwd)_g"` if it's shell scripts. there are other tricks in the pantry.

To be honest, I don't know if it's occuring because of path difference or because Lua isn't there. (It's specified as dependency, so it should be). Weirdly, `pkg test` passes.

jhheider:
our pkg test runs in a different environment than the build (with a different PKGX_DIR), to ensure the packages are relocatable. usually an error like that (referring to build paths) means something has become hard-coded in a {{prefix}}/bin file. when it's a shell or other script, there's easy tricks, like above, to fix that.

projects/luarocks.org/package.yml
```diff
@@ -22,6 +22,7 @@ build:
   script: |
     ./configure $ARGS
     make --jobs {{ hw.concurrency }} install
+    sed -i.bak "s_{{prefix}}_\$(cd \$(dirname \$0)/.. \&\& pwd)_g"
```
jhheider: you need to look at the files in `{{prefix}}/bin` to see if any of them need this (this invocation isn't modifying any files). this also wouldn't be a proper fix for a shebang; that should use our script.

jhheider:
```shell
jacob@illithid pantry % rg jacob ~/.pkgx/luarocks.org/v3.9.2/bin 
/Users/jacob/.pkgx/luarocks.org/v3.9.2/bin/luarocks-admin
1:#!/Users/jacob/.pkgx/lua.org/v5.4.6/bin/lua
2:package.loaded["luarocks.core.hardcoded"] = { FORCE_CONFIG = true, SYSCONFDIR = [[/Users/jacob/.pkgx/luarocks.org/v3.9.2/etc/luarocks]] }
3:package.path=[[/Users/jacob/.pkgx/luarocks.org/v3.9.2/share/lua/5.4/?.lua;]] .. package.path
4:local list = package.searchers or package.loaders; table.insert(list, 1, function(name) if name:match("^luarocks%.") then return loadfile([[/Users/jacob/.pkgx/luarocks.org/v3.9.2/share/lua/5.4/]] .. name:gsub([[%.]], [[/]]) .. [[.lua]]) end end)

/Users/jacob/.pkgx/luarocks.org/v3.9.2/bin/luarocks
1:#!/Users/jacob/.pkgx/lua.org/v5.4.6/bin/lua
2:package.loaded["luarocks.core.hardcoded"] = { FORCE_CONFIG = true, SYSCONFDIR = [[/Users/jacob/.pkgx/luarocks.org/v3.9.2/etc/luarocks]] }
3:package.path=[[/Users/jacob/.pkgx/luarocks.org/v3.9.2/share/lua/5.4/?.lua;]] .. package.path
4:local list = package.searchers or package.loaders; table.insert(list, 1, function(name) if name:match("^luarocks%.") then return loadfile([[/Users/jacob/.pkgx/luarocks.org/v3.9.2/share/lua/5.4/]] .. name:gsub([[%.]], [[/]]) .. [[.lua]]) end end)
```

so, the shebangs need to be fied, as well as the path references; unfortunately, that'll have to be done using `lua` syntax. let me see if chatgpt knows it. ;)

jhheider: chatgpt thinks this might do it:
```yaml
  - run: |
      sed -i.bak \
          -e 's|{{prefix}}|\" .. debug.getinfo(1).source:match("@?(.*/)") .. \"|g' \
          luarocks-admin luarocks
          rm luarocks-admin.bak luarocks.bak
    working-directory: ${{prefix}}/bin
```

jhheider: ok, think i got it.

jhheider: yup, just had to learn _very very little_ about lua syntax.

PR #3754, opened 2023-10-22
+julialang.org

### LINKED ISSUE #3697: Julia
I welcome the addition of the [Julia](https://github.com/JuliaLang/julia/) programming language.

isworking:
Adds a recipe for [Julia](https://julialang.org).

closes #3697

### DIFF
--- a/projects/julialang.org/fixture.jl
+++ b/projects/julialang.org/fixture.jl
@@ -1,0 +1,1 @@
+println("Hello, World!")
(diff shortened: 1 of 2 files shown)


jhheider:
that fix-machos error is suspicious. it almost looks like it's being called on a broken symlink. since the constructor for the Fixer class is called like this:

https://github.com/pkgxdev/brewkit/blob/efcd2fb990d3674eea85329e626c17855610ae53/share/brewkit/fix-machos.rb#L231-L245

jhheider:
ah, it looks like /lib/julia/sys.dylib.dSYM/ is probably the issue here. let's see if anything breaks if we just remove that.

jhheider:
passes tests. ship it!

ShalokShalom:
Thanks!

isworking:
> passes tests. ship it!

Thanks for the help!

PR #3765, opened 2023-10-22
swift-server.github.io/swiftly

### LINKED ISSUE #3297: +swiftly
[Swiftly](https://github.com/swift-server/swiftly) is the officially recognized universal Swift installer by the Swift Server Workgroup, which just got its first official [0.1.0](https://github.com/swift-server/swiftly/releases/tag/0.1.0) release. Only compatible with Linux at the moment, but it'll be compatible with macOS eventually, so it feels like an essential building block for tea.

and-ri:
closes #3297


and-ri:
Hmm.. this is really a problem
It works on my Ubuntu, but I didn't take into account that the CI system might not be on the supported list (https://www.swift.org/download/)

jhheider:
yes, that'll prevent viable testing, probably. however, maybe we can dissect their shell script to pull out the binary fetch/unpack and it's "Just Work" since ubuntu is a debian derivation?

and-ri:
I think I can't do without @jhheider  help with this one 😇

and-ri:
We can of course just copy the precompiled binary. But I would like to build it if possible

jhheider:
ugh, requires a newer glibc. that's thorny.

PR #3775, opened 2023-10-23
sdkman.io

### LINKED ISSUE #809: :heavy_plus_sign: sdkman
https://sdkman.io/install

and-ri:
closes #809


jhheider:
it looks like the docker file just runs gradle. is there something that's preventing building?

and-ri:
hmm, I just followed their instructions for homebrew https://github.com/sdkman/homebrew-tap/blob/master/Formula/sdkmancli.rb

jhheider:
dug into this a bit. non-trivial (java), looks like it needs an older openjdk, and jdk's versions are complicated. guess we'll vendor for now.

PR #3776, opened 2023-10-23
+vercel.com/pkg

isworking:
Adds a recipe for [Vercel Pkg](https://github.com/vercel/pkg).

### DIFF
--- a/projects/vercel.com/pkg/package.yml
+++ b/projects/vercel.com/pkg/package.yml
@@ -7,17 +7,18 @@ versions:
 
 dependencies:
   #FIXME I don't know which versions it supports, but Node.js v18 is one of the most well-supported ones.
-  nodejs.org: ^18
+  nodejs.org: "*"
 
 build:
   dependencies:
     npmjs.com: "*"
 
   script: |
-    npm i
     npm install --global --build-from-source --prefix={{prefix}}
 
 test:
+  dependencies:
+    nodejs.org: "*"
   script: |
     test "$(pkg --version)" = "{{version.raw}}"
     pkg fixture.js -t node18-$TARGET -o fixture
--- a/projects/vercel.com/pkg/package.yml
+++ b/projects/vercel.com/pkg/package.yml
@@ -12,7 +12,6 @@ dependencies:
 build:
   dependencies:
     npmjs.com: "*"
-    vercel.com/pkg: "*"
 
   script: |
     npm i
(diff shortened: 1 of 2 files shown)


projects/vercel.com/pkg/package.yml
```diff
@@ -0,0 +1,41 @@
...
+dependencies:
+  #FIXME I don't know which versions it supports, but Node.js v18 is one of the most well-supported ones.
+  nodejs.org: ^18
+
+build:
+  dependencies:
+    npmjs.com: "*"
+
+  script: |
+    npm i
```
jhheider, line 17: is this additional install doing anything?

projects/vercel.com/pkg/package.yml
```diff
@@ -0,0 +1,41 @@
+distributable:
+  url: https://github.com/vercel/pkg/archive/refs/tags/{{version.raw}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: vercel/pkg/tags
+
+dependencies:
+  #FIXME I don't know which versions it supports, but Node.js v18 is one of the most well-supported ones.
+  nodejs.org: ^18
```
jhheider: might be safer to use '*' so it doesn't cause conflicts (unless it's not ready for 21, in which case <21 might be better.)

isworking:
I've no idea why it's failing. (Probably because the binary is `pkg`?)

projects/vercel.com/pkg/package.yml
```diff
@@ -0,0 +1,44 @@
...
+    }
+    darwin/x86-64: {
+      TARGET: "macos-x64"
+    }
+    darwin/aarch64: {
+      TARGET: "macos-arm64"
+    }
+
+provides:
+  - bin/pkg
```
jhheider: we normally like to leave software's commands as default, but this one could make using brewkit confusing. @mxcl you think this might be better as `- bin/vercel-pkg`?

jhheider:
> I've no idea why it's failing. (Probably because the binary is `pkg`?)

i suspect it's the `--install-links` flag. most of our npm packages use:
```yaml
    - npm install .
      --global
      --prefix={{prefix}}
      --install-links
```

isworking:
> > I've no idea why it's failing. (Probably because the binary is `pkg`?)
> 
> i suspect it's the `--install-links` flag. most of our npm packages use:
> 
> ```yaml
>     - npm install .
>       --global
>       --prefix={{prefix}}
>       --install-links
> ```

Let's see

isworking:
Finally!

projects/vercel.com/pkg/package.yml
mxcl: This project is the reason I didn't add `pkg` to brewkit’s provides. There are also others, but this one is common.

So to be clear: this `package.yml` should indeed provide `pkg`

PR #3783, opened 2023-10-23
+vitejs.dev

isworking:
Adds a recipe for [Vite](https://vitejs.dev). Compiled with Deno.

### DIFF
--- a/projects/vitejs.dev/package.yml
+++ b/projects/vitejs.dev/package.yml
@@ -9,12 +9,15 @@ build:
     deno.land: "*"
   script: |
     mkdir -p "{{prefix}}"/bin
-    deno compile npm:vite@{{version.raw}} -o {{prefix}}/bin/vite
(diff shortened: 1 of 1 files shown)


isworking:
@jhheider Any idea why it's not working?

jhheider:
probably because brewkit 0.51.0 hasn't built yet, due to the github api not returning that tag :/

isworking:
> probably because brewkit 0.51.0 hasn't built yet, due to the github api not returning that tag :/

Or could be due to some cache maybe? (Although having older BrewKit cached might not be the case, since all the caches are cleaned)

isworking:
#3786 is the reason

jhheider:
correct. it's not coming back from the github api to the version resolver yet.

jhheider:
that was a lot of casting about. thanks, @Rajdeep-TG !

PR #3832, opened 2023-10-25
Initial packaging of asciinema

ranok:


projects/asciinema.org/asciinema/package.yml
```diff
@@ -0,0 +1,25 @@
...
+
+build:
+  script: python-venv.sh {{prefix}}/bin/asciinema
+
+provides:
+  - bin/asciinema
+
+test:
+  dependencies:
+    python.org: ^3.12
```
jhheider: tests will bring in runtime dependencies; so this shouldn't be needed here.

jhheider (changes requested):
package name should just be `asciinema.org`

thanks for the contribution!

ranok:
> package name should just be `asciinema.org`
> 
> thanks for the contribution!

I am going to package agg, which is by the same author, but converts the asciinema files into animated GIFs, I was thinking I'd put it under the same asciinema.org parent, but it's a different binary (Rust vs python)

projects/asciinema.org/asciinema/package.yml
ranok: Fixed, thanks

jhheider:
> I am going to package agg, which is by the same author, but converts the asciinema files into animated GIFs, I was thinking I'd put it under the same asciinema.org parent, but it's a different binary (Rust vs python)

that should be fine. there's no requirement that a node not be both a namespace and a package. see: `rust-lang.org` and `rust-lang.org/cargo`. i think it's intuitive that the package with the namespace's name should be named that way.

ranok:
> > I am going to package agg, which is by the same author, but converts the asciinema files into animated GIFs, I was thinking I'd put it under the same asciinema.org parent, but it's a different binary (Rust vs python)
> 
> that should be fine. there's no requirement that a node not be both a namespace and a package. see: `rust-lang.org` and `rust-lang.org/cargo`. i think it's intuitive that the package with the namespace's name should be named that way.

Ok, I will put agg as a sub-dir in the parent, and I moved this one up

jhheider (approved):

PR #3861, opened 2023-10-27
bottle on github infra

mxcl:


jhheider:
will require changes to https://github.com/pkgxdev/brewkit/blob/main/actions/bottle/bottle.ts to not use 
`const installation = await cellar.resolve(pkg)` and
`const tarball = useCache().path({ pkg, type: 'bottle', compression })`, minimally.

mxcl:
> `const installation = await cellar.resolve(pkg)`

Not sure what needs changing here. I fixed the other usage which I think is all that is required.

PR #3888, opened 2023-10-29
+wails.io

mycrEEpy:
Adding [wails](https://wails.io/)

Draft because we are currently missing [some dependencies](https://wails.io/docs/gettingstarted/installation#platform-specific-dependencies) (see comments).

### DIFF
--- a/projects/wails.io/package.yml
+++ b/projects/wails.io/package.yml
@@ -1,0 +1,46 @@
+distributable:
+  url: https://github.com/wailsapp/wails/archive/refs/tags/v{{version}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: wailsapp/wails
+  strip: /^v/
+
+dependencies:
+  go.dev: ^1.18
+  npmjs.com: "*"
+  linux:
+    gnu.org/gcc: "*"
+    freedesktop.org/pkg-config: "*"
+    # we are missing dependencies for:
+    #   - libgtk-3
+    #   - libwebkit
+  darwin:
+    # we are missing dependencies for:
+    #   - xcode
+
+build:
+  script: |
+    cd ./v2
+    go build -v -ldflags="$LDFLAGS" -o wails ./cmd/wails
+    mkdir -p "{{ prefix }}"/bin
+    mv wails "{{ prefix }}"/bin
+  env:
+    CGO_ENABLED: 0
+    LDFLAGS:
+      - -extldflags=-static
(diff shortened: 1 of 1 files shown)


isworking:
Can `gtk.org/gtk3` be used? I think it provides `libgtk3`.

mycrEEpy:
> Can `gtk.org/gtk3` be used? I think it provides `libgtk3`.

I'll try, thanks!

EDIT: Sadly `wails doctor` still says it's not installed.

isworking:
> > Can `gtk.org/gtk3` be used? I think it provides `libgtk3`.
> 
> I'll try, thanks!
> 
> EDIT: Sadly `wails doctor` still says it's not installed.

Have you put that outside of `build`? (If you've put it inside `build`, it'll be a build dependency, not a runtime one)

mycrEEpy:
> Have you put that outside of `build`? (If you've put it inside `build`, it'll be a build dependency, not a runtime one)

Yes, tested it as runtime dependency.

isworking:
> > Have you put that outside of `build`? (If you've put it inside `build`, it'll be a build dependency, not a runtime one)
> 
> Yes, tested it as runtime dependency.

I think there could be some version issue. Currently, GTK3 in PKGX has a hard-coded version of 3.24.38. Maybe Wails needs a newer version? Otherwise, there could be some path issue.

mxcl:
the `test` should run `wails doctor` then…?

We can skip linux for now if you want to avoid these problems, but everything passes currently so I can't assist further without building locally, which eventually I will but it's easier to read logs here

mycrEEpy:
We can currently build & package the binary but are unable to provide all runtime dependencies (`wails doctor` checks these dependencies) which are needed to build a project with wails.

I can enable the `doctor` and `build` commands in the `test` script to show logs if favorable.

mxcl:
so yeah it cannot find gtk3 even though it is provided. However they check and consume it is not working with our environment setup. We'll have to dig into that.

libwebkit is outstanding @ArionThinker can you pkg that?

The other platforms (even darwin?!) want `sysctl` which I expect we can pkg, but it's possibly something that is too custom and should be provided by the OS.

and-ri:
> libwebkit is outstanding @ArionThinker can you pkg that?

Yes, sure. Do you mean this https://webkitgtk.org/?

mxcl:
thanks, I believe so yes

PR #3898, opened 2023-10-30
+(opencanaryd)

ranok:
I tried using the deps.python.version.major in the PYTHONPATH, but it did not work correctly, so I left it hard-coded to 3.10.

Also, I'm not sure if there's a comment option, but you need to run this as root in order for it to work (it opens low ports), and the script itself will use sudo if not root which will wipe the env and break the PYTHONPATH.

### DIFF
--- a/projects/github.com/thinkst/opencanary/README.md
+++ b/projects/github.com/thinkst/opencanary/README.md
@@ -1,0 +1,27 @@
+# Running OpenCanary via pkgx
+
+## Configuring OpenCanary
+
+When OpenCanary starts it looks for config files in the following locations and will stop when the first configuration is found:
+
+1. `./opencanary.conf` (i.e. the directory where OpenCanary is installed)
+2. `~/.opencanary.conf` (i.e. the home directory of the user, usually this will be `root` so `/root/.opencanary.conf`)
+3. `/etc/opencanaryd/opencanary.conf`
+
+To create an initial configuration, run as `root` (you may be prompted for a `sudo` password):
+```
+$ pkgx opencanaryd --copyconfig
+[*] A sample config file is ready /etc/opencanaryd/opencanary.conf
+
+[*] Edit your configuration, then launch with "pkgx opencanaryd --start"
+```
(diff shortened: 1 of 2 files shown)


jhheider:
> deps.python.version.major

should be `{{ deps.python.org.version.major }}`.

> Also, I'm not sure if there's a comment option

you can add a README.md.

ranok:
The issue I was running into was that `{{deps.python.org.version.major}}` is just 3, whereas the Python modules are in the `{{deps.python.org.version.marketing}}`/site-packages directory, which is the major and minor version (e.g., 3.10).

This PR now contains that fix, as well as a README.md to clarify the usage instructions.

jhheider:
> The issue I was running into was that `{{deps.python.org.version.major}}` is just 3, whereas the Python modules are in the `{{deps.python.org.version.marketing}}`/site-packages directory, which is the major and minor version (e.g., 3.10).
> 
> This PR now contains that fix, as well as a README.md to clarify the usage instructions.

the trick we've using in the past looks like this:
```yaml
runtime:
  env:
    PYTHONPATH: {{prefix}}/lib/python{{deps.python.org.version.major}}/site-packages:$PYTHONPATH

build:
...
  - run: ln -s python{{deps.python.org.version.marketing}} python{{deps.python.org.version.major}
    working-directory: ${{prefix}}/lib/
```

that way, it's not bound forever to a specific minor.

ranok:
Ok, I made that change, seems to work just fine, and yeah, it's nice to be able to "float" around minor versions.

jhheider:
it seems like it doesn't like ~3.7. you might want either `^3.7` or `>=3.7<3.12` (if it's not 3.12 ready yet).

ranok:
Not sure what the issue is with aarch64 on py3 <=3.8, but it seems to work fine on 3.9+. I'll submit an upstream PR to update their documentation to say that 3.9 is needed on aarch64

jhheider:
thanks, @ranok !

PR #3951, opened 2023-11-03
+garden -- Develop and test complex systems faster.

kevinmichaelchen:
> Automation for Kubernetes development and testing. Spin up production-like environments for development, testing, and CI on demand. Use the same configuration and workflows at every step of the process. Speed up your builds and test runs via shared result caching

https://github.com/garden-io/garden
https://garden.io/

### DIFF
--- a/projects/garden.io/package.yml
+++ b/projects/garden.io/package.yml
@@ -15,6 +15,10 @@ build:
   script: |
     npm install
     npm run build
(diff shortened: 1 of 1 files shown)


kevinmichaelchen:
The CLI is being published to NPM here: https://github.com/garden-io/garden/pull/5087

PR #3952, opened 2023-11-03
Switch tart.run to vendored binary with non-OSS license so that we can get --net-bridged

magnusviri:
--net-bridged requires an entitlement and the binary must be signed by the vendor. Your signing certificate wont work. --net-bridged allows VM's to switch networking, like on a laptop that has both wifi and ethernet, --net-bridged is a must.


projects/tart.run/package.yml
```diff
@@ -1,34 +1,34 @@
...
-    - codesign -f -s - --entitlement Resources/tart.entitlements .build/release/tart
-    - run: cp "$SRCROOT"/.build/release/tart .
-      working-directory: ${{prefix}}/bin
-  env:
-    SOFTNET: https://github.com/cirruslabs/softnet/archive/refs/tags/0.6.2.tar.gz
-
+    curl.se: '*'
+  working-directory: ${{prefix}}
+  script: |
+    curl -Lfo tart.tar.gz "https://github.com/cirruslabs/tart/releases/download/{{version}}/tart.tar.gz"
```
jhheider: why not just make this the distributable link for simplicity? then you can skip this step and the next two.

jhheider (changes requested):

projects/tart.run/package.yml
magnusviri: I played with that but I don't think I understood how it worked at first. I'll play with it again to see if it will work. Also, I copied the pattern in [bun](https://github.com/pkgxdev/pantry/blob/main/projects/bun.sh/package.yml), but that was a zip. Maybe that's why it was using curl.

jhheider: i think it's because bun has a variable ($PLATFOM) which is computed, and not available at download time.

magnusviri: This is great. Thanks for accepting the update!

PR #3963, opened 2023-11-05
Qemu (make it work on linux)

tannevaled:


projects/qemu.org/package.yml
```diff
@@ -57,17 +57,53 @@ dependencies:
...
+    #
+    python.org: '<3.10'
     ninja-build.org: '*'
     freedesktop.org/pkg-config: '*'
+
   script:
+    - if: linux
+      run: |
+        sudo apt -o DPkg::Lock::Timeout=60 update
+        sudo apt -o DPkg::Lock::Timeout=60 -y install build-essential
```
jhheider: we definitely _shouldn't_ do this. this is likely to link in OS libraries, which might not be available on out systems. what was this intended to solve?

tannevaled: i know but with all the tests i did (version max of python, need of bison/flex), compiling in the docker pkgxdev/pkgx container needed that to compile. better using things from pkgx but i think we are mising some dependencies.

jhheider: right, but this permanently alters the runners, and will have extremely far-reaching impacts on numerous packages. i need to go clean this up.

jhheider: here's the list of things it added:
```
binutils binutils-aarch64-linux-gnu binutils-common cpp cpp-8 dpkg-dev g++ g++-8 gcc gcc-8 libasan5 libatomic1 libbinutils libcc1-0 libdpkg-perl libgcc-8-dev
  libgdbm-compat4 libgdbm6 libgomp1 libisl19 libitm1 liblsan0 libmpc3 libmpfr6 libperl5.28 libstdc++-8-dev libtsan0 libubsan1 make patch perl perl-modules-5.28
```

tannevaled: > right, but this permanently alters the runners, and will have extremely far-reaching impacts on numerous packages. i need to go clean this up.

sorry, i thought the builds were isolated.

jhheider: We clean up all the directories used by the builds, but self-hosted runners aren't destroyed and rebuilt every time, sadly. A short coming of needing aarch64 runners.

tannevaled: > here's the list of things it added:
> 
> ```
> binutils binutils-aarch64-linux-gnu binutils-common cpp cpp-8 dpkg-dev g++ g++-8 gcc gcc-8 libasan5 libatomic1 libbinutils libcc1-0 libdpkg-perl libgcc-8-dev
>   libgdbm-compat4 libgdbm6 libgomp1 libisl19 libitm1 liblsan0 libmpc3 libmpfr6 libperl5.28 libstdc++-8-dev libtsan0 libubsan1 make patch perl perl-modules-5.28
> ```

well, it seems finally that build-essential is not needed.
it now builds without it in the pkgxdev/pkgx container.

PR #3977, opened 2023-11-06
qemu on linux opens these up

jhheider:
these should all build on linux now, with qemu available


jhheider:
lol. they have to build one-by-one. i'll do podman first.

PR #3996, opened 2023-11-08
gpg fix

### LINKED ISSUE #3995: Generating GPG Key with pkgx
Has anyone succesfully managed to generate a GPG key using pkgx?

Running `pkgx gpg --full-gen-key` leads to an error because gpg-agent is not found, although technically pkgx does mark it as a bin:

```bash
gpg: error running '/opt/gnupg.org/v2.4.3/bin/gpg-agent': probably not installed
gpg: failed to start gpg-agent '/opt/gnupg.org/v2.4.3/bin/gpg-agent': Configuration error
gpg: can't connect to the gpg-agent: Configuration error
gpg: agent_genkey failed: No agent running
Key generation failed: No agent running
```

and-ri:
closes #3995


jhheider:
it looks like it still can't find `gpg-agent` :(

jhheider:
probably need to dig into https://github.com/gpg/gnupg/blob/64f5f7b74e428b15205f6e8ae14dec84663c3076/common/asshelp.c#L382 to see how to make it try the relative `./` before the baked in path.

and-ri:
or we need to somehow write the absolute path to gpg-agent in gpg.conf already after installation

jhheider:
> or we need to somehow write the absolute path to gpg-agent in gpg.conf already after installation

That could be done with a shim. Look at what python-venv-stubber.sh in brewkit does. It rewrites the launcher file on every run, if needed.

jhheider:
Did you get it? It looks like you got it!

PR #4040, opened 2023-11-12
rockdaboot.github.io/libpsl

and-ri:
libpsl => libsoup => webkitgtk

### DIFF
--- a/projects/rockdaboot.github.io/libpsl/package.yml
+++ b/projects/rockdaboot.github.io/libpsl/package.yml
@@ -1,0 +1,36 @@
+distributable:
+  url: https://github.com/rockdaboot/libpsl/releases/download/{{version}}/libpsl-{{version}}.tar.gz
+  strip-components: 1
+versions:
+  github: rockdaboot/libpsl
+dependencies:
+  unicode.org: '*'
+build:
+  dependencies:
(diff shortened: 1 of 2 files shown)


projects/rockdaboot.github.io/libpsl/package.yml
```diff
@@ -0,0 +1,36 @@
...
+    python.org: ~3.11
+    linux:
+      gnu.org/gcc: '*'
+  script:
+    - meson setup build $MESON_ARGS
+    - meson compile -C build
+    - meson install -C build
+  env:
+    linux:
+      LDFLAGS: -fPIC
```
mxcl, line 22: stuff like this should also have a comment explaining why pls

mxcl (approved):

projects/rockdaboot.github.io/libpsl/package.yml
and-ri: > stuff like this should also have a comment explaining why pls

Oh okay 👌🏻 
In this case, this solves the error:
`undefined reference to 'main'`

and-ri: I will comment this from now

PR #4041, opened 2023-11-12
musepack.net/libcuefile

and-ri:
libcuefile => musepack => gstreamer => webkitgtk

### DIFF
--- a/projects/musepack.net/libcuefile/package.yml
+++ b/projects/musepack.net/libcuefile/package.yml
@@ -1,0 +1,36 @@
+distributable:
+  url: https://files.musepack.net/source/libcuefile_r{{version.major}}.tar.gz
+  strip-components: 1
+versions:
+  url: https://www.musepack.net/index.php?pg=src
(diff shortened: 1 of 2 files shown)


projects/musepack.net/libcuefile/package.yml
```diff
@@ -0,0 +1,38 @@
...
+  url: https://www.musepack.net/index.php?pg=src
+  match: /libcuefile_r\d+\.tar\.gz/
+  strip:
+    - /^libcuefile_r/
+    - /\.tar\.gz/
+build:
+  dependencies:
+    cmake.org: '*'
+    linux:
+      gnu.org/gcc: '*'
```
mxcl, line 14: I think we should add a comment explaining why we are building with `gcc` from now on pls.

and-ri: > I think we should add a comment explaining why we are building with `gcc` from now on pls.

Don't we already need to add a compiler on Linux? Is it already there automatically? Like `make`?

PR #4044, opened 2023-11-12
cisco.com/libsrtp

and-ri:
libsrtp => gstreamer => webkitgtk

### DIFF
--- a/projects/cisco.com/libsrtp/package.yml
+++ b/projects/cisco.com/libsrtp/package.yml
@@ -1,0 +1,33 @@
+distributable:
+  url: https://github.com/cisco/libsrtp/archive/v{{version}}.tar.gz
+  strip-components: 1
+versions:
+  github: cisco/libsrtp
+dependencies:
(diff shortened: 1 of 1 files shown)


projects/cisco.com/libsrtp/package.yml
```diff
@@ -0,0 +1,33 @@
...
+    freedesktop.org/pkg-config: '*'
+    linux:
+      gnu.org/gcc: '*'
+  script:
+    - ./configure $CONFIGURE_ARGS
+    - make --jobs {{ hw.concurrency }} test
+    - make --jobs {{ hw.concurrency }} shared_library
+    - make --jobs {{ hw.concurrency }} install
+    - mkdir -p {{prefix}}/libexec
+      # for test
```
mxcl, line 19: we prefer not to increase pkg size for tests.

However there is no other easy way to get fixtures into tests, so we’ll allow for now. But yeah.

PR #4049, opened 2023-11-12
unicode * is toxic

mxcl:


mxcl:
failure is because LDFLAGS hasn't got our rpath in it. this is odd because we forcibly insert it with our cc shims

jhheider:
explicit gcc means no shims, right?

jhheider:
ohhh, right. rpath. grr.

jhheider:
good error: `building for macOS-x86_64 but attempting to link with file built for macOS-x86_64`

projects/postgresql.org/package.yml
```diff
@@ -40,8 +40,18 @@ build:
...
       - --with-libxslt
+    LDFLAGS:
+      - -Wl,-rpath,{{pkgx.prefix}}
     darwin:
-      LDFLAGS: -headerpad_max_install_names
+      LDFLAGS:
+        - -headerpad_max_install_names
+    linux/x86-64:
+      CFLAGS:
+        - -fPIC
```
mxcl: we add these in brewkit

projects/postgresql.org/package.yml
```diff
@@ -40,8 +40,18 @@ build:
       - --with-lz4
       - --with-libxml
       - --with-libxslt
+    LDFLAGS:
+      - -Wl,-rpath,{{pkgx.prefix}}
```
mxcl: added to brewkit 0.55.1

mxcl:
@jhheider hmm not good https://github.com/pkgxdev/pantry/actions/runs/6877165192/job/18704274799?pr=4049#step:7:2121

mxcl:
nice tricks. No clue how you figured that out!

jhheider:
it's always some set of the same tricks. i think tools get confused calling clang as gcc. though no one should.

Bhagirath65:
nice

Yakin15437 (commented):
Good

PR #4068, opened 2023-11-14
+kaspa

jhheider:
kaspad
kaspa-miner

### DIFF
--- a/projects/crates.io/kaspa-miner/package.yml
+++ b/projects/crates.io/kaspa-miner/package.yml
@@ -1,0 +1,22 @@
+distributable:
+  url: https://github.com/elichai/kaspa-miner/archive/refs/tags/v{{ version }}.tar.gz
+  strip-components: 1
+
+provides:
+  - bin/kaspa-miner
+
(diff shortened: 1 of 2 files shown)


mxcl:
Seems like a cool pkg №1000

jhheider:
on the grow.

mxcl:
Should the name be https://kaspa.org ? Edit: no

jhheider:
> Should the name be https://kaspa.org ? Edit: no

that'd be nice.

mxcl:
> that'd be nice.

I went to kaspa.org and this is just one of many clients, it isn't their official one. So this name seems correct.

davdroman:
Just a heads-up about the imminent replacement of kaspad with a Rust rewritten successor https://github.com/kaspanet/rusty-kaspa which enables 10bps (vs current 1bps, still impressive nonetheless). Might be worth getting it in as well.

jhheider:
Should be simple. Rust and Go are two of the easiest build systems.

PR #4098, opened 2023-11-17
add `restic`

MarkusZoppelt:
Adds pkg for [restic](https://restic.net) by @MichaelEischer

### DIFF
--- a/projects/restic.net/restic/package.yml
+++ b/projects/restic.net/restic/package.yml
@@ -1,0 +1,22 @@
+distributable:
+  url: https://github.com/restic/restic/releases/download/v{{ version }}/restic-{{version}}.tar.gz
+  strip-components: 1
+
+display-name: restic
+
+versions:
+  github: restic/restic
+
+provides:
+  - bin/restic
+
+build:
+  script: |
+    go run build.go
+    mkdir -p "{{ prefix }}"/bin
+    mv restic "{{ prefix }}"/bin
+  dependencies:
+    go.dev: ^1.20
+
+test:
+  restic version


mxcl:
pkg has a homepage so we tend to prefer to use that: [restic.net](https://restic.net/)

You mind renaming? Thanks

projects/github.com/restic/restic/package.yml
```diff
@@ -0,0 +1,22 @@
...
+build:
+  script: |
+    go run build.go
+    mkdir -p "{{ prefix }}"/bin
+    mv restic "{{ prefix }}"/bin
+  dependencies:
+    go.dev: ^1.20
+
+test:
+  restic version
```
MichaelEischer: Why not run `go test ./...`?

projects/github.com/restic/restic/package.yml
```diff
@@ -0,0 +1,22 @@
...
+provides:
+  - bin/restic
+
+build:
+  script: |
+    go run build.go
+    mkdir -p "{{ prefix }}"/bin
+    mv restic "{{ prefix }}"/bin
+  dependencies:
+    go.dev: ^1.20
```
MichaelEischer: The latest restic release is already built using Go 1.21.

jhheider: generally, this should be the _oldest_ release that will correctly build the binary (since ^1.x == >=1.x && <2)

projects/github.com/restic/restic/package.yml
jhheider: the source isn't available in the `test` node. if you wanted to run source tests, you'd do that in the `build` node. the `test` node is for ensuring the resulting binary(ies) work properly.

projects/github.com/restic/restic/package.yml
MichaelEischer: Ah, ok, then 1.20 is fine.

MarkusZoppelt:
> pkg has a homepage so we tend to prefer to use that: [restic.net](https://restic.net/)
> 
> You mind renaming? Thanks

Done!

mxcl:
https://pkgx.dev/pkgs/restic.net/restic/

projects/github.com/restic/restic/package.yml
mxcl: note that I really should make the sources and build dirs available since this is a common thing we want to do. Only reason we do not is due to github actions needing these dirs to be artefacts so that we can test on different machines.

And well, that means everything would need to be rebuilt since (for linux builds at least) we use different operating systems for the tests

We could perhaps have a “test step in the build step” sort of thing. lol

PR #4114, opened 2023-11-18
unicode * is toxic

mxcl:


mxcl:
possibly just because of the matrix, dunno. What you think @jhheider https://github.com/pkgxdev/pantry/actions/runs/6914534410/job/18812647035?pr=4114#step:8:137

jhheider:
I think I recall postgres 16 needs v73. so, we're in a deep crevasse here.

mxcl:
> I think I recall postgres 16 needs v73. so, we're in a deep crevasse here.

k well, maybe this is the first one we can't get away without a versioning variation matrix

PR #4121, opened 2023-11-18
Improve git subcommand detection

fredrikaverpil:
Related: https://github.com/pkgxdev/pkgx/issues/784

### Why this change?

I noticed in https://github.com/pkgxdev/pkgx/issues/784#issuecomment-1817484656 that git-scm.org's `git-shim` doesn't handle configuration options properly. This cause delays when git commands are executed with the `-c` parameter (e.g. `-c submodule.recurse=false`).

### What was changed?

Given that `git --help` (v2.42.1) says the following:

```
usage: git [-v | --version] [-h | --help] [-C <path>] [-c <name>=<value>]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | -P | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           [--config-env=<name>=<envvar>] <command> [<args>]
```

I expanded a bit on the cases to handle this.

This shell function which is meant to detect subcommands, is brittle, as bugs can be introduced as git introduces new arguments.

An option is to instead look for the known subcommands:

```
start a working area (see also: git help tutorial)
   clone     Clone a repository into a new directory
   init      Create an empty Git repository or reinitialize an existing one

work on the current change (see also: git help everyday)
   add       Add file contents to the index
   mv        Move or rename a file, a directory, or a symlink
   restore   Restore working tree files
   rm        Remove files from the working tree and from the index

examine the history and state (see also: git help revisions)
   bisect    Use binary search to find the commit that introduced a bug
   diff      Show changes between commits, commit and working tree, etc
   grep      Print lines matching a pattern
   log       Show commit logs
   show      Show various types of objects
   status    Show the working tree status

grow, mark and tweak your common history
   branch    List, create, or delete branches
   commit    Record changes to the repository
   merge     Join two or more development histories together
   rebase    Reapply commits on top of another base tip
   reset     Reset current HEAD to the specified state
   switch    Switch branches
   tag       Create, list, delete or verify a tag object signed with GPG

collaborate (see also: git help workflows)
   fetch     Download objects and refs from another repository
   pull      Fetch from and integrate with another repository or a local branch
   push      Update remote refs along with associated objects
```

But this is as brittle, I guess, since git likely will change this set of subcommands at some point.

### DIFF
--- a/projects/git-scm.org/git-shim
+++ b/projects/git-scm.org/git-shim
@@ -21,9 +21,6 @@ find_git_subcommand() {
         skip_next=1  # The next argument is a value for these options.
         continue
         ;;
-      --exec-path=* | -c=* | --git-dir=* | --work-tree=* | --namespace=* | --super-prefix=* | --config-env=*)
-        continue  # Skip options that include their values with '='.
-        ;;
       -*)
         continue  # Skip other options that start with '-'.
         ;;
--- a/projects/git-scm.org/git-shim
+++ b/projects/git-scm.org/git-shim
@@ -2,26 +2,43 @@
 
 set -e
 
-find_git_command() {
+# This function searches through the provided arguments
+# to find the first non-option argument, which is assumed
+# to be the git subcommand.
+find_git_subcommand() {
+  skip_next=0  # Flag to indicate if the next argument should be skipped.
+
   for arg in "$@"; do
-    if test "$SKIP" = 1; then
-      SKIP=0
+    # If the skip_next flag is set, skip this argument.
+    if [ "$skip_next" -eq 1 ]; then
+      skip_next=0
       continue
     fi
+
+    # Check for options that require skipping the next argument.
     case $arg in
-    -C) SKIP=1; continue;;
-    -*);;
-    *)
-      echo "$arg"
-      return  # found the command!
+      -C | --git-dir | --work-tree | --namespace | --super-prefix | --config-env | -c)
+        skip_next=1  # The next argument is a value for these options.
+        continue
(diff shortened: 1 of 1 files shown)


projects/git-scm.org/git-shim
```diff
@@ -2,26 +2,43 @@
...
-      return  # found the command!
+      -C | --git-dir | --work-tree | --namespace | --super-prefix | --config-env | -c)
+        skip_next=1  # The next argument is a value for these options.
+        continue
+        ;;
+      --exec-path=* | -c=* | --git-dir=* | --work-tree=* | --namespace=* | --super-prefix=* | --config-env=*)
+        continue  # Skip options that include their values with '='.
+        ;;
+      -*)
+        continue  # Skip other options that start with '-'.
```
jhheider: this first case will be handled by the second (i.e. doesn't need an additional case).

fredrikaverpil: Ah, yes. Fixed in fb16d39!

jhheider:
great addition!

fredrikaverpil:
@jhheider  do you think the GHA workflow needs to be rerun once more, so to get this published?

jhheider:
it does and i have. it's because of this:
```yaml
      - uses: technote-space/get-diff-action@v6
        id: get-diff
        with:
          PATTERNS: projects/**/package.yml
```
should probably change that `package.yml` to a `*`.

PR #4126, opened 2023-11-19
+mupdf

vcvvvc:
mupdf.com
https://github.com/ArtifexSoftware/mupdf
https://mupdf.readthedocs.io/en/latest/quick-start-guide.html

`pkg build` has an error, need help


projects/mupdf.com/package.yml
```diff
@@ -0,0 +1,42 @@
...
+    x.org/xinerama: '*'
+    x.org/xrandr: '*'
+    mesa3d.org: '*'
+    gnome.org/glib: '*'
+    x.org/xtrans: '*'
+
+build:
+  dependencies:
+    git-scm.org: '*'
+    gnu.org/binutils: '*'
```
MarkusZoppelt: You'll probably also need `gnu.org/gcc` for `g++`

```suggestion
    gnu.org/binutils: '*'
    gnu.org/gcc: '*'
```

vcvvvc:
CXX build/release/source/fitz/tessocr.o

vcvvvc:
muraster: error while loading shared libraries: libcrypto.so.1.1: cannot open shared object file: No such file or directory

vcvvvc:
Hello, if this method fails the test, we can consider switching to this way.

I passed the local test, but I haven’t used pkgx to test yet.

https://mupdf.readthedocs.io/en/latest/using-mupdf.html 

```
gcc -I/usr/local/include -o example \
	test.c \
	/usr/local/lib/libmupdf.a \
	/usr/local/lib/libmupdf-third.a \
	-lm
```

jhheider:
looks like it's good! thanks, @6923403 !

vcvvvc:
> looks like it's good! thanks, @6923403 !

Cool, bro

PR #4130, opened 2023-11-19
sass-lang.com/sassc

and-ri:
sassc => gtk4 => gstreamer => webkitgtk

### DIFF
--- a/projects/sass-lang.com/sassc/package.yml
+++ b/projects/sass-lang.com/sassc/package.yml
@@ -1,0 +1,26 @@
+distributable:
+  url: https://github.com/sass/sassc/archive/refs/tags/{{version}}.tar.gz
+  strip-components: 1
+versions:
+  github: sass/sassc
+dependencies:
+  sass-lang.com/libsass: '*'
+build:
+  dependencies:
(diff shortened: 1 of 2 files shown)


projects/sass-lang.com/sassc/package.yml
```diff
@@ -0,0 +1,26 @@
+distributable:
+  url: https://github.com/sass/sassc/archive/refs/tags/{{version}}.tar.gz
+  strip-components: 1
+versions:
+  github: sass/sassc
+dependencies:
+  sass-lang.com/libsass: '*'
```
mxcl: Sorry. We do not permit `'*'` dependencies any more.

mxcl (changes requested):

projects/sass-lang.com/sassc/package.yml
and-ri: Okay, got it 👌🏻

and-ri: Am I correct in understanding that we should not use `'*'` in runtime dependencies? In the build section we can use this as before

PR #4131, opened 2023-11-19
github.com/antirez/smallchat

and-ri:
https://ossinsight.io/analyze/antirez/smallchat


mxcl:
hmm this seems to just be a programming example/tutorial. I don’t think it belongs in `pkgx`. Unless I'm mistaken?

and-ri:
> hmm this seems to just be a programming example/tutorial. I don’t think it belongs in `pkgx`. Unless I'm mistaken?

Hmm, yes you're absolutely right

PR #4162, opened 2023-11-21
npx; Fixes #4160

### LINKED ISSUE #4160: `npx prettier --version` reports npx version and not prettier version
`npx prettier -- --version` works.

Apparently installing node via eg. brew the above works.

mxcl:


KubaJastrz:
FYI: this will fail if someone is running flags on npx, like: `npx --help`

mxcl:
@KubaJastrz good catch!

PR #4165, opened 2023-11-21
fix whisper.cpp metal; remove helpers

mxcl:
Trying to standardize on huggingface-cli downloads the models and ScriptHub scripts or endpoints make use of that.


mxcl:
My thinking here is that the pkgs are raw, but we add endpoints and or ScriptHub scripts that facilitate the model aquisition, thus delegating that to other utilities. Here we use hugging face cli which I think should be the standard.

Then we ensure pkg pages link to scripts and endpoints in oss.app and the web.

PR #4175, opened 2023-11-22
fix tilt.dev

### LINKED ISSUE #4171: tilt.dev is built without web assets and fails
```shell
pkgx tilt up
Tilt started on http://localhost:10350/
v0.33.6, built 
Error: running in prod mode, but JS/CSS files are not available.
Please report this: https://github.com/tilt-dev/tilt/issues
```

mycrEEpy:
Closes #4171

### DIFF
--- a/projects/tilt.dev/package.yml
+++ b/projects/tilt.dev/package.yml
@@ -10,11 +10,9 @@ provides:
 
 build:
   dependencies:
-    go.dev: ^1.19
+    go.dev: ~1.19
     nodejs.org: ^20
     classic.yarnpkg.com: "*"
-    linux:
-      gnu.org/gcc: '*'
   script:
     - make build-js
     - go mod download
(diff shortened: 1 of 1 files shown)


mycrEEpy:
Anyone any idea why ld is not available on linux/aarch64?

jhheider:
> Anyone any idea why ld is not available on linux/aarch64?

if you don't add a toolchain, brewkit should set one up: https://github.com/pkgxdev/brewkit/blob/82724a21c9a8639fa5a4e231b98c9cfd44c96f8e/libexec/stage.ts#L61-L96

but i think it's linking ld to the wrong place (lld rather than ld.lld).

jhheider:
might just need the `gold` linker. adding to [binutils](https://github.com/pkgxdev/pantry/pull/4187).

jhheider:
sweet. only had to compile a different linker.

PR #4185, opened 2023-11-23
+dblab — The database client every command line junkie deserves.

kevinmichaelchen:
https://github.com/danvergara/dblab

> The database client every command line junkie deserves.


jhheider:
probably ought to name it: dblab.danvergara.com

Issue #4190, opened 2023-11-23
Hardcoded GNUPGHOME for gnupg.org


Cytrian:
I have this issue:
In `projects/gnupg.org/package.yml` GNUPGHOME is hardcoded to `GNUPGHOME: '{{prefix}}/etc/gnupg'`
(which in my case is `~/.pkgx/gnupg.org/v2.4.3/etc`)

So it is not possible to use a different GNUPGHOME directory, e.g. `~/.gnupg`, which is the default directory.

jhheider:
Yes, this was a concern. GNUPGHOME was the way we fixed `gpg` looking at the build path for its `gpg-agent` instead of the run path. We are still thinking we might be able to solve it in the source (as we do with `openssl`).

To work around, try setting GNUPGHOME to `${GNUPGHOME:-"{{prefix}}/etc/gnupg"}` in your `~/Library/Application\ Support/pkgx/pantry/projects/gnupg.org/package.yml`. Assuming that works for you, that's likely the real fix.

Cytrian:
I don't get it. You mean like this?
```
runtime:
  env:
    GNUPGHOME: "${GNUPGHOME:-'{{prefix}}/etc/gnupg'}"
```
That only gives errors
```
gpg: keyblock resource '/Users/a-user/${GNUPGHOME:-'/Users/a-user/.pkgx/gnupg.org/v2.4.3/etc/gnupg'}/pubring.kbx': No such file or directory
```
That assignment should go into the script `.local/bin/gpg` but I don't see a way to put code there through pkgx.

jhheider:
Well, putting it in the .local/bin/gpg shim won't help people who don't use `pkgx install gpg`. Hm. @mxcl any way to get a conditional run-time env at present? Or do we need to add it?

mxcl:
Well, like I've said previously, it is not acceptable for us to override env vars for our own purposes that a user may need so the GPG package yml needs fixing whatever.

I guess I'll look at it since nobody else is willing to go the extra mile.

mxcl:
So looking through the sources for gpg I see that there is support for a hidden conf variable that exposes an env var to set this dynamically, so setting that and seeing how we go.

However in the process I find that building gpg depends on gpg and I cannot for the life of me figure out what is causing this circular dep. This is problematic and seems to prevent rebuilds of gpg (ie. the install step doesn't copy our newly built product since it finds it is already installed).

mxcl:
[...]

PR #4209, opened 2023-11-25
make gpg relocatable without hacks

mxcl:
Refs https://github.com/pkgxdev/pantry/issues/4190


mxcl:
k all tests pass, however I dunno if this actually tests what is required to prove it works as before /cc @jhheider

jhheider:
Interesting. Yes, all the rigamarole was designed to let `gpg` auto-start its `gpg-agent`, which it was trying to do with build paths. So, the test should prove it out.

PR #4221, opened 2023-11-27
fix turso build

johannesg:
I'm new at using pkgx so apologies if I'm doing something wrong.

The turso cli does not work currently, it gives the following error when running any command against [turso.tech](https://turso.tech):

```
Error: failed to get database listing: Get "https://api.turso.tech/v1/databases": net/http: invalid header field value for "Tursocliversion"
```

I'm not a fluent gopher either but I managed to get it working by adding a `go generate...`. I copied it from [This file](https://github.com/tursodatabase/turso-cli/blob/main/.goreleaser.yaml)

### DIFF
--- a/projects/turso.tech/package.yml
+++ b/projects/turso.tech/package.yml
@@ -13,6 +13,7 @@ build:
   script:
     - echo "{{version}}" > internal/cmd/version.txt
     - go mod download
+    - go generate --tags=prod ./...
     - go build $ARGS -ldflags="$LDFLAGS" ./cmd/turso
   dependencies:
     go.dev: ^1.20
--- a/projects/turso.tech/package.yml
+++ b/projects/turso.tech/package.yml
@@ -33,6 +33,8 @@ build:
       # or segmentation fault
       # fix found here https://github.com/docker-library/golang/issues/402#issuecomment-982204575
       LDFLAGS:
-      - -buildmode=pie
+        - -buildmode=pie
 
-test: turso --version | grep {{version}}
+test:
+  - turso --version
+  - 'test "$(turso db locations 2>&1 || true)" = "Error: user not logged in, please login with turso auth login"'
--- a/projects/turso.tech/package.yml
+++ b/projects/turso.tech/package.yml
@@ -35,6 +35,4 @@ build:
       LDFLAGS:
         - -buildmode=pie
 
-test:
-  - turso --version
-  - 'test "$(turso db locations 2>&1 || true)" = "Error: user not logged in, please login with turso auth login"'
+test: 'test "$(turso db locations 2>&1 || true)" = "Error: user not logged in, please login with turso auth login"'


jhheider:
looks good! odd that it messes up the version flag, but a better test is a better test.

johannesg:
Thank you for looking at this! Unfortunately, it seems to be even more broken than before :-/. Every command seems to end with an exception and the `--version` does not work at all.

I suppose I need to dig deeper (unfortunately I don't know much about it, I'm just a turso user)

```panic: runtime error: slice bounds out of range [1:0]

goroutine 1 [running]:
github.com/tursodatabase/turso-cli/internal/turso.(*Client).newRequest(0xc00026c320, {0x5556869d52de, 0x3}, {0x5556869dfaa9, 0x10}, {0x0, 0x0})
        github.com/tursodatabase/turso-cli/internal/turso/turso.go:84 +0x609
github.com/tursodatabase/turso-cli/internal/turso.(*Client).do(0x0?, {0x5556869d52de?, 0x5556866ff1c0?}, {0x5556869dfaa9?, 0xc559480300000000?}, {0x0?, 0x0?})
        github.com/tursodatabase/turso-cli/internal/turso/turso.go:92 +0x37
github.com/tursodatabase/turso-cli/internal/turso.(*Client).Get(...)
        github.com/tursodatabase/turso-cli/internal/turso/turso.go:136
github.com/tursodatabase/turso-cli/internal/cmd.fetchLatestVersion()
        github.com/tursodatabase/turso-cli/internal/cmd/utils.go:314 +0x6d
github.com/tursodatabase/turso-cli/internal/cmd.init.27.func1(0xc000046b00?, {0x5556874aafc0?, 0x4?, 0x5556869d5816?})
        github.com/tursodatabase/turso-cli/internal/cmd/root.go:44 +0xed
github.com/spf13/cobra.(*Command).execute(0x555687467da0, {0x5556874aafc0, 0x0, 0x0})
        github.com/spf13/cobra@v1.6.1/command.go:936 +0xa22
github.com/spf13/cobra.(*Command).ExecuteC(0x55568746ba00)
        github.com/spf13/cobra@v1.6.1/command.go:1044 +0x3a5
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v1.6.1/command.go:968
github.com/tursodatabase/turso-cli/internal/cmd.Execute()
        github.com/tursodatabase/turso-cli/internal/cmd/root.go:22 +0x1a
main.main()
        github.com/tursodatabase/turso-cli/cmd/turso/main.go:6 +0xf
```

johannesg:
It looks like version generation is dependent on it being in the original git repository. And without it, it simply does not work at all, at least when building with `-tags=prod`.

I feel I have reached my limit here. Maybe it's better to simply remove it until the turso team comes in and adds it themselves? The official version is actually available in homebrew tap, maybe it's better to simply download from there? This is the original install script: https://get.tur.so/install.sh

jhheider:
hm. alright; well we can certainly use git checkout instead of tarball download if that's what it requires. i'll open a new PR; would you be able to test locally to see if it works?

PR #4242, opened 2023-11-28
freedesktop.org/polkit

### LINKED ISSUE #4203: +polkit
https://www.freedesktop.org/software/polkit/releases/

ref: 6583ea5b7c51cf8dcc5bbfe09ab8ce6dea3e423f

and-ri:
closes #4203


and-ri:
This looks like a hardcoded path in the binary :(

jhheider:
> This looks like a hardcoded path in the binary :(

in decreasing order of ease:
- if it's a script, direct edit
- if it'll find a config file, direct config
- if it has a well-known envvar, use with caution
- if it has a secret envvar (gnupg), use with abandon
- if all else fails, do complicated C things in a patch to find your supporting binaries (openssl).

jhheider:
the answer seems likely to involve something like this:
```yaml
- DBUS_SESSION_BUS_ADDRESS=$(dbus-daemon --config-file={{deps.freedesktop.org/dbus.prefix}}/share/dbus-1/session.conf --fork --print-address)
```
but i wasn't able to convince it that was sufficient.

PR #4244, opened 2023-11-28
fix(turso)

jhheider:
ref: https://github.com/pkgxdev/pantry/pull/4221

### DIFF
--- a/projects/turso.tech/package.yml
+++ b/projects/turso.tech/package.yml
@@ -1,6 +1,6 @@
 distributable:
-  url: https://github.com/tursodatabase/turso-cli/archive/refs/tags/v{{version}}.tar.gz
-  strip-components: 1
+  url: git+https://github.com/tursodatabase/turso-cli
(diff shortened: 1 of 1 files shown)


jhheider:
@johannesg can you try this locally and see if the resulting binary works against their servers? `turso version be1a1863`. if not, we'll likely need to vendor their provided binaries. :/

jhheider:
if you don't want to build it yourself, you should be able to download the compiled binaries from the artifacts (for your appropriate platform/architecture) from the bottom of the Summary page of the [workflow run](https://github.com/pkgxdev/pantry/actions/runs/7021407971)

johannesg:
I managed to build it locally and it seems to work!

``` 
> ./turso --version
turso version v0.87.4
```

`./turso db locations` also work with no errors.

Thank you for making it work!

jhheider:
no problem; it seems like they're performing some kind of security check based on git-hash. not something i would have done, but not my place to say.

PR #4266, opened 2023-11-29
+xtls.github.io

and-ri:
https://ossinsight.io/analyze/XTLS/Xray-core


jhheider:
once #4263 is merged, gitlab.org/procps-ng/procps will be available (on linux) which has `pkill`. might be the right answer.

and-ri:
> once #4263 is merged, gitlab.org/procps-ng/procps will be available (on linux) which has `pkill`. might be the right answer.

oh great

PR #4292, opened 2023-12-03
+radicle -- Sovereign p2p network for code collaboration, built on top of Git

kevinmichaelchen:
> Radicle is a sovereign peer-to-peer network for code collaboration, built on top of Git.

### Changes

- New website [**radicle.xyz**](https://radicle.xyz/#try)
- Renamed pantry namespace
- New repo ([`radicle-dev/heartwood`](https://github.com/radicle-dev/heartwood)) has no Git tags :(
- Much fewer provided binaries

### Questions

- Is modifying everything like this okay, or would it be better to preserve the initial `radicle.org` project and create a new one?
- What happens if both pantries provide `rad` but one of the underlying projects has been deprecated?

### DIFF
--- a/projects/radicle.xyz/package.yml
+++ b/projects/radicle.xyz/package.yml
@@ -1,23 +1,23 @@
 distributable:
-  url: https://github.com/radicle-dev/radicle-cli/archive/refs/tags/v{{version}}.tar.gz
+  url: https://github.com/radicle-dev/heartwood/archive/1af9480b088c3eec800040a1bdcd09941a31bcf3.tar.gz
   strip-components: 1
 
 versions:
-  github: radicle-dev/radicle-cli
+  - 2023.11.30
 
 dependencies:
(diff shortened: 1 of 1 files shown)


jhheider:
this is a tough one; it's generally better not to change to names, so i might add the new version as new. in terms of disambiguation, `pkgx` will ask to you to be explicit.

mxcl:
renaming is acceptable.

* we don’t yet guarantee that kind of stability 
* it won’t effect anyone that just types `rad`
* the liklihood that anyone has the FQD name in a script or whatever somewhere is low

In general we would prefer pkgx to be smarter and be able to route to the right package based on version, thus you wouldn't need to delete the old one, typing `rad` would load the latest from the two and specific versions would route to the right package.

# HOWEVER

Don’t delete the old one, just remove the provides. For now this is a perfect solution.

PR #4314, opened 2023-12-05
use brewkit-v1

### LINKED ISSUE #4375: ❌ build issues: pkgx.sh/brewkit=1.0.0
Running log of build failures for pkgx.sh/brewkit=1.0.0

mxcl:
* ci.yml is pull requests
* ci-squared runs to verify changes to the deploy yamls can be tested without uploading anything to bottle storage
* pkger.yml is the workhorse
* built/test/bottle per platform/arch combo are in same workflow
  * this means that bottles will upload if a platform succeeds rather than all bottles failing if one platform fails


mxcl:
Not passing because brewkit v1 needs a formal release before this will be green.

.github/workflows/pkg-platform.yml
```diff
@@ -0,0 +1,210 @@
...
+        description: a JSON array of runner-names
+        required: true
+        type: string
+      test-container:
+        description: >
+          A JSON array of docker image names or `[null]`.
+          Indeed! You cannot leave this as `null` or undefined.
+          Sorry, GHA is not flexible enough to efficiently work around this.
+        required: true
+        type: string
```
jhheider, line 37: simpler to make it `default: '[null]'`?

pkgx.yaml
```diff
@@ -1,5 +1,4 @@
 dependencies:
-  pkgx.sh/brewkit: ^0
+  pkgx.sh/brewkit: ^0 || ^1
```
jhheider, line 2: you like this better than `<2`?

jhheider (approved):
this, too, looks good. i am certain there will be some breakage we cannot find until it is live.

.github/workflows/pkg-platform.yml
mxcl: I think I couldn't easily get this to work because the github actions expressions capabilities are too inadequate.

pkgx.yaml
mxcl: What and allow `v−1` 😂. I prefer this, but yeah either would be fine.

.github/workflows/pkg-platform.yml
mxcl: ie. you cannot `||` off an array and you cannot check if the content of a variable is an array either.

mxcl:
failure because we have exotic unicode characters in our paths and Ruby 2 doesn’t support unicode paths and this failure is during the brewkit `package.yml` test that runs `bk build` lol.

Got a fix in mind.

jhheider:
> Got a fix in mind.

is it not using exotic unicode characters? or am i taking crazy pills? :)

mxcl:
Wasn't sure where I should use `{group: 'linux-x86-64'}` vs `[self-hosted, linux, X64]`

mxcl:
> > Got a fix in mind.
> 
> is it not using exotic unicode characters? or am i taking crazy pills? :)

we swap `/` for a unicode slash for the user’s readability. We could not do this for the “not building in a pantry checkout case” (which builds to XDG_DATA_HOME) but coincidentally our CI runs in a pantry checkout so it would still happen.

I could make the actions/checkout step always go to a subdir to workaround that and probably should.

Or abandon the slash replacement altogether. It's nice to be able to easily browse the various folders in a pantry checkout, but in general I haven't felt it *required* for usage.

jhheider:
> Wasn't sure where I should use `{group: 'linux-x86-64'}` vs `[self-hosted, linux, X64]`

we moved off of `[self-hosted, linux, X64]` due to blocking. it was the slowest, by a wide margin, and singular. by using 4GB GHA instances, its still kind of slow, but it doesn't block.

makes `[self-hosted, macOS, X64]` our bottleneck, but it's much less of a problem, since it's _only_ for build steps.

mxcl:
All green. Will merge tomorrow.

PR #4318, opened 2023-12-05
+angular.dev

and-ri:
https://ossinsight.io/analyze/angular/angular

### DIFF
--- a/projects/angular.dev/package.yml
+++ b/projects/angular.dev/package.yml
@@ -1,0 +1,31 @@
+distributable: ~
+versions:
+  github: angular/angular
+dependencies:
+  nodejs.org: ^20
+build:
+  dependencies:
+    npmjs.com: ^10
+    curl.se: '*'
+  script:
+    - curl -L https://registry.npmjs.org/@angular/cli/-/cli-{{version}}.tgz --output cli-{{version}}.tgz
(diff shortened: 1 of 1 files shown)


projects/angular.dev/package.yml
```diff
@@ -0,0 +1,31 @@
...
+versions:
+  github: angular/angular
+dependencies:
+  nodejs.org: ^20
+build:
+  dependencies:
+    npmjs.com: ^10
+    curl.se: '*'
+  script:
+    - curl -L https://registry.npmjs.org/@angular/cli/-/cli-{{version}}.tgz --output cli-{{version}}.tgz
```
jhheider: why can't this url be used as distributable? you should be able to use `.` as the path once its upacked, i believe.

and-ri: It looks like #3561 
There I was getting an error when using `.`

and-ri: But I'll try now

and-ri:
If we do not use the `.tar.gz.` archive, we get a symbolic link to the build directory in `{{prefix}}/libexec/lib/node_modules/@angular`

jhheider:
try the confusingly-named `--install-links` (which they've inverted the meaning of before :( )

PR #4322, opened 2023-12-06
add `ykman` (yubikey-manager)

MarkusZoppelt:
adds `ykman` ([yubikey-manager](https://github.com/Yubico/yubikey-manager))

### DIFF
--- a/projects/developers.yubico.com/yubikey-manager/package.yml
+++ b/projects/developers.yubico.com/yubikey-manager/package.yml
@@ -1,0 +1,30 @@
+distributable:
+  url: https://github.com/Yubico/yubikey-manager/archive/refs/tags/{{ version }}.tar.gz
+  ref: v{{version}}}
+
+display-name: yubikey-manager
+
+versions:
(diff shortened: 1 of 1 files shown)


projects/developers.yubico.com/yubikey-manager/package.yml
```diff
@@ -0,0 +1,30 @@
...
+  python.org: '^3'
+  linux:
+    pcsclite.apdu.fr: '^2'
+
+build:
+  dependencies:
+    pip.pypa.io: '*'
+    swig.org: '*'
+  script: |
+    pip install install --prefix={{prefix}} yubikey-manager
```
jhheider: this looks odd (`install install`?). if `yubikey-manager` uses a venv, consider the python-venv.sh helper. my guess, from the errors, is that it's not installing everything needed to the prefix.

MarkusZoppelt: > this looks odd (`install install`?)

good catch

jhheider:
thanks for this!

PR #4323, opened 2023-12-06
add `gnu-getopt` (util-linux)

MarkusZoppelt:


jhheider:
i believe github.com/util-linux/util-linux already provides this on linux. do you see that as well?

MarkusZoppelt:
> i believe github.com/util-linux/util-linux already provides this on linux. do you see that as well?

Yeah sorry. I was coming from needing `gnu-getopt` from brew and not finding it here, totally missing that the actual package is already here 🤦‍♂️.

Closing

PR #4324, opened 2023-12-06
add `yubikey-agent`

MarkusZoppelt:
# **HELP NEEDED**

adds `yubikey-agent` but unsure if the `service` can be ported from https://github.com/Homebrew/homebrew-core/blob/d60bc504f4526917b7aae5cdae51f188edcc7d8e/Formula/y/yubikey-agent.rb#L4


jhheider:
we do not have service management at this time. you can add a README.md if you think people will need the help figuring out how to run the server.

jhheider:
looks good!

PR #4334, opened 2023-12-06
add `bazel`/`bazelisk`

MarkusZoppelt:
add [`bazelisk`](https://github.com/bazelbuild/bazelisk) (`bazel`)

### DIFF
--- a/projects/github.com/bazelbuild/bazelisk/package.yml
+++ b/projects/github.com/bazelbuild/bazelisk/package.yml
@@ -2,13 +2,11 @@ distributable:
   url: https://github.com/bazelbuild/bazelisk/archive/refs/tags/v{{version}}.tar.gz
   strip-components: 1
 
-display-name: bazelbuild/bazel
-
 versions:
   github: bazelbuild/bazelisk
 
 dependencies:
-  curl.se/ca-certs: ^2023.8.22
+  curl.se/ca-certs: '*'
 
 build:
   dependencies:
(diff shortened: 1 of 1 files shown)


projects/github.com/bazelbuild/bazelisk/package.yml
```diff
@@ -0,0 +1,32 @@
+distributable:
+  url: https://github.com/bazelbuild/bazelisk/archive/refs/tags/v{{version}}.tar.gz
+  ref: v{{version}}}
```
jhheider: `ref:` is used with git checkouts; you want `strip-compents: 1`, so the name-version of the package is stripped from the extraction path.

projects/github.com/bazelbuild/bazelisk/package.yml
```diff
@@ -0,0 +1,32 @@
...
+versions:
+  github: bazelbuild/bazelisk
+
+build:
+  dependencies:
+    go.dev: '^1.20'
+  script:
+    - go mod download
+    - go build -ldflags "$LDFLAGS" -o "{{prefix}}/bin/bazelisk"
+    - ln -n {{prefix}}/bin/bazelisk {{prefix}}/bin/bazel
```
jhheider: this should definitely be: `ln -s bazelisk {{prefix}}/bin/bazel`. using an absolute path will make it nonportable.

projects/github.com/bazelbuild/bazelisk/package.yml
MarkusZoppelt: done!

projects/github.com/bazelbuild/bazelisk/package.yml
MarkusZoppelt: changed, thx!

jhheider:
`failed to verify certificate: x509: certificate signed by unknown authority`

needs a runtime-dep on `curl.se/ca-certs`

PR #4341, opened 2023-12-07
+openai.com/whisper

and-ri:
https://ossinsight.io/analyze/openai/whisper

### DIFF
--- a/projects/openai.com/whisper/package.yml
+++ b/projects/openai.com/whisper/package.yml
@@ -7,7 +7,6 @@ dependencies:
   python.org: ^3.11
   ffmpeg.org: ^6.1
   huggingface.co: ^0.19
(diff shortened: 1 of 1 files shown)


projects/openai.com/whisper/package.yml
```diff
@@ -0,0 +1,27 @@
+distributable:
+  url: https://github.com/openai/whisper/archive/v{{version.major}}.tar.gz
+  strip-components: 1
+versions:
+  github: openai/whisper
+dependencies:
+  python.org: ^3.11
+  ffmpeg.org: ^6.1
+  huggingface.co: ^0.19
+  llvm.org: ^14 # Issue for newer LLVM: https://github.com/numba/llvmlite/issues/914
```
jhheider: is llvm a run-time dependency?

and-ri: The homebrew formula says yes, but let's try without it

PR #4349, opened 2023-12-08
Added expo.dev

michaelessiet:
- created eas.dev
- added eas.dev
- added the build from source argument

### DIFF
--- a/projects/eas.dev/eas-cli/package.yml
+++ b/projects/eas.dev/eas-cli/package.yml
@@ -1,125 +1,40 @@
-# this is a basic template for packages we strongly recommend looking at the
-# contents of both:
-#
-# * https://github.com/pkgxdev/pantry
-#
-# as a learning guide and reference. For example, if you are package a `go`
-# project, then browse the listings and find a go package.
-
-
-# this is the source tarball. Swap the version out for {{version}} or we won’t
-# be able to upgrade the package when it updates!
 distributable:
-  url: https://example.com/v{{version}}.tar.gz
-  # this obscure option dictates how many components are removed from the
-  # extracted tarball. It only works with tarballs. You almost always want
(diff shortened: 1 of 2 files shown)


jhheider:
i'd try python.org~3.10 or ~3.11. a lot of stuff still doesn't work with 3.12.

michaelessiet:
darwin systems seem to pass without the python dependency. But Linux fails to pass either way. I can make the dependency optional based on system right?

jhheider:
you can, but we shouldn't use system python if we can avoid it.

the compiler error for node-gyp is interesting. warrants more research. possibly it doesn't like llvm17. you can try setting build.dependencies.linux.llvm.org: <17

michaelessiet:
2 failures from pkgs not being found, when they certainly exist. That's weird

jhheider:
the build issues could be node versions. if you want main-line releases, try something like `^16 || ^18 || ^20 || ^22`

michaelessiet:
Let me try setting it to ~20

michaelessiet:
Should I remove the build from source flag?

jhheider:
oh, definitely if it's not needed.

michaelessiet:
Well that kinda worked. I wonder what's up with aarch linux

jhheider:
getting close. `vaultproject.io` has this comment:
```yaml
# python 3.11 removes 'rU' mode which breaks node-gyp
    python.org: ~3.10
```
i wonder if that's the last piece.

michaelessiet:
I set python to 3.10 tho. Same as theirs

michaelessiet:
Success! Thanks you

jhheider:
> Success! Thanks you

there's a million tricks to try. that's the dark secret of packaging.

PR #4366, opened 2023-12-11
add openjdk to appium deps

j1philli:
it isn't needed for building but is needed for most operations of the package.
Sorry for not initially including this.

### DIFF
--- a/projects/appium.io/package.yml
+++ b/projects/appium.io/package.yml
@@ -28,5 +28,5 @@ provides:
   - bin/appium
 
 test: 
-  appium --version | grep {{version}}
-  appium driver install uiautomator2
+  - appium --version | grep {{version}}
+  - appium driver install uiautomator2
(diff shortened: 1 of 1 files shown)


jhheider:
can we add a test that requires this? i know it can be hard with server-type systems.

j1philli:
> can we add a test that requires this? i know it can be hard with server-type systems.

I don't think so. I mispoke, appium doesn't use it directly but a lot of the plugins it grabs itself will use openjdk but appium can't grab openjdk for you so I thought it would be handy to already have.
The plugins then require custom code to be run against an emulated device in order to actually use openjdk.

jhheider:
maybe install a plugin in the test?

j1philli:
> maybe install a plugin in the test?

added install of a driver that uses the JDK. I don't know if it uses it during the install but actually running something through everything would require writing a test file and emulating a device in the test

PR #4452, opened 2023-12-15
fix(zlib)

jhheider:
boost complaining about -fPIC:

https://github.com/pkgxdev/pantry/actions/runs/7226847979/job/19693299131

### DIFF
--- a/projects/zlib.net/package.yml
+++ b/projects/zlib.net/package.yml
@@ -9,6 +9,9 @@ build:
   script: |
     ./configure --prefix="{{prefix}}"
     make --jobs {{ hw.concurrency }} install
+  env:
+    linux/aarch64:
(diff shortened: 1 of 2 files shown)


projects/zlib.net/package.yml
```diff
@@ -9,6 +9,9 @@ build:
   script: |
     ./configure --prefix="{{prefix}}"
     make --jobs {{ hw.concurrency }} install
+  env:
+    linux/aarch64:
+      CFLAGS: '$CFLAGS -fPIC'
```
mxcl, line 14: This is what we do in brewkit, should we be doing this for all Linux then? I dunno.

image

jhheider: i think yes. it's non-harmful, minor performance cost, and many things dislike non-aligned binaries.

PR #4456, opened 2023-12-16
fix(pandoc)

jhheider:
and possibly cabal; i worry that the inclusion of gcc is breaking stuff


jhheider:
removing utf-8 from brewkit paths fixed most of this.

PR #4459, opened 2023-12-16
+google.com/gcloud

### LINKED ISSUE #2648: Add gcloud CLI
The title says it all

and-ri:
closes #2648

### DIFF
--- a/projects/google.com/gcloud/package.yml
+++ b/projects/google.com/gcloud/package.yml
@@ -43,7 +43,6 @@ provides:
   - bin/gsutil
   - bin/bq
   - bin/anthoscli
-  - bin/docker-creditnail-gcloud
 
 test:
   - gcloud --version | grep {{version}}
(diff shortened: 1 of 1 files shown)


projects/google.com/gcloud/package.yml
```diff
@@ -0,0 +1,49 @@
...
+dependencies:
+  python.org: ^3.11
+
+build:
+  dependencies:
+    curl.se: '*'
+  script:
+    - mkdir -p {{prefix}}/libexec
+    - curl -L $DISTRIBUTABLE_URL | tar -xz --strip-components=1
+    - ./install.sh $ARGS
```
jhheider, line 18: should this be `warnings: vendored` or does `install.sh` build something?

and-ri: > should this be `warnings: vendored` or does `install.sh` build something?

you're right, we can definitely add `warnings: vendored` here

PR #4513, opened 2023-12-20
rebuild with -fPIC (fbthrift)

and-ri:


projects/facebook.com/fbthrift/package.yml
```diff
@@ -27,6 +27,10 @@ build:
     - cmake -S . -B build/shared $CMAKE_ARGS
     - cmake --build build/shared
     - cmake --install build/shared
+    - run: |
+        sed -i.bak "s|{{pkgx.prefix}}|\${_IMPORT_PREFIX}/../../..|g" FBThriftTargets.cmake
```
jhheider: this should be handled by fix-up.ts in brewkit (though it currently has a bug being fixed in https://github.com/pkgxdev/brewkit/pull/233

PR #4525, opened 2023-12-21
Enable YJIT; Fixes #3538

### LINKED ISSUE #3538: Ruby is missing YJIT support
When Ruby is built it needs rustc sothat YJIT gets also built.

mxcl:


.github/workflows/pkg-platform.yml
```diff
@@ -218,7 +218,7 @@ jobs:
         if: ${{ ! inputs.dry-run }}
 
   complain:
-    needs: [build, test, bottle]
+    needs: bottle
```
mxcl, line 221: this was pointless readability optimization that I hope doesn’t cause it to fail to complain on build failure.

PR #4533, opened 2023-12-21
fix(zlib)

jhheider:
llvm17 breaks shlib support. this fixes that.

### DIFF
--- a/projects/zlib.net/package.yml
+++ b/projects/zlib.net/package.yml
@@ -10,8 +10,11 @@ build:
     ./configure --prefix="{{prefix}}"
     make --jobs {{ hw.concurrency }} install
   env:
-    linux/aarch64:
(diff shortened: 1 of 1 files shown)


jhheider:
```
gcc -shared -Wl,-soname,libz.so.1,--version-script,zlib.map -fPIC -Wl,--undefined-version -fPIC -D_LARGEFILE64_SOURCE=1 -o libz.so.1.3 adler32.lo crc32.lo deflate.lo infback.lo inffast.lo inflate.lo inftrees.lo trees.lo zutil.lo compress.lo uncompr.lo gzclose.lo gzlib.lo gzread.lo gzwrite.lo  -lc 
rm -f libz.so libz.so.1
ln -s libz.so.1.3 libz.so
ln -s libz.so.1.3 libz.so.1
```

good

PR #4554, opened 2023-12-22
+nx.dev

diegoholiveira:


projects/nx.dev/package.yml
```diff
@@ -0,0 +1,32 @@
...
+    - cp props/nx.json {{prefix}}
+  env:
+    ARGS:
+      - --global
+      - --install-links
+      - --prefix={{prefix}}
+
+test:
+  script: |
+    npm i nx
```
jhheider: if `nx` is installed properly, then this shouldn't be necessary, i don't believe.

projects/nx.dev/package.yml
```diff
@@ -0,0 +1,32 @@
...
+
+dependencies:
+  nodejs.org: ^18
+
+build:
+  dependencies:
+    npmjs.com: ^10
+  script:
+    - npm i $ARGS .
+    - cp props/nx.json {{prefix}}
```
jhheider: is this a test fixture? it shouldn't be necessary to copy if so.

diegoholiveira:
@jhheider thanks for reviewing it. I fixed the issues.

diegoholiveira:
@jhheider github releases is not synced with npmjs regard the released versions. anything that I can do here? Or it's just about wait it sync?

jhheider:
> @jhheider github releases is not synced with npmjs regard the released versions. anything that I can do here? Or it's just about wait it sync?


Yes, some packages watch npm instead. If you search the pantry for `registry.npmjs` you can find several examples that do that.

diegoholiveira:
> > @jhheider github releases is not synced with npmjs regard the released versions. anything that I can do here? Or it's just about wait it sync?
> 
> Yes, some packages watch npm instead. If you search the pantry for `registry.npmjs` you can find several examples that do that.

Not sure if I would change that because there's a `9999.0.1` version published: https://www.npmjs.com/package/nx/v/9999.0.1

wdyt?

jhheider:
use the `versions.ignore` key: https://github.com/pkgxdev/pantry/blob/dc4060218a00ad6c3ba5c6378dffa8c2630575ce/projects/gnome.org/librsvg/package.yml#L5-L7

diegoholiveira:
I was reading the `handleURLVersions` function and it does not have an `ignore` option (it's only available to GitHub and GitLab versions). 

Can we move forward with this solution knowing that GitHub + npm sometimes will not be synced? This never happened with other npm based packages?

jhheider:
there are other githubs which don't match npm versioning. ignore functionality should definitely be added to that watcher.

you might be able to fake it by using a match that looks for `/\d{1,3}(\.\d+)*/`; i don't know offhand if Deno's regexp uses that extension, but it probably does.

jhheider:
this looks good, but let's see if we get your brewkit PR through. that'll simplify version management considerably.

mxcl:
brewkit PR merged

PR #4574, opened 2023-12-22
fix docker-credential-gcloud symlink path

uesyn:


jhheider:
this seems like a good idea. :)

PR #4602, opened 2023-12-26
Props no longer need escaping which means fix them

mxcl:
Strictly this was a breaking change that should require brewkit v2.

I didn’t consider this, and it’s too late. My bad.


mxcl:
Strictly this was a breaking change that should require brewkit v2.

I didn’t consider this, and it’s too late. My bad.

mxcl:
Build issue unrelated to this PR

PR #4619, opened 2023-12-28
Fix build for older Ruby versions

m1ch4ls:


jhheider:
all looks reasonable.

PR #4625, opened 2023-12-28
fix use of `bundler` as a gem

mxcl:
many little fixes and tweaks here. Requires newer brewkit.


projects/rubygems.org/package.yml
```diff
@@ -24,12 +24,12 @@ build:
         --no-rdoc
 
     - run: |
-        patch -p1 < "$SRCROOT"/props/fit-n-finish.patch
+        patch -p1 < "$SRCROOT"/props/fit-n-finish.1.patch
         patch -p1 < "$SRCROOT"/props/fit-n-finish.2.patch
-        sed -i -e 's/Gem.default_dir/Gem.default_system_dir/' lib/rubygems/path_support.rb
+        sed -ie 's/Gem.default_dir/Gem.default_system_dir/' lib/rubygems/path_support.rb
```
jhheider, line 29: did you check this? i think this makes a backup with `e` on the end:
```sh
$ echo hi >foo
$ pkgx +sed sed -ie 's/hi/bye/' foo
$ cat foo
bye
$ ls
foo	fooe
```

mxcl: Oops if true.

PR #4661, opened 2023-12-29
Added auto-update for geni

emilpriver:
Enabled auto-update for geni


jhheider:
thanks, @emilpriver ! always great when maintainers take an interest.

emilpriver:
> thanks, @emilpriver ! always great when maintainers take an interest.

Yeah indeed, I did find pkgx to be a handy tool so adding geni here is really nice

PR #4683, opened 2024-01-01
Geni: Changed so pkgx don't change src/main.rs if version is more then 0.0.3

emilpriver:


jhheider:
happy to apply this. fortunately, it was "safe" for future versions, (I suppose, until 0.0.30).

PR #4684, opened 2024-01-01
dotnet: use gh tags & azureedge links

MrLuje:
azureedge links are the defaults used by [dotnet-install-script](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script#options:~:text=https%3A//dotnetcli.azureedge.net/dotnet)


jhheider:
ah, simplification! thanks, @MrLuje !

PR #4736, opened 2024-01-04
fix(ollama)

jhheider:
remove build hack (addressed upstream: https://github.com/jmorganca/ollama/pull/1780)


jhheider:
big thanks to @jmorganca for the super-quick turnaround upstream.

PR #4798, opened 2024-01-10
+clisp.org

and-ri:
clisp => texlive

### DIFF
--- a/projects/clisp.org/package.yml
+++ b/projects/clisp.org/package.yml
@@ -38,6 +38,7 @@ build:
       - --disable-silent-rules
       - --with-readline=yes
       - --with-libsigsegv-prefix={{deps.gnu.org/libsigsegv.prefix}}
+      - --enable-portability
     darwin:
       ARGS:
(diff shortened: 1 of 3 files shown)


jhheider:
```
/Users/builder/actions-runner/_work/_actions/pkgxdev/brewkit/v1/lib/bin/fix-machos.rb:57:in `fix': undefined method `Error' for #<Fixer:0x0000000108b5a0c8 @file=#<MachO::MachOFile:0x0000000108b127c8 @filename="/opt/clisp.org/v2.49.92/bin/clisp.dSYM/Contents/Resources/DWARF/clisp", @options={}, 
```
should be able to just remove the `.dSYM` directories, as they're just debugging symbols. i _thought_ we had other packages that did this, but i'm not finding them.

jhheider:
`clisp: /opt/clisp.org/v2.49.92+brewing/lib/clisp-2.49.92/base/lisp.run: No such file or directory`

and that one looks like some hard-coded paths need to be genericized.

PR #4806, opened 2024-01-11
parse image magick versions to semver

mxcl:


mxcl:
oh god:

```bash
# libheif.pc
prefix=${pcfiledir}/../..+brewing
```

The `+brewing` isn't corrected in the pcfiles fixer either. This is possibly insidious!

My bad Jacob and co.

edit: oh perhaps we fixed this already and we just need rebuilds? Checking

jhheider:
> we fixed this already and we just need rebuilds

yup.

PR #4818, opened 2024-01-12
+opensearch

scottjg:
my attempt at adding the opensearch package. there is an existing homebrew package that i used as a reference. this is my first time contributing to the pantry so please let me know if any changes are needed.

i would also like to contribute building the k-NN plugin and native binaries for opensearch. this is normally shipped with opensearch binaries by the upstream project but annoyingly they don't include the native libraries for macos despite them working fine on the platform. i was hoping someone could maybe point me at an example of including a second distribution tarball.

### DIFF
--- a/projects/opensearch.org/package.yml
+++ b/projects/opensearch.org/package.yml
@@ -27,7 +27,7 @@ build:
     - gradle -Dbuild.snapshot=false ":distribution:archives:no-jdk-{{hw.platform}}-tar:assemble"
     - run: tar --strip-components=1 -xf distribution/archives/no-jdk-{{hw.platform}}-tar/build/distributions/opensearch-*.tar.gz
       working-directory: ${{prefix}}
-    - run: sed -i 's|#\s*cluster.name: .*|cluster.name: opensearch_pkgx|' opensearch.yml
+    - run: 'sed -i "s|#\s*cluster.name: .*|cluster.name: opensearch_pkgx|" opensearch.yml'
       working-directory: ${{prefix}}/config
 
     # checkout k-NN plugin
--- a/projects/opensearch.org/package.yml
+++ b/projects/opensearch.org/package.yml
@@ -25,14 +25,13 @@ build:
 
   script:
     - gradle -Dbuild.snapshot=false ":distribution:archives:no-jdk-{{hw.platform}}-tar:assemble"
-    - mkdir -p "{{prefix}}"
-    - tar --strip-components=1 -xf distribution/archives/no-jdk-{{hw.platform}}-tar/build/distributions/opensearch-*.tar.gz -C "{{prefix}}"
+    - run: tar --strip-components=1 -xf distribution/archives/no-jdk-{{hw.platform}}-tar/build/distributions/opensearch-*.tar.gz
+      working-directory: ${{prefix}}
+    - run: sed -i 's|#\s*cluster.name: .*|cluster.name: opensearch_pkgx|' opensearch.yml
+      working-directory: ${{prefix}}/config
+
+    # checkout k-NN plugin
     - run: |
-        sed -i 's|#\s*cluster.name: .*|cluster.name: opensearch_pkgx|' "{{prefix}}/config/opensearch.yml"
-    - run: |
-        # checkout k-NN plugin
-        mkdir -p k-NN
-        cd k-NN
         if [ -d .git ]; then
           git fetch
         else
@@ -42,12 +41,16 @@ build:
         git reset --hard
         git submodule foreach --recursive git reset --hard
         git submodule update --init --recursive
+      working-directory: k-NN
+
+    # workarounds for m1 build. see: https://github.com/opensearch-project/k-NN/blob/main/DEVELOPER_GUIDE.md#extra-setup-for-mac-m1-machines
     - run: |
-        # workarounds for m1 build. see: https://github.com/opensearch-project/k-NN/blob/main/DEVELOPER_GUIDE.md#extra-setup-for-mac-m1-machines
-        sed -i -e 's/-march=native/-mcpu=apple-m1/g' jni/external/nmslib/similarity_search/CMakeLists.txt
-        sed -i -e 's/-mcpu=apple-a14/-mcpu=apple-m1/g' jni/external/nmslib/python_bindings/setup.py
-        sed -i -e 's/__aarch64__/__undefine_aarch64__/g' jni/external/faiss/faiss/utils/distances_simd.cpp
+        sed -i -e 's/-march=native/-mcpu=apple-m1/g' nmslib/similarity_search/CMakeLists.txt
+        sed -i -e 's/-mcpu=apple-a14/-mcpu=apple-m1/g' nmslib/python_bindings/setup.py
+        sed -i -e 's/__aarch64__/__undefine_aarch64__/g' faiss/faiss/utils/distances_simd.cpp
       if: darwin/aarch64
+      working-directory: k-NN/jni/external
+
     - run: |
         IFS=:
         for p in ${LD_LIBRARY_PATH}; do
@@ -56,6 +59,7 @@ build:
             break
           fi
         done
+        unset IFS
         [ -z "${libomp_path}" ] && echo "libomp.dylib not found" && exit 1
 
         sed -i -e "s|/usr/local/opt/libomp/|${libomp_path}/|g" jni/CMakeLists.txt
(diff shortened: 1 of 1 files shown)


projects/opensearch.org/package.yml
```diff
@@ -0,0 +1,49 @@
...
+dependencies:
+  openjdk.org: '*'
+
+build:
+  dependencies:
+    gradle.org: '*'
+  script:
+    - gradle -Dbuild.snapshot=false ":distribution:archives:no-jdk-{{hw.platform}}-tar:assemble"
+    - mkdir -p "{{prefix}}"
+    - tar --strip-components=1 -xf distribution/archives/no-jdk-{{hw.platform}}-tar/build/distributions/opensearch-*.tar.gz -C "{{prefix}}"
```
jhheider: what a weird way to have to do it; sadly, python poetry is similar.

projects/opensearch.org/package.yml
```diff
@@ -0,0 +1,49 @@
...
+    - run: |
+        sed -i.bak 's|#\s*cluster.name: .*|cluster.name: opensearch_pkgx|' "{{prefix}}/config/opensearch.yml"
```
jhheider: since build-envs include `gnu.org/sed` you can just use `-i` instead of `-i.bak`. saves you having to clean up the backup file.

projects/opensearch.org/package.yml
```diff
@@ -0,0 +1,49 @@
...
+  - bin/opensearch-plugin
+  - bin/opensearch-shard
+
+test:
+  dependencies:
+     gnu.org/coreutils: ^9
+     stedolan.github.io/jq: "*"
+     curl.se: "*"
+  script: |
+    opensearch-plugin list
```
jhheider: this might be simpler using arrays:
```yaml
    - opensearch-plugin list
    # While we'd love a good test like this, `initdb` doesn't run as root, and managing
    # all the environment passthrough with `sudo` is a nightmare.
    - run: opensearch -version |grep "Version: {{version}}"
      if: linux

    - run: |
        mkdir -p test/{data,logs}
        PORT=$(shuf -i 2000-65000 -n 1)
        opensearch -Ehttp.port=$PORT -Epath.data=$PWD/test/data -Epath.logs=$PWD/test/logs &
        pid=$!
        for i in $(seq 1 30); do
          curl -k --silent --fail http://localhost:$PORT/ > output.txt && break || sleep 1
        done
        kill $pid
        test "$(jq .version.number output.txt)" = \"{{version}}\"
      if: darwin
```

jhheider (approved):
looks good; just some style notes.

jhheider:
> i would also like to contribute building the k-NN plugin and native binaries for opensearch. this is normally shipped with opensearch binaries by the upstream project but annoyingly they don't include the native libraries for macos despite them working fine on the platform. i was hoping someone could maybe point me at an example of including a second distribution tarball.

you should be able to add it to the install script, assuming you can build all the required parameters. would you like me to wait until you add that to merge?

scottjg:
sure - let me dig into it.

projects/opensearch.org/package.yml
```diff
@@ -0,0 +1,150 @@
...
+      if: darwin/aarch64
+    # this is recommended in https://github.com/opensearch-project/k-NN/blob/45e9e542aef60ef7073ee726e6ac14dec27bfa04/scripts/build.sh#L91-L94
+    - run: sed -i -e 's/-march=native/-march=x86-64/g' ./jni/external/nmslib/similarity_search/CMakeLists.txt
+      if: x86-64
+    -  run: |
+        ./gradlew build --refresh-dependencies -x integTest -x test -DskipTests=true -Dopensearch.version={{version}} -Dbuild.snapshot=false -Dbuild.version_qualifier=
+        ./gradlew publishPluginZipPublicationToZipStagingRepository -Dopensearch.version={{version}} -Dbuild.snapshot=false -Dbuild.version_qualifier=
+        ./gradlew publishPluginZipPublicationToMavenLocal -Dbuild.snapshot=false -Dbuild.version_qualifier= -Dopensearch.version={{version}}
+        mkdir -p ./build/distributions/lib
+        cp -v ./jni/release/libopensearchknn* ./build/distributions/lib
```
scottjg, line 93: this build generates native libraries with the `.jnilib` extension. when i was packaging this for my team (pre-pkgx) i had to codesign and notarize the libraries. i know you guys have some infra around this stuff already but i just wanted to make sure since it wasn't a dylib that it would be ok?

it seems to work locally fwiw.

jhheider: generally, yes, our codesigning logic has worked for over a year. the only package i can see that's needed anything different is: https://github.com/pkgxdev/pantry/blob/main/projects/github.com/sindresorhus/macos-term-size/package.yml

scottjg:
oops, sorry.. didn't see that force push. might've clobbered each other there.

jhheider:
no worries; i just did update-with-rebase

scottjg:
great progress here, but still a few issues.

- mac x86-64 build: seems to be having some kind of networking issue (maybe?) fetching a manifest
- mac m1 build: rpaths are not properly cleaned up in the jnilibs and they fail to load
- linux builds: likely similar rpath issues but tests don't run on them

jhheider:
from boost.org:
```yaml
    # boost.org has libs that end up with name @rpath/libboost_atomic.dylib (offset 24)
    # so we need to add @loader_path to the rpath
    - run: |
        for LIB in *.dylib; do
          install_name_tool -add_rpath @loader_path $LIB
        done
      if: darwin
```

i'll re-run darwin-x86_64 once the rest finish (though it should have the same issue).

linux doesn't use `rpath`, so that's not likely to affect them. it uses `LD_LIBRARY_PATH`, which we set. looks like linux-aarch64 passed (unless you mean it has no tests).

jhheider:
that .meta url is fine, so something is up with `JsonSlurper` on darwin-x86-64. i wonder if it's relying on a system tool that isn't present. probably have to try one of these to see why:
```
* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
```

scottjg:
> linux doesn't use rpath, so that's not likely to affect them. it uses LD_LIBRARY_PATH, which we set. looks like linux-aarch64 passed (unless you mean it has no tests).

the real tests i wrote in the yml only run on darwin. opensearch can't run as root (mysql package has a similar comment in it), so i just output the version instead which can at least sanity check the package. the problem is that the knn plugin is lazily loaded so that's not really tested at all on linux.

maybe there's some simple-ish solution munging env with sudo to make that work.

> that .meta url is fine, so something is up with JsonSlurper on darwin-x86-64. i wonder if it's relying on a system tool that isn't present. probably have to try one of these to see why:

makes sense. unfortunately past my bedtime so will have to try in the morning.

getting close though! m1 build is totally working end to end.

scottjg:
> mysql package has a similar comment in it

err, i meant postgres https://github.com/pkgxdev/pantry/blob/fde48d56a4d0369581b019ecef9eee456f163e31/projects/postgresql.org/package.yml#L80-L85

scottjg:
not sure what happened with that `JsonSlurper` issue but it's gone this morning. i also sanity checked the linux build locally and it seemed to work. i think this is ready for a final review.

jhheider:
looks good. i tried to clean up the various `cd`s with `working-directories:` to make the flow more clear. however, if i didn't follow it, i might have messed it up. we'll know after it builds.

scottjg:
nice!

jhheider:
And there we go. That was a big one. Thanks @scottjg !

jhheider:
hm, failing to find libc on macos12: https://github.com/pkgxdev/pantry/actions/runs/7516912520/job/20462627197

possibly we need to add an rpath somewhere?

projects/opensearch.org/package.yml
```diff
@@ -0,0 +1,166 @@
...
+        ./gradlew publishPluginZipPublicationToMavenLocal -Dbuild.snapshot=false -Dbuild.version_qualifier= -Dopensearch.version={{version}}
+      working-directory: k-NN
+    - run: |
+        mkdir -p ./build/distributions/lib
+        cp -v ./jni/release/libopensearchknn* ./build/distributions/lib
+        cd ./build/distributions
+        zip -r opensearch-knn-{{version}}.0.zip lib/
+        {{prefix}}/bin/opensearch-plugin install --batch file:`pwd`/opensearch-knn-{{version}}.0.zip
+      working-directory: k-NN
+    - run: echo 'export OPENSEARCH_JAVA_OPTS="-Djava.library.path=$OPENSEARCH_HOME/plugins/opensearch-knn/lib $OPENSEARCH_JAVA_OPTS"' >> opensearch-env
```
scottjg, line 98: if it's happening when opensearch boots, then it shouldn't be loading the knn plugin yet. it might be related to this. 

i tried to modify the java library path here so that the native plugin dependencies would be loaded properly. maybe it needs some more paths? hmm...

scottjg:
so, i guess it's in opensearch itself where it's trying to call `setrlimit`  which presumably is why it's trying to load libc

```
java.lang.UnsatisfiedLinkError: 'int org.opensearch.bootstrap.JNACLibrary.setrlimit(int, org.opensearch.bootstrap.JNACLibrary$Rlimit)'
	at org.opensearch.bootstrap.JNACLibrary.setrlimit(Native Method) ~[opensearch-2.11.1.jar:2.11.1]
	at org.opensearch.bootstrap.SystemCallFilter.bsdImpl(SystemCallFilter.java:609) ~[opensearch-2.11.1.jar:2.11.1]
	at org.opensearch.bootstrap.SystemCallFilter.init(SystemCallFilter.java:669) ~[opensearch-2.11.1.jar:2.11.1]
	at org.opensearch.bootstrap.JNANatives.tryInstallSystemCallFilter(JNANatives.java:281) ~[opensearch-2.11.1.jar:2.11.1]
	at org.opensearch.bootstrap.Natives.tryInstallSystemCallFilter(Natives.java:128) ~[opensearch-2.11.1.jar:2.11.1]
	at org.opensearch.bootstrap.Bootstrap.initializeNatives(Bootstrap.java:129) ~[opensearch-2.11.1.jar:2.11.1]
	at org.opensearch.bootstrap.Bootstrap.setup(Bootstrap.java:191) ~[opensearch-2.11.1.jar:2.11.1]
	at org.opensearch.bootstrap.Bootstrap.init(Bootstrap.java:404) ~[opensearch-2.11.1.jar:2.11.1]
	at org.opensearch.bootstrap.OpenSearch.init(OpenSearch.java:180) ~[opensearch-2.11.1.jar:2.11.1]
	at org.opensearch.bootstrap.OpenSearch.execute(OpenSearch.java:171) ~[opensearch-2.11.1.jar:2.11.1]
	at org.opensearch.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:104) ~[opensearch-2.11.1.jar:2.11.1]
	at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:138) ~[opensearch-cli-2.11.1.jar:2.11.1]
	at org.opensearch.cli.Command.main(Command.java:101) ~[opensearch-cli-2.11.1.jar:2.11.1]
	at org.opensearch.bootstrap.OpenSearch.main(OpenSearch.java:137) ~[opensearch-2.11.1.jar:2.11.1]
	at org.opensearch.bootstrap.OpenSearch.main(OpenSearch.java:103) ~[opensearch-2.11.1.jar:2.11.1]
```

projects/opensearch.org/package.yml
jhheider: does it maybe need a `runtime.env.OPENSEARCH_JAVA_OPTS:` key?

scottjg:
https://github.com/pkgxdev/pantry/pull/4841

PR #4819, opened 2024-01-12
Add ffmpeg support for drawtext filter

scottjg:
requires linking against harfbuzz and freetype.

### DIFF
--- a/projects/ffmpeg.org/package.yml
+++ b/projects/ffmpeg.org/package.yml
@@ -20,6 +20,8 @@ provides:
 dependencies:
   lame.sourceforge.io: '>=3.98.3'
   libsdl.org: ^2
+  freetype.org: "*"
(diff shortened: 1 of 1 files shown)


projects/ffmpeg.org/package.yml
```diff
@@ -20,6 +20,8 @@ provides:
 dependencies:
   lame.sourceforge.io: '>=3.98.3'
   libsdl.org: ^2
+  freetype.org: "*"
+  harfbuzz.org: "*"
```
jhheider: due to the way linking works, library dependencies should be specified as `^{{major}}` if there's no tighter requirement.

jhheider (approved):
great addition; just lock the library deps and it should be good.

jhheider:
looks good. thanks!

PR #4825, opened 2024-01-12
fix(hunspell) & more tests

and-ri:
To work with Hunspell, at least one dictionary is required. I suggest adding at least an English dictionary

### DIFF
--- a/projects/hunspell.github.io/package.yml
+++ b/projects/hunspell.github.io/package.yml
@@ -12,11 +12,20 @@ dependencies:
   invisible-island.net/ncurses: '*'
   gnu.org/gettext: '*'
 
+runtime:
+  env:
+    DICPATH: $DICPATH:{{prefix}}/share/dictionaries/en
+
 build:
-  script: |
-    ./configure $ARGS
-    make --jobs {{ hw.concurrency }}
-    make --jobs {{ hw.concurrency }} install
+  dependencies:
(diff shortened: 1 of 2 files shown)


projects/hunspell.github.io/package.yml
```diff
@@ -12,11 +12,20 @@ dependencies:
...
-    make --jobs {{ hw.concurrency }} install
+  dependencies:
+    git-scm.org: '*'
+  script:
+    - ./configure $ARGS
+    - make --jobs {{ hw.concurrency }}
+    - make --jobs {{ hw.concurrency }} install
+    - git clone https://github.com/LibreOffice/dictionaries
+    - mkdir -p {{prefix}}/share/dictionaries
+    - cp -r dictionaries/en {{prefix}}/share/dictionaries/
```
jhheider, line 28: how big are the dictionaries? would it make sense to provide them all? looks like a given file is a few KB, but there's probably a lot of them.

and-ri: @jhheider  we actually only need the `en_US.aff` (3 KB) and `en_US.dic` (552 KB) files

But now I'm already doubting whether it's really necessary... I thought it would help to build [enchant](https://github.com/pkgxdev/pantry/pull/3938) with `hunspell`. But even downloading dictionaries does not help

PR #4827, opened 2024-01-12
+dotenvx

mxcl:


projects/dotenvx.com/package.yml
```diff
@@ -0,0 +1,21 @@
+distributable:
+  url: https://registry.npmjs.org/@dotenvx/dotenvx/-/dotenvx-{{version}}.tgz
+  strip-components: 1
+
+versions:
+  github: dotenvx/dotenvx
```
jhheider: might want to track `npm:` versions instead of `github:`, so we don't end up building before the sources go up.

mxcl: good call

PR #4830, opened 2024-01-12
fix(openjdk^17)

### LINKED ISSUE #4829: version request: openjdk.org@^17
while working on https://github.com/pkgxdev/pantry/pull/4818, unfortunately i think the knn plugin (you need it for ai vector search) requires jdk17 to build. it seems like the existing package.yml for openjdk.org works if you replace the distribution and BOOT_JDK for the older version... but i'm not sure how to modify the distribution to list different urls for different versions. 

in this case, the current url is:
https://github.com/openjdk/jdk20u/archive/jdk-20.0.2+9.tar.gz
while the jdk17 url is:
https://github.com/openjdk/jdk17u/archive/jdk-17.0.10+6.tar.gz
so it's not _just_ the full version number that needs to be replaced

any advice?

jhheider:
closes #4829

### DIFF
--- a/projects/openjdk.org/package.yml
+++ b/projects/openjdk.org/package.yml
@@ -1,10 +1,19 @@
 distributable:
-  url: https://github.com/openjdk/jdk20u/archive/jdk-20.0.2+9.tar.gz
-  strip-components: 1
+  # TODO: more majors
+  - url: https://github.com/openjdk/jdk20u/archive/{{version.tag}}.tar.gz
+    strip-components: 1
+  - url: https://github.com/openjdk/jdk17u/archive/{{version.tag}}.tar.gz
+    strip-components: 1
 
 versions:
-  - 20.0.2.9
-  #FIXME
+  # TODO: more majors
+  - github: openjdk/jdk20u/tags
+    # each repo contains all the prior repo tags at the time of the fork,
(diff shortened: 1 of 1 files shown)


scottjg:
oh wow! thanks for the quick turnaround.

jhheider:
need to resolve https://github.com/pkgxdev/brewkit/issues/270, and also get a v16 or v17 BOOT_JDK. the latter will probably use something like: https://github.com/pkgxdev/pantry/blob/2e4cdfa83b884f181065bf436c6135ad41d3c8ec/projects/gnu.org/gcc/package.yml#L77-L85

jhheider:
once https://github.com/pkgxdev/brewkit/issues/270 is merged, this works for me.

jhheider:
i think GHA is caching `pkgxdev/brewkit/id@v1`, so i'll retry ^17 tomorrow. for now, this builds the newest ^20, so it's already an improvement.

jhheider:
big kudos to the following for making this now possible:
- @mxcl: https://github.com/pkgxdev/brewkit/pull/265
- @Rajdeep-TG: https://github.com/pkgxdev/brewkit/pull/192 and https://github.com/pkgxdev/brewkit/pull/214

PR #4840, opened 2024-01-14
add 'swift'

krzyzanowskim:
Vendored Swift packages for Darwin and Linux; however, the Linux platform is disabled for now due to platform requirement: ubuntu 22.04. Unless somebody can make it run tests on Ubuntu


krzyzanowskim:
and now I wonder why it works on my machine (macOS 14) but fails to unarchive on the CI

Screenshot 2024-01-14 at 15 49 39

mxcl (approved):
Great job

mxcl:
Ideally we’d avoid resigning this since Apple signed it.

Also: I assume we have license to redistribute, but one day we may find out we didn’t.

krzyzanowskim:
> I assume we have license to redistribute

since it is downloaded by the user from the https://swift.org it's not really distributed by pkgx, or is it?

mxcl:
It is distributed by us. The build script runs on our infra to convert everything into bottles that pkgx then downloads from our hosting.

krzyzanowskim:
right, I assume it all is under Apache 2.0 as it states on https://www.swift.org/download/ let's assume that ;)

krzyzanowskim:
@mxcl I'd like you to weigh in on Linux support. I lack the skills how to make it "per distribution" only for a vendored package

mxcl:
@krzyzanowskim distributions for a specific hardware platform are (almost always) just variations on underlying deps. So if Apple provide a swift for ubuntu where the ubuntu glibc matches our requirements (v23 I think) then it'll be fine. We provide the other dependencies and it should just work at runtime.

Failing that we have to build from source.

Though once we finish packaging glibc (#147) that issue could also be alleviated.

> [!IMPORTANT]
> I say all this but distributions can do bad things with underlying dependencies that cause binaries built for them to
> become extremely specific. Off hand I don't recall ever seeing this, but I'm sure it has happened.

PR #4841, opened 2024-01-14
fix(opensearch): fix loader paths

scottjg:
try appending to library path env vars instead of overriding java.library.path which clobbers other necessary paths

### DIFF
--- a/projects/opensearch.org/package.yml
+++ b/projects/opensearch.org/package.yml
@@ -95,8 +95,12 @@ build:
         zip -r opensearch-knn-{{version}}.0.zip lib/
         {{prefix}}/bin/opensearch-plugin install --batch file:`pwd`/opensearch-knn-{{version}}.0.zip
       working-directory: k-NN
-    - run: echo 'export OPENSEARCH_JAVA_OPTS="-Djava.library.path=$OPENSEARCH_HOME/plugins/opensearch-knn/lib $OPENSEARCH_JAVA_OPTS"' >> opensearch-env
+    - run: echo 'export JAVA_LIBRARY_PATH="$OPENSEARCH_HOME/plugins/opensearch-knn/lib:$JAVA_LIBRARY_PATH"' >> opensearch-env
       working-directory: ${{prefix}}/bin
+      if: darwin
+    - run: echo 'export LD_LIBRARY_PATH="$OPENSEARCH_HOME/plugins/opensearch-knn/lib:$LD_LIBRARY_PATH"' >> opensearch-env
+      working-directory: ${{prefix}}/bin
+      if: linux
 
 provides:
   - bin/opensearch


scottjg:
@jhheider i didn't see your comment before i pushed this up. maybe this could be done with a `runtime.env...` key? it's different per platform and depends on the prefix. not sure if that makes it harder

jhheider:
certainly might. this is worth a try.

jhheider:
here we go: https://github.com/pkgxdev/pantry/actions/runs/7517849640

scottjg:
hm.. no luck. it kinda feels like something wrong with maybe the jvm build? maybe i can try reproduce the loadlibrary issue on my x86 mac in isolation

scottjg:
tried this on a macos 14 x86 machine... worked fine.
```
scottjg@Scotts-MacBook-Pro ~ % cat testload.java              
import com.sun.jna.Library;
import com.sun.jna.Native;
import com.sun.jna.Platform;

public class testload {
  public interface CLibrary extends Library {
    CLibrary INSTANCE = (CLibrary)
      Native.load((Platform.isWindows() ? "msvcrt" : "c"),
                  CLibrary.class);
      void printf(String format, Object... args);
  }

  public static void main(String[] args) {
    CLibrary.INSTANCE.printf("Hello, World\n");
  }
}
scottjg@Scotts-MacBook-Pro ~ % pkgx +openjdk.org@17 javac -cp ./jna.jar testload.java
scottjg@Scotts-MacBook-Pro ~ % JAVA_LIBRARY_PATH="test" /Users/scottjg/.pkgx/openjdk.org/v17.0.10.6/bin/java -cp ./jna.jar:. testload
Hello, World
```

trying to get a macos 12 installer to do a VM with that version.

scottjg:
test program reproduces on x86-64 macos 12 with openjdk.org@17 and openjdk.org@20:
Screenshot 2024-01-14 at 2 31 07 PM

scottjg:
on macos it looks like there is no libc. libsystem_* exports some of the same stuff libc would normally provide.
Screenshot 2024-01-14 at 2 40 28 PM

i guess depending on how you look at it, this seems like it might be a legitimate bug in either the linker/loader in macos or in opensearch. 
  - if you blame macos, the same binaries run fine on later versions, so they must've put a workaround in the loader to handle the case where dlopen libc will load the libs out of libsystem? or at least something changed about the loader behavior.
  - if you blame opensearch, maybe they should be conditionally loading the symbols from libsystem vs libc depending on the platform?

either way since this seems to be an upstream bug that affects the package on x86_64 macos12, would it be possible to put a conditional in the yml to say only macos >= 13 is supported? not sure if that is something you folks allow?

jhheider:
We don't have a method for versions, but we could limit it to `platforms: [linux, darwin/aarch64]` with a comment if we can't figure out a fix. That's going to break everything that's already working, however.

Perhaps it needs to explicitly `-lSystem` in the `LDFLAGS`? Or we might be grabbing the wrong jdk?

scottjg:
hmm...

> Perhaps it needs to explicitly -lSystem in the LDFLAGS? 

in this case, opensearch is explicitly loading libc at runtime to call some specific functions from java, so i'm not sure that it's a compile-time linking issue. also, when i run `otool -L java` on the binary, it says it's linked against libSystem. 

> Or we might be grabbing the wrong jdk?

possible i messed something up, but at least i found that both versions of the jdk on macos12 suffer the same issue.

i was flipping through the opensearch source, and i saw this:
https://github.com/opensearch-project/OpenSearch/blob/988dea88a0aaae192db602f35f5cd6d714d07fce/server/src/main/java/org/opensearch/bootstrap/Natives.java#L55-L65
```java
    static {
        boolean v = false;
        try {
            // load one of the main JNA classes to see if the classes are available. this does not ensure that all native
            // libraries are available, only the ones necessary by JNA to function
            Class.forName("com.sun.jna.Native");
            v = true;
        } catch (ClassNotFoundException e) {
            logger.warn("JNA not found. native methods will be disabled.", e);
        } catch (UnsatisfiedLinkError e) {
            logger.warn("unable to load JNA native support library, native methods will be disabled.", e);
        }
```
it seems like maybe just deleting the `jna-*.jar` library on affected versions of macos might disable the behavior. is there any kind of install-time steps in the manifest? if not i guess we can just delete it in the package for macos x86 and just remove that later when macos 12 becomes unsupported.

scottjg:
https://github.com/pkgxdev/pantry/pull/4847

PR #4847, opened 2024-01-15
fix(opensearch): don't include the jna library on mac for x86-64

scottjg:
there seems to be some linker issue with loading libc. calling these native functions doesn't seem strictly necessary. maybe better that the package works at all.

in macos 13 and later, it works fine, so maybe this workaround could be removed when macos 12 is no longer supported

### DIFF
--- a/projects/opensearch.org/package.yml
+++ b/projects/opensearch.org/package.yml
@@ -24,6 +24,11 @@ build:
       openblas.net: "*"
 
   script:
+      # there seems to be a loader bug with trying to dlopen libc on macos 12 and earlier.
+      # remove this step when macos 12 and under are no longer supported.
+    - run: |
+        sed -i 's|JNAKernel32Library.getInstance();|//JNAKernel32Library.getInstance();|' server/src/main/java/org/opensearch/bootstrap/Bootstrap.java
+      if: darwin/x86-64
     - gradle -Dbuild.snapshot=false ":distribution:archives:no-jdk-{{hw.platform}}-tar:assemble"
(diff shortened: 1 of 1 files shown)


jhheider:
Good sleuthing. If this fixes it, I'm fine with the comment.

scottjg:
hmm.. no that doesn't seem to work

scottjg:
x86 mac build seems to work. you can see the changes took effect in the test log:
```
[2024-01-14T19:40:30,336][WARN ][o.o.b.Natives            ] [tea-darwin-x86-64-1.local] cannot check if running as root because JNA is not available
[2024-01-14T19:40:30,336][WARN ][o.o.b.Natives            ] [tea-darwin-x86-64-1.local] cannot install system call filter because JNA is not available
[2024-01-14T19:40:30,337][WARN ][o.o.b.Natives            ] [tea-darwin-x86-64-1.local] cannot register console handler because JNA is not available
[2024-01-14T19:40:30,337][WARN ][o.o.b.Natives            ] [tea-darwin-x86-64-1.local] cannot getrlimit RLIMIT_NPROC because JNA is not available
[2024-01-14T19:40:30,337][WARN ][o.o.b.Natives            ] [tea-darwin-x86-64-1.local] cannot getrlimit RLIMIT_AS because JNA is not available
[2024-01-14T19:40:30,337][WARN ][o.o.b.Natives            ] [tea-darwin-x86-64-1.local] cannot getrlimit RLIMIT_FSIZE because JNA is not available
```

scottjg:
hmm... looking at this a bit more, there might be some more changes needed to fix this on macos12....

scottjg:
scratch that. i think we're good

jhheider:
looks like a success! https://github.com/pkgxdev/pantry/actions/runs/7523480856/job/20477471616

scottjg:
phew. finally!

PR #4887, opened 2024-01-18
+overmind

scottjg:
adds the overmind package

### DIFF
--- a/projects/github.com/darthsim/overmind/package.yml
+++ b/projects/github.com/darthsim/overmind/package.yml
@@ -9,7 +9,7 @@ versions:
 
 platforms:
   - darwin
-  - linux/x86-64
+  - linux
 
 dependencies:
   github.com/tmux/tmux: "*"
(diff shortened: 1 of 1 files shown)


projects/github.com/darthsim/overmind/package.yml
```diff
@@ -0,0 +1,37 @@
+distributable:
+  url: https://github.com/DarthSim/overmind/archive/refs/tags/v{{version}}.tar.gz
+  strip-components: 1
+
+display-name: overmind
+
+versions:
+  github: darthsim/overmind
+
+platforms:
```
jhheider: if it doesn't work on `linux/aarch64`, please put in a comment as to why. usually `go` bin are pretty easy to build.

jhheider (changes requested):

projects/github.com/darthsim/overmind/package.yml
scottjg: oops, not intentional. had just copied from another formula

PR #4891, opened 2024-01-18
+z

mxcl:


SushyDev:
pkgx z does not work

running `sh <(curl https://pkgx.sh) +github.com/rupa/z sh` also does not seem to add the `z` command to the shell

jhheider:
`sh <(curl https://pkgx.sh) +github.com/rupa/z sh` just starts a shell. you'd want:

`sh <(curl https://pkgx.sh) +github.com/rupa/z source "${PKGX_DIR:-$HOME/.pkgx}/github.com/rupa/zv*/z.sh"` to load it that way, i believe. though, most likely, that wouldn't load it in your particular session.

```shell
$ sh <(curl https://pkgx.sh) +github.com/rupa/z
$ source "${PKGX_DIR:-$HOME/.pkgx}/github.com/rupa/zv*/z.sh"
```
is probably where you need to be.

PR #4895, opened 2024-01-18
Try to do swift linux

mxcl:
refs #4840

### DIFF
--- a/projects/swift.org/package.yml
+++ b/projects/swift.org/package.yml
@@ -11,14 +11,10 @@ versions:
 
 platforms:
   - darwin/aarch64
-# It is ready to run on linux. It going to work with ubuntu 22.04,
-# and not with debian-buster and tests going to fail
-#  - linux/x86-64
-#  - linux/aarch64
+  - linux
 
 dependencies:
   linux:
-    gnu.org/gcc: '*'
     gnu.org/binutils: '*'
     gnupg.org: ^2
     gnome.org/libxml2: '*'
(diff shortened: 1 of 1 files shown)


mxcl:
needs `libtinfo` which is an optional library built by ncurses.

Can probs fix with a few things, let me play around

mxcl:
k locally I have demonstrated that swift on linux only needs libncurses5 to work. It also requires that it be built the same as ubuntu, thus with libtinfo separate (usually it is built into the same `.so`)

We have no ncurses 5 at this time, I will try to build one. The `libtinfo` issue is separate. I don't think it's enough to build v5, but I'll try that first.

mxcl:
Alternatively we can just build this ourselves. I tried a couple of times and it is quite a bitch. Same as emscripten.org in that it requires a very specific LLVM toolchain to build.

mxcl:
Our delivery of this package may prove kinda problematic on mac since it depends on specific Xcode toolchain versions (xcode command line tools) and we have no control over that.

I don't think our LLVM is quite right enough for swift.

This is all unfortunate ofc. We may be better off delegating to swiftly in that respect, which we still have not built.

Ultimately one blames Apple. If OSS people without pay can make their software more portable than Apple, one of the richest companies in the world, then WTF is going on?

krzyzanowskim:
is `linux/aarch64` no go? because you know, handy in VM on M1

mxcl:
> is `linux/aarch64` no go? because you know, handy in VM on M1

yeah we want it. Unfortunately we'll need to build it from source ourselves. I'll get to it eventually. I tried once and it's tricky because Apple.

krzyzanowskim:
> Unfortunately we'll need to build it from source ourselves.

I don't understand. why so? I tried in docker and it seems just like x86_64, isn't it?

PR #4901, opened 2024-01-19
fix(postgresql.org): install contrib extensions

scottjg:
in the homebrew formula, postgres does `make install-world`, not just `make install`, which installs extensions under the `contrib/` source dir.

in my case, i need the pg_trgm (trigram index) extension


jhheider:
Ooh, good catch.

PR #4908, opened 2024-01-19
+dest-unreach.org/socat

and-ri:


and-ri:
hmmmm

jhheider:
maybe `<<< 'GET /\n'`?

and-ri:
> maybe `<<< 'GET /\n'`?

I will try, but it's strange that it doesn't work on Linux ARM only

and-ri:
perfect, thank you :)

PR #4925, opened 2024-01-21
+glfw.org

and-ri:


and-ri:
I was hoping it would work on Linux. But it's the same in the hombrew formula

```bash
return if OS.linux? && ENV["HOMEBREW_GITHUB_ACTIONS"]
system "./test"
```

jhheider:
hm. maybe `-fPIC -pie`?

and-ri:
no 😞

PR #4931, opened 2024-01-21
+ktlint.github.io

and-ri:


projects/ktlint.github.io/package.yml
```diff
@@ -0,0 +1,24 @@
...
+
+warnings:
+  - vendored
+
+dependencies:
+  openjdk.org: ^20
+
+build:
+  - mkdir -p {{prefix}}/bin
+  - cp ktlint-{{version}}/bin/ktlint {{prefix}}/bin/
```
mxcl, line 15:
```
install -D ktlint-{{version}}/bin/ktlint {{prefix}}/bin/ktlint
```

is a one-liner to do this ffr

PR #4935, opened 2024-01-22
+defenseunicorns.com/zarf

Noxsios:
<https://github.com/defenseunicorns/zarf>

Zarf is a statically compiled Go binary for packaging K8s deployments for the air gap.

I'm doing some exploration into using `pkgx` within Zarf packages as an integrated package manager, in the meantime thought I might add Zarf to `pkgx` as an alternative to Brew for install.


jhheider:
@Noxsios is it not possibly to compile this from source? If not, the `distributable:` stanza should be removed and a `warnings: [vendored]` added.

Noxsios:
@jhheider Most certainly, I was unsure what the team's stance on precompiled vs compile from source was. Since Zarf is a static binary there is little need to compile locally from source because it has zero outside dependencies.

If compile from source is the preferred build strat, please let me know and I will fix the build step.

jhheider:
compiled by source is preferred, followed by signed binary, checksummed binary (as you've done), and non-validated binary.

jhheider:
~`mkdir -p '{{prefix}}'/bin`~

no, hm. you got that. possibly the built artifact path is different?

jhheider:
ah, platform-specific names:
`-o build/zarf-mac-apple`

Noxsios:
mb, will fix

jhheider:
thanks, @Noxsios !

Noxsios:
@jhheider Thanks for the rapid responses!

jhheider:
no problem. contributions are _always_ welcome, and we'll try to help anytime we're awake.

PR #4943, opened 2024-01-22
+github.com/bats-core/bats-core

and-ri:


and-ri:
hmm `bc: command not found`
Is it necessary to add the `gnu.org/bc` command?
it is strange that it is not in Linux

jhheider:
if `bats` needs it, then it looks like yes. should be trivial, i assume.

jhheider:
oops, missed that it was a dep in the test. could have just used `$[]` or `let` probably to remove the extra dep. but we've got it now.

and-ri:
> oops, missed that it was a dep in the test. could have just used `$[]` or `let` probably to remove the extra dep. but we've got it now.

you are right, I missed that too 😃

PR #4963, opened 2024-01-23
+openapi-generator.tech

and-ri:


projects/openapi-generator.tech/package.yml
```diff
@@ -0,0 +1,59 @@
...
+  script:
+    - mvn clean package -Dmaven.javadoc.skip=true
+    - install -D modules/openapi-generator-cli/target/openapi-generator-cli.jar {{prefix}}/libexec/lib/openapi-generator-cli.jar
+    - run: |
+        cat > openapi-generator <<EOF
+        #!/bin/sh
+        exec java -jar \$(dirname \$0)/../libexec/lib/openapi-generator-cli.jar "\$@"
+        EOF
+        chmod +x openapi-generator
+      working-directory: ${{prefix}}/bin
```
jhheider, line 23: this can also be:
```yaml
  - run: install $PROP openapi-generator
    prop: |
      #!/bin/sh
      exec java -jar $(dirname $0)/../libexec/lib/openapi-generator-cli.jar "$@"
    working-directory: ${{prefix}}/bin
```
i believe, if you find that simpler. (double check the output, though. i think we removed the need to escape `$`, but i'm not _positive_.)

projects/openapi-generator.tech/package.yml
```diff
@@ -0,0 +1,59 @@
...
+            '200':
+              description: A JSON array of user names
+              content:
+                application/json:
+                  schema:
+                    type: array
+                    items:
+                      type: string
+  script:
+    - cp $FIXTURE minimal.yaml
```
jhheider, line 54: and for this one, i think fixture can be:
```yaml
  fixture:
    contents: | ...
    extname: yaml
```
which would let you skip the copy and just use `openapi-generator generate -i $FIXTURE`. again, test to perfect syntax, but all of those are added to improve usability.

PR #4970, opened 2024-01-24
fix(jenkins)

### LINKED ISSUE #4969: ❌ build issues: jenkins.io=2.426.3
Running log of build failure for jenkins.io=2.426.3

jhheider:
closes #4969

### DIFF
--- a/projects/jenkins.io/package.yml
+++ b/projects/jenkins.io/package.yml
@@ -44,12 +44,9 @@ provides:
 test:
   dependencies:
     curl.se: '*'
+    pkgx.sh: ^1
   script:
-    - run: |
-        FREE_PORT=$((RANDOM % 10000 + 1024))
-        while lsof -i:$FREE_PORT >/dev/null 2>&1; do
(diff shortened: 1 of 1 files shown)


projects/jenkins.io/package.yml
```diff
@@ -46,14 +46,17 @@ test:
     curl.se: '*'
   script:
     - run: |
-        for i in $(seq 1 10); do
-          FREE_PORT=$((RANDOM % 10000 + 1024))
-          if ! nc -z localhost $FREE_PORT; then
-            echo $FREE_PORT
-          fi
+        FREE_PORT=$((RANDOM % 10000 + 1024))
```
mxcl: there's an `npx` utility that can do this. I remember not what it is called but we use it in the sd-webui endpoint script.

jhheider: `get-port`. not sd-webui, but others. gotta bring in the node ecosystem then, of course.

PR #4983, opened 2024-01-25
Remove openjdk dependency from ktlint

jakobvase:
I'm not certain it's the right solution, but for earlier versions of ktlint at least, I'm pretty sure it depends on earlier versions of java.

Additionally, you could also use another jdk like temurin or zulu, so I'm leaning towards not specifying the dependency at all?

### DIFF
--- a/projects/ktlint.github.io/package.yml
+++ b/projects/ktlint.github.io/package.yml
@@ -6,13 +6,12 @@ versions:
 
 warnings:
   - vendored
-
-dependencies:
-  openjdk.org: ^20
-
+  
 build:
-  - mkdir -p {{prefix}}/bin
-  - cp ktlint-{{version}}/bin/ktlint {{prefix}}/bin/
+  dependencies:
+    openjdk.org: ^20
+  script:
+    - install -D ktlint-{{version}}/bin/ktlint {{prefix}}/bin/ktlint
 
 provides:
   - bin/ktlint
--- a/projects/ktlint.github.io/package.yml
+++ b/projects/ktlint.github.io/package.yml
@@ -9,7 +9,7 @@ warnings:
 
 dependencies:
   openjdk.org: '*'
-  
+
 build:
   - install -D ktlint-{{version}}/bin/ktlint {{prefix}}/bin/ktlint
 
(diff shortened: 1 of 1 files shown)


projects/ktlint.github.io/package.yml
```diff
@@ -1,18 +1,14 @@
 distributable:
-  url: https://github.com/pinterest/ktlint/releases/download/{{version}}/ktlint-{{version}}.zip
+  url: https://github.com/pinterest/ktlint/releases/download/{{version}}/ktlint
```
jhheider: if we're going to use the raw file, it can't be via a distributable URL; you'll need to remove this whole stanza, include `curl.se` as a build-dep, and construct and grab the URL in the build script.

projects/ktlint.github.io/package.yml
```diff
@@ -1,18 +1,14 @@
...
 
 versions:
   github: pinterest/ktlint
 
 warnings:
   - vendored
-
-dependencies:
-  openjdk.org: ^20
```
jhheider: given that this is the only jdk we have (and we don't have `||` dependencies), won't this break the package? or does it not require a jdk at all?

jhheider (changes requested):
distributable URL won't accept a non-archive.

projects/ktlint.github.io/package.yml
jakobvase: Thanks for taking a look at this. It does require a jdk to run, but Java is complicated, and if it does specify the dependency, I don’t think the user will be able to use another jdk (installed through brew or some Java version manager).

In any case, it should at least be changed to ‘*’, as they handle all the Java versions in their executable. I’ll change this mr to that tomorrow, and maybe take a look at how to handle the jdk mess afterwards.

jhheider: You'd likely have to pick a different jdk via careful construction of the path.

mxcl: we can handle complicated we just need to know details, eg. is this a runtime or link time dep? Best open a discussion and we can figure out the best path forwards

projects/ktlint.github.io/package.yml
jakobvase: I'll just leave it as a zip then. Probably the people depending on versions before 0.49.1 are few.

jhheider: zero people would be ideal, but i'm not naive enough to think it. ;)

projects/ktlint.github.io/package.yml
jakobvase: https://github.com/orgs/pkgxdev/discussions/940 Discussion :)

PR #4991, opened 2024-01-25
Consolidate on unicode.org^73

mxcl:
Consolidating on a single unicode.org version is a temporary measure at best. Longer term we need to be able to vary dependencies based on version somehow.

### DIFF
--- a/projects/dotnet.microsoft.com/package.yml
+++ b/projects/dotnet.microsoft.com/package.yml
@@ -14,7 +14,7 @@ runtime:
     DOTNET_ROOT: "{{prefix}}"
 dependencies:
   linux:
-    unicode.org: ^71
+    unicode.org: ^73
 build:
   dependencies:
     curl.se: '*'
--- a/projects/harfbuzz.org/package.yml
+++ b/projects/harfbuzz.org/package.yml
@@ -10,7 +10,7 @@ dependencies:
   freetype.org: 2
   gnome.org/glib: 2
   graphite.sil.org: '*'
-  unicode.org: ^71
+  unicode.org: ^73
 
 build:
   dependencies:
--- a/projects/mysql.com/package.yml
+++ b/projects/mysql.com/package.yml
@@ -7,7 +7,7 @@ versions:
   strip: /^mysql-/
 
 dependencies:
-  unicode.org: ^71
+  unicode.org: ^73
   libevent.org: ^2
   lz4.org: ^1
   openssl.org: ^1.1
--- a/projects/nodejs.org/package.yml
+++ b/projects/nodejs.org/package.yml
@@ -10,7 +10,7 @@ companions:
 
 dependencies:
   # prior builds of nodejs used our icu4c
-  unicode.org: ^71
+  unicode.org: ^73
   openssl.org: 1.1
   zlib.net: 1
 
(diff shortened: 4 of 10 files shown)


mxcl:
Like, this means we have to restock everything, and that may not succeed. This is a little problematic.

Note: failures are unrelated afaict.

Other note: I’m somewhat confused as to how these all passed and makes me wonder what breaking changes actually happen between unicode major versions. May just be C++ ABI shit ofc.

jhheider:
some of these packages haven't had a new build across one or more breaking brewkit changes. i _suspect_ unicode's libs are forward-compatible, since they're almost certainly just adding glyph definitions (but don't check if i'm right).

jhheider:
and, yes, restocking `node` will be the worst thing ever. followed by php and mysql.

mxcl:
> i _suspect_ unicode's libs are forward-compatible, since they're almost certainly just adding glyph definitions (but don't check if i'm right).

If true we would not expect `unicode.org: *` to be a problem, but it has been (specifically link problems when trying to instantiate programs that linked to unicode*).

OTOH as I said could just be C++ ABI shit. You can change very little in C++ and the ABI breaks.

ie. API is forwards compatible but the ABI is not

jhheider:
it's because of the path. we reduce linkage paths to `$PKGX/$pkg/v$major/lib/libfoo.dylib`, so if they pull `v73` but they're looking in `/v71/lib` they miss it.

mxcl:
> it's because of the path. we reduce linkage paths to $PKGX/$pkg/v$major/lib/libfoo.dylib, so if they pull v73 but they're looking in /v71/lib they miss it.

Certainly it could be that. But we set `LD_LIBRARY_PATH` which should be used in preference. So it would still end using v71 in that case

jhheider:
as i recall, that is the issue. it was the rpaths.

interestingly, surveying things that dep unicode, I can't find any unicode rpaths. so, it might be compiled in dynload path? i don't recall, but that was certainly the issue that caused us to lock versions to begin with, as i remember it. if changes to fix-machos/fix-elf since have fixed us, maybe we're good.

mxcl:
It's not rpaths:

```
$ pkgx node@19 --version
 ✓  ~/.pkgx/unicode.org/v73.2.0
dyld[18029]: Symbol not found: __ZNK6icu_716number23NumberFormatterSettingsINS0_24LocalizedNumberFormatterEE10toSkeletonER10UErrorCode
  Referenced from: <6C5723F6-2A2A-3696-BA26-D50B36C51A2E> /Users/mxcl/.pkgx/nodejs.org/v19.9.0/bin/node
  Expected in:     <FF276526-D48A-39A8-9ED8-36CF8F910ADA> /Users/mxcl/.pkgx/unicode.org/v73.2.0/lib/libicui18n.73.2.dylib
```

This is running node^19 against unicode^73 where node^19 was built against unicode^71

Edit: rebuilding node against ^73 fixes this.

Basically if everything will work with unicode 73 if we rebuild it all then this will do as a solution for now. But I will need to devise something more clever to solve the cannot intersect ^71 && ^73 stuff. †

The cleverer solution is allowing multiple conflicting versions of things in the same env. I have thought a lot about it and basically: it's mostly fine. We can start by allowing this on a case by case basis, eg. unicode and python.

> † and what we need for that is some kind of atomic all builds are ready to be uploaded when this PR is merged type system. Though at this early stage in pkgx’s life we can probs just merge and rebuild everything. Node, as you say, is the real bitch here. That will take days just by itself sigh.
> 
> **So I guess realistically we should not merge this PR**

jhheider:
no longer needed?

PR #4992, opened 2024-01-25
+geoff.greer.fm/ag

and-ri:


jhheider:
i went to package silver_searcher ages ago, but the last commit is 4 years ago and the last merged PR over 3 years ago. i assume pretty much all the heat has moved to `ripgrep`, no?

if it builds and works, no reason not to include it, but i figured it's pretty much dead.

and-ri:
> i went to package silver_searcher ages ago, but the last commit is 4 years ago and the last merged PR over 3 years ago. i assume pretty much all the heat has moved to `ripgrep`, no?
> 
> if it builds and works, no reason not to include it, but i figured it's pretty much dead.

Yes you are right. But it has 53,253 installs in the last 90 days on homebrew. I thought he could be considered popular

jhheider:
probably true. i was still using it until 18 months ago. go for it.

PR #4993, opened 2024-01-25
+cryptography.io

and-ri:


and-ri:
@jhheider  can you please help with the test section? I tried to do as you wrote earlier (here https://github.com/pkgxdev/pantry/pull/4963). But it seems I'm doing something wrong

jhheider:
looks like it's working: https://github.com/pkgxdev/pantry/actions/runs/7659720377/job/20875465999?pr=4993

and-ri:
> looks like it's working: https://github.com/pkgxdev/pantry/actions/runs/7659720377/job/20875465999?pr=4993

Yes,  I found an example here https://github.com/pkgxdev/pantry/blob/main/projects/ruby-lang.org/package.yml

PR #5015, opened 2024-01-28
fix(mc)

### LINKED ISSUE #5008: ❌ build issues: midnight-commander.org=4.8.31
Running log of build failure for midnight-commander.org=4.8.31

jhheider:
closes #5008

### DIFF
--- a/projects/midnight-commander.org/package.yml
+++ b/projects/midnight-commander.org/package.yml
@@ -51,5 +51,10 @@ build:
       - --with-screen=ncurses
 
 # FIXME: hard to test interactive tools
-test: test "$(mc --datadir | cut -d ' ' -f1 | cut -d':' -f1)" = "{{prefix}}/etc/mc/"
+test:
+  env:
+    TERM: vt100  # needs $TERM set to run
+  script:
+    - mc --datadir
+    - test "$(mc --datadir | cut -d ' ' -f1 | cut -d':' -f1)" = "{{prefix}}/etc/mc/"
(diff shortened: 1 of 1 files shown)


projects/midnight-commander.org/package.yml
```diff
@@ -51,5 +51,10 @@ build:
       - --with-screen=ncurses
 
 # FIXME: hard to test interactive tools
-test: test "$(mc --datadir | cut -d ' ' -f1 | cut -d':' -f1)" = "{{prefix}}/etc/mc/"
+test:
+  env:
+    TERM: vt100  # needs $TERM set to run
```
mxcl, line 56: Question is, should brewkit set this? Possibly should be set to `dumb`?

jhheider: I had a vague memory that things didn't like `dumb`. At the present time, this hack is only in two packages; that said, if this is a recent change in brewkit, more `curses` programs may display this behavior.

mxcl: At v1 I began sanitizing build/test env so that it is about as bare as possible.

`TERM` might be something we should let through or set, but I'm not really sure about what it means.

jhheider: likely should not sanitize:

https://www.baeldung.com/linux/term-environment-variable

jhheider: i would include `LC_*` vars as things we shouldn't modify.

PR #5029, opened 2024-01-28
+btop

mxcl:


mxcl:
oddly this works fine for me on macOS locally. Worrying divergence of brewkits

mxcl:
linux can't find the C++20 headers… :-/

IIRC llvm doesn't provide these so seems like a pre-requisites issue.

Eventually surely we need to provide these tho…?

jhheider:
maybe gnu.org/gcc?

jhheider:
the darwin ones are odd; the linux one can be fixed by making `gnu.org/gcc` a runtime dep (for libstdc++`.

jhheider:
as if i didn't consider c++'s 8 generations of variable syntax already a nightmare, everything about building it feels like the Apollo VII launch. everyone should be using go or rust. _everyone_.

jhheider:
stackoverflow says include `-ldl` to fix missing `dlopen`.

mxcl:
> as if i didn't consider c++'s 8 generations of variable syntax already a nightmare, everything about building it feels like the Apollo VII launch. everyone should be using go or rust. _everyone_.

c++ really never learned from the mistake of letting other people decide on how builds, linking and build configuration should work out.

mxcl:
hmm is it possible the two apple silicon builders have different macOS or Xcode versions installed? The build worked last run and failed with “clang too old” this time

jhheider:
Definitely possible. I can update both.

mxcl:
> Definitely possible. I can update both.

seems risky to update both, but we should try to ensure they are at least the same for macOS & Xcode.

jhheider:
the older one was on ventura and 14.3; i'll get them both on sonoma and 15.

jhheider:
but i agree about risk; if we only ever leap-frog the older over the newer it'll help catch more bugs

PR #5058, opened 2024-01-30
fix(cocoapods)

### LINKED ISSUE #5022: ❌ build issues: cocoapods.org=1.15.0
Running log of build failure for cocoapods.org=1.15.0

jhheider:
closes #5022

### DIFF
--- a/projects/cocoapods.org/package.yml
+++ b/projects/cocoapods.org/package.yml
@@ -18,6 +18,9 @@ build:
 
     # FIXME ruby’s mkmf needs patching to check CPATH
     CFLAGS: -I{{deps.sourceware.org/libffi.prefix}}/include
+    linux/aarch64:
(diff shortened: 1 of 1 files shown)


projects/cocoapods.org/package.yml
```diff
@@ -34,4 +37,6 @@ test:
   env:
     # some of our test containers are raw and thus run as root
     COCOAPODS_ALLOW_ROOT: 1
+    linux:
+      LANG: en_US.UTF-8
```
mxcl, line 41: This is surfaced from the underlying environment. If it is not set on Linux then the builders should be modified. Since I don’t like relying on the underlying env, brewkit should set it.

jhheider:
Not the fix, fortunately, so it won't be in the final solution.

jhheider:
closed in favor of a4b56ebfdbf1d7cab9cd922b816b9287e5d486dc

PR #5060, opened 2024-01-31
Add tf-import-gen command

kishaningithub:
This pull request adds [tf-import-gen command](https://github.com/kishaningithub/tf-import-gen) to the pantry

### DIFF
--- a/projects/github.com/kishaningithub/tf-import-gen/package.yml
+++ b/projects/github.com/kishaningithub/tf-import-gen/package.yml
@@ -18,9 +18,8 @@ build:
     LDFLAGS:
       - -s
       - -w
-      - -X main.version={{version}}
+      - -X main.Version={{version}}
 
 test:
-  - tf-import-gen --version 2>&1 || true # test for exit code
   - tf-import-gen --version
   - tf-import-gen --version 2>&1 | grep -F "{{version}}"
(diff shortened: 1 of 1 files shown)


kishaningithub:
Need some help to understand what exactly is failing here (link and screenshot below)

## Link
https://github.com/pkgxdev/pantry/actions/runs/7722074891/job/21049648385?pr=5060

## Screenshot
image

jhheider:
probably because the grep is failing. adding non-piped variant to see.

jhheider:
```
+ cd /Users/builder/actions-runner/_work/pantry/pantry/testbeds/github.com__kishaningithub__tf-import-gen-0.4.1
+ tf-import-gen --version
tf-import-gen version dev
+ tf-import-gen --version
tf-import-gen version dev
+ tf-import-gen --version
+ grep -F 0.4.1
```
so, the correct variable for versioning isn't being set. it's not `main.Version`. we'll have to check the source.

jhheider:
ah, looking at the [goreleaser.yml](https://github.com/kishaningithub/tf-import-gen/blob/main/.goreleaser.yml), it *is* `main.Version`, not `main.version`.

jhheider:
thanks, @kishaningithub !

kishaningithub:
Thanks for the merge! ☺️ Is it possible to make the error messages a bit more descriptive?

jhheider:
Interesting question. The last test failed, which was the grep. pkgx doesn't know _why_ it failed, just its exit code. Hence my first test was to run the same command without grep.

@mxcl any ideas on exposing these kinds of tests to greater introspectability?

mxcl:
The error message could say “your test failed” but there is nothing more we can do since it's a shell script and it errors how it errors. We output everything it outputs.

PR #5067, opened 2024-01-31
use new brewkit for gcc ar/ranlib fix

mxcl:


jhheider:
open-mpi takes _just a few hours_ to build.

mxcl:
I'm just gunna assume it works.

PR #5080, opened 2024-02-01
New/gnu.org/glibc

mxcl:
My attempt at #147

### DIFF
--- a/projects/gnu.org/glibc/package.yml
+++ b/projects/gnu.org/glibc/package.yml
@@ -71,9 +71,6 @@ test:
     linux/aarch64:
       ARCH: aarch64
   script:
-    # Putting ourselves in the LD_LIBRARY_PATH breaks literally everything else...
-    - export LD_LIBRARY_PATH="$(echo LD_LIBRARY_PATH | tr ':' '\n' | grep -v {{prefix}} | tr '\n' ':')"
-
     - gcc -o test1 test.c -fPIC -pie
     - ./test1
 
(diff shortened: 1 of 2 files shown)


mxcl:
This fails, but I'm hoping dosu bot has a clue

mxcl:
oh ok, it didn't build successfully for me on my local machine. I'll have to look into that… could be the pkgx docker image doesn't have the pre-reqs installed? That would be an oopsie.

Still, nice! It built. Tests fail. Which begs the question: how do you test this?

jhheider:
> how do you test this?

right. i think the answer is something like this: https://github.com/pkgxdev/pantry/blob/7976e914b9a90a5cee5aa9edcba86bff69426b52/projects/gnu.org/glibc/package.yml#L68-L100

mxcl:
k the local build issues are due to us building into macOS mounted volume and the case insensitive filesystem.

Ref https://github.com/openrisc/or1k-glibc/issues/10

Which is annoying. What we need I guess is the inverse: everything is in a persistent linux docker volume and somehow we make it mounted locally so you can easily browse it.

Or something

mxcl:
Running in a local docker container I can get it all built, actually using it is problematic.

The resulting binary we create segfaults in some of the glibc runtime code that starts the executable. Tracing it with ChatGPT I ended up trying to run `bin/ld.so` manually, which also segfaults. Not sure if that is the key here or not.

PR #5081, opened 2024-02-01
ci/plan is less wasteful and more legible

mxcl:


mxcl:
ugh this results in 2 “plan steps” to save spinning up instances that will immediately cancel for pkgs that skip platforms.

So considering most pkgs are all platforms this reduces readability for the general case.

I can consolidate the plan steps but it's more work

PR #5103, opened 2024-02-02
upload file listing

mxcl:


.github/workflows/pkg-platform.yml
```diff
@@ -135,6 +135,13 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: us-east-1
 
+      - name: mock `aws`
+        if: inputs.dry-run
+        run: |
+          echo '#!/bin/sh' > /usr/local/bin/aws
+          echo 'echo "$@"' >> /usr/local/bin/aws
+          chmod +x /usr/local/bin/aws
+
```
jhheider: won't have to worry about path issues (or polluting our builders!!!) if you just set `AWS` in github_env, and replace lower calls with `$AWS`.

.github/workflows/pkg-platform.yml
```diff
@@ -125,6 +125,7 @@ jobs:
         compression: [xz, gz]
     runs-on: ubuntu-latest
     env:
+      AWS: ${{ inputs.dry-run && 'aws' || 'echo '}}
```
jhheider: ooh, very slick.

jhheider: if it had worked. hm. i wonder if it's needs the `== true`. might explain why the other `input.dry-run &&` things look like they're not working.

.github/workflows/pkg-platform.yml
mxcl: notably this only ran on `ubuntu-latest` so was probs safe

.github/workflows/pkg-platform.yml
jhheider: oh. ha. just logic reversed. i'll fix it.

PR #5112, opened 2024-02-04
+practical-scheme.net/gauche

and-ri:


projects/practical-scheme.net/gauche/package.yml
```diff
@@ -0,0 +1,57 @@
...
+  env:
+    GAUCHE_LOAD_PATH: ${{prefix}}/etc/GAUCHE_LOAD_PATH:{{prefix}}/etc/GAUCHE_LOAD_PATH/gauche:$GAUCHE_LOAD_PATH
+    GAUCHE_DYNLOAD_PATH: ${{prefix}}/etc/GAUCHE_DYNLOAD_PATH:$GAUCHE_DYNLOAD_PATH
+
+build:
+  script:
+    - ./configure $ARGS
+    - make --jobs {{hw.concurrency}}
+    - make --jobs {{hw.concurrency}} install
+    - run: sed -i 's|#!{{prefix}}/bin/gosh|#!/usr/bin/env gosh|g' gauche-cesconv gauche-install gauche-package
```
jhheider: should be able to just run `fix-shebangs.ts gauche-cesconv gauche-install gauche-package` which currently does just that.

projects/practical-scheme.net/gauche/package.yml
```diff
@@ -0,0 +1,57 @@
...
+    - run: GAUCHE_DIR=$(ls -d */ | grep gauche)
+      working-directory: ${{prefix}}/lib
+    - run: ln -s ../share/${GAUCHE_DIR}/{{version}}/lib GAUCHE_LOAD_PATH
+      working-directory: ${{prefix}}/etc
+
+    # create symlink to lib/gauche-0.98/0.9.14/aarch64-apple-darwin23.2.0
+    - run: GAUCHE_PLATFORM_DIR=$(ls -d */ | grep -i $(uname -s))
+      working-directory: ${{prefix}}/lib/${GAUCHE_DIR}/{{version}}
+    - run: ln -s ../lib/${GAUCHE_DIR}/{{version}}/${GAUCHE_PLATFORM_DIR} GAUCHE_DYNLOAD_PATH
+      working-directory: ${{prefix}}/etc
```
jhheider: this seems like it should be under `lib`, not `etc`, right? these are loadable libraries?

and-ri: > this seems like it should be under `lib`, not `etc`, right? these are loadable libraries?

Yes, you are absolutely right. It will be correct to place `GAUCHE_LOAD_PATH` in `share` and `GAUCHE_DYNLOAD_PATH` in `lib`

PR #5127, opened 2024-02-06
+portaudio.com

magnusviri:


magnusviri:
The Linux version depends on jackaudio, which isn't in pkgx, so it's commented out. But I didn't make this macOS only. I guess I should do that.

magnusviri:
Yeah, it's supposed to be include. Whoops! Thanks for catching that.

PR #5128, opened 2024-02-06
github.com/squidfunk/mkdocs-material add common plugins

tannevaled:


jhheider:
thanks, @tannevaled ! i moved it to our (better engineered) `bkpyvenv` solution to future-proof it, and it works great. i'll merge as soon as the last test completes.

PR #5133, opened 2024-02-06
fix(ncurses)

and-ri:
I found this solution [here](https://github.com/Homebrew/homebrew-core/blob/e2b240c98eb652ddde59492a8b622da3db17213a/Formula/n/ncurses.rb#L74C5-L81C8) 

```
    on_linux do
      # libtermcap and libtinfo are provided by ncurses and have the
      # same api. Help some older packages to find these dependencies.
      # https://bugs.centos.org/view.php?id=11423
      # https://bugs.launchpad.net/ubuntu/+source/ncurses/+bug/259139
      lib.install_symlink "libncurses.so" => "libtermcap.so"
      lib.install_symlink "libncurses.so" => "libtinfo.so"
    end
```


jhheider:
yay! /s

PR #5150, opened 2024-02-08
+tuist.io/tuist

dcvz:
Depends on #5148 as its used to create the necessary xcframeworks

### DIFF
--- a/projects/tuist.io/tuist/package.yml
+++ b/projects/tuist.io/tuist/package.yml
@@ -14,6 +14,7 @@ provides:
 build:
   dependencies:
     github.com/unsignedapps/swift-create-xcframework: "*"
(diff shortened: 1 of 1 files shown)


dcvz:
@jhheider do you happen to know what's happening with this pipeline? seems like its a brewkit issue?

jhheider:
It looks like swift version issues. I'll check the builders and see if they need updates.

jhheider:
everything is updated; it looks like you were forcing it to use swift 5.9.2 (hence the tools version, i think). i removed that line so it uses xcode.

jhheider:
the main binary builds, but the framework is still complaining. using `swift package tools-version --set-current` is the suggested fix, but it seems insufficient.

PR #5192, opened 2024-02-13
+mpmath.org

and-ri:


projects/mpmath.org/package.yml
```diff
@@ -0,0 +1,24 @@
...
+
+versions:
+  url: https://mpmath.org
+  match: /mpmath-\d+\.\d+\.\d+\.tar\.gz/
+  strip:
+    - /^mpmath-/
+    - /\.tar\.gz$/
+
+dependencies:
+  python.org: '>=3.11'
```
mxcl: Python breaks between minor versions. In no case should we depend on it like this. Always depend on python with `~` eg. `python.org: ~3.11`

projects/mpmath.org/package.yml
```diff
@@ -0,0 +1,24 @@
...
+
+runtime:
+  env:
+    PYTHONPATH: ${{prefix}}/lib/python{{deps.python.org.version.major}}/site-packages:$PYTHONPATH
+
+build:
+  - python -m pip install --prefix={{prefix}} .
+  - run: ln -s python{{deps.python.org.version.marketing}} python{{deps.python.org.version.major}}
```
mxcl: this seems foolish unless mpmath have an explicit guarantee in their documentation that this will work.

If so then it is absolutely needed that we add a comment for that above this line.

mxcl (approved):

PR #5199, opened 2024-02-13
+github.com/jpsim/SourceKitten

and-ri:


projects/github.com/jpsim/SourceKitten/package.yml
```diff
@@ -0,0 +1,23 @@
...
+  ref: ${{version.tag}}
+
+versions:
+  github: jpsim/SourceKitten
+
+platforms:
+  - darwin
+
+build:
+  - patch -p1 < props/patch.diff
```
mxcl: patches require comments justifying why they are necessary.

mxcl (changes requested):

projects/github.com/jpsim/SourceKitten/package.yml
and-ri: I've added this comment and restricted the patch to version binding 👌🏻 

```
# https://github.com/jpsim/SourceKitten/pull/794 remove in release > 0.34.1
# fix this issue https://github.com/jpsim/SourceKitten/issues/793
- run: patch -p1 < props/patch.diff
  if: <=0.34.1
```

mxcl (approved):

mxcl:
Depends on a newer swift that macos 11 provides https://github.com/pkgxdev/pantry/actions/runs/7918947298/job/21618717961

Not sure the intricacies of this yet. That is: do we actually need this to depend on macOS 12 or can we provide a swift version that is newer?

and-ri:
> Depends on a newer swift that macos 11 provides https://github.com/pkgxdev/pantry/actions/runs/7918947298/job/21618717961
> 
> Not sure the intricacies of this yet. That is: do we actually need this to depend on macOS 12 or can we provide a swift version that is newer?

```
SourceKitten now requires Swift 5.6 or higher to build, and macOS 12
or higher to run.
```

from: https://github.com/jpsim/SourceKitten/releases/tag/0.33.0

mxcl:
k well we need to be able to specify these platform dependencies.

PR #5205, opened 2024-02-14
+vamp-plugins.org

and-ri:


mxcl:
Delegating to debian for tarballs and versions is not acceptable (unless absolutely necessary).

AFAICT the project provide tarballs, eg. https://code.soundsoftware.ac.uk/attachments/download/2588/vamp-plugin-sdk-2.9.0.tar.gz

Possibly they provide no listing, which would change my perspective a little. I still think we should source the tarball from the source. There are incredible security implications to the alternatives.

and-ri:
> Delegating to debian for tarballs and versions is not acceptable (unless absolutely necessary).
> 
> AFAICT the project provide tarballs, eg. https://code.soundsoftware.ac.uk/attachments/download/2588/vamp-plugin-sdk-2.9.0.tar.gz
> 
> Possibly they provide no listing, which would change my perspective a little. I still think we should source the tarball from the source. There are incredible security implications to the alternatives.

I got the archive from the official site. But I had to additionally get a link to the archive due to the fact that there is a variable part in it

```
# get archive url
- DIST_URL=$(curl -s https://vamp-plugins.org/develop.html | grep -o 'https://code.soundsoftware.ac.uk/attachments/download/[0-9]*/vamp-plugin-sdk-[0-9.]*.tar.gz')
- curl -L $DIST_URL | tar -xz --strip-components 1
```

mxcl:
Seems like we should use the URL scraper then if possible.

In general we need to get the sources from the *source* because:

1. security
2. revlock with other vendors

projects/vamp-plugins.org/package.yml
```diff
@@ -0,0 +1,47 @@
...
+  xiph.org/ogg: ^1.3
+  github.com/libsndfile/libsndfile: ^1.2
+
+build:
+  dependencies:
+    gnu.org/automake: '*'
+    curl.se: '*'
+  script:
+    # get archive url
+    - DIST_URL=$(curl -s https://vamp-plugins.org/develop.html | grep -o 'https://code.soundsoftware.ac.uk/attachments/download/[0-9]*/vamp-plugin-sdk-[0-9.]*.tar.gz')
```
mxcl: how does this download the specific version we want?

I’m confused why we can't use the `distributable` key.

and-ri: > how does this download the specific version we want?
> 
> I’m confused why we can't use the `distributable` key.

I'm sorry, it's my inattention
I fixed it
But still we have to find this url on the page. Because it contains characters (number) that are not related to the version

and-ri: For example: `https://code.soundsoftware.ac.uk/attachments/download/2588/vamp-plugin-sdk-2.9.0.tar.gz`

we need to find this part `/2588/`

Do we have a way to do this in `distributable`?

mxcl: Right I see. No. We don’t.

So yeah you'll have to do the curl line, but you need to at least fetch the right `version`.

and-ri: It seems to me that this version will be correct.

https://github.com/pkgxdev/pantry/commit/fb6678fa0abcebd1b3c860b487c2839eab05201d

PR #5207, opened 2024-02-14
+github.com/jaraco/keyring

and-ri:


projects/github.com/jaraco/keyring/package.yml
```diff
@@ -0,0 +1,25 @@
...
+
+dependencies:
+  pkgx.sh: ^1
+  github.com/python-cffi/cffi: ^1.16
+  github.com/eliben/pycparser: ^2.21
+  cryptography.io: ^42
+
+build:
+  dependencies:
+    python.org: '>=3.11'
```
mxcl, line 16: This is more acceptable because it is a build dependency. but at the same time it will result in build failures sporadically and it is better to constrain python correctly

mxcl (changes requested):

projects/github.com/jaraco/keyring/package.yml
```diff
@@ -0,0 +1,25 @@
...
+  script:
+    - bkpyvenv stage {{prefix}} {{version}}
+    - ${{prefix}}/venv/bin/pip install .
+    - bkpyvenv seal {{prefix}} keyring
+
+provides:
+  - bin/keyring
+
+test: keyring --list-backends | grep 'keyring'
```
mxcl, line 25: it is a tradition that text files end with newlines since the 70s, please try to adhere to it.

mxcl (approved):

PR #5208, opened 2024-02-14
+ladspa.org

and-ri:


projects/ladspa.org/package.yml
```diff
@@ -0,0 +1,42 @@
...
+
+platforms:
+  - linux
+
+dependencies:
+  github.com/libsndfile/libsndfile: ^1.2
+
+runtime:
+  env:
+    LADSPA_PATH: ${{prefix}}/lib/ladspa:$LADSPA_PATH
```
mxcl, line 20: are we sure this variable supports colon separation?

and-ri: > are we sure this variable supports colon separation?

I think yes https://archive.flossmanuals.net/pure-data/list-of-objects/hosting-plugins.html

PR #5211, opened 2024-02-14
+rbenv.org/ruby-build

and-ri:


projects/rbenv.org/ruby-build/package.yml
```diff
@@ -0,0 +1,37 @@
...
+  curl.se: ^8
+  linux:
+    gnu.org/make: '*'
+    gnu.org/gcc: '*'
+
+runtime:
+  env:
+    darwin:
+      LDFLAGS:
+        - -Wl,-rpath,{{pkgx.prefix}}
```
mxcl: Should add $LDFLAGS at the end

projects/rbenv.org/ruby-build/package.yml
```diff
@@ -0,0 +1,37 @@
+distributable:
+  url: git+https://github.com/rbenv/ruby-build.git
+  ref: ${{version.tag}}
+
+versions:
+  github: rbenv/ruby-build
+
+dependencies:
+  gnu.org/autoconf: ^2.72
```
mxcl: is this really needed?

projects/rbenv.org/ruby-build/package.yml
mxcl: Yeah the array format doesn't work here I think https://github.com/pkgxdev/pantry/actions/runs/7919080197/job/21619032587?pr=5211#step:9:77

projects/rbenv.org/ruby-build/package.yml
```diff
@@ -0,0 +1,36 @@
...
+  pyyaml.org/libyaml: ^0.2
+  freedesktop.org/pkg-config: ^0.29
+  gnu.org/readline: ^8.2
+  openssl.org: '>=1.1'
+  curl.se: ^8
+  linux:
+    gnu.org/make: '*'
+    gnu.org/gcc: '*'
```
mxcl, line 11: I think I need to provide a “build tools” shim package.

projects/rbenv.org/ruby-build/package.yml
mxcl: Most of these deps are `companions` at most and not `dependencies`. Since without them the binaries will run they just error for specific sub commands.

As a rule of thumb, if `--help` runs without the dep then its a companion.

projects/rbenv.org/ruby-build/package.yml
mxcl: I am constructing the shims pkg that can replace some of these deps. I'll merge in the mean time.

projects/rbenv.org/ruby-build/package.yml
```diff
@@ -0,0 +1,37 @@
+distributable:
+  url: git+https://github.com/rbenv/ruby-build.git
+  ref: ${{version.tag}}
+
+versions:
+  github: rbenv/ruby-build
+
+companions:
+  openssl.org: '*'
```
mxcl: This is a dynamic library and definitely is a required dependency and not a companion.

mxcl: well not that I looked, but companions are things that the package doesn't need to work. I feel it unlikely anything that needs openssl can work without it.

projects/rbenv.org/ruby-build/package.yml
```diff
@@ -0,0 +1,37 @@
...
+  curl.se: '*'
```
mxcl: same

projects/rbenv.org/ruby-build/package.yml
```diff
@@ -0,0 +1,37 @@
...
+  gnu.org/autoconf: '*'
+  freedesktop.org/pkg-config: '*'
+  gnu.org/readline: '*'
```
mxcl: same

projects/rbenv.org/ruby-build/package.yml
```diff
@@ -0,0 +1,37 @@
...
+  pyyaml.org/libyaml: '*'
```
mxcl: same

mxcl (changes requested):

PR #5232, opened 2024-02-16
+microcks.io

### LINKED ISSUE #5231: +microcks-cli
https://github.com/microcks/microcks-cli

and-ri:
closes #5231


mxcl:
`projects/microcks.io/microcks-cli` do microcks provide other things? If not then this can just be `projects/microcks.io` and in general that is our preference for the first open source package that domains provide even if in future they might provide more.

As a rule of thumb since the github is `microcks/microcks` I think it is clear that even they consider this the definitive package for their domain.

and-ri:
> `projects/microcks.io/microcks-cli` do microcks provide other things? If not then this can just be `projects/microcks.io` and in general that is our preference for the first open source package that domains provide even if in future they might provide more.
> 
> As a rule of thumb since the github is `microcks/microcks` I think it is clear that even they consider this the definitive package for their domain.

Yes, you are absolutely right, I fixed it

PR #5239, opened 2024-02-17
new/radmind.org/libsnet

magnusviri:
This library is required for radmind. I'm not sure if I should build this as a separate project or if I should just build it when I build radmind. I am going the separate route for now. If you guys think I should just include it in the radmind build I can close this and modify my radmind build.

I am actually the maintainer of radmind, even though I haven't written a line of code for it. It's a long story. Anyway, I'm trying to get radmind in pkgx. This library is on sourceforge and it is never going to be updated because the people who have access wont reply (ok, I didn't try too hard, my understanding is they don't want to do anything with the project anymore). So I've got the 1.0.0 version hardcoded because it's never going to change. I tried to figure out how to extract the version from the webpage but I couldn't figure it out and now I don't see the point. I need to move it to github actually and that's a lot of work and I don't know how to do it right yet. So anyway, I just want to get this packaged for now. let me know what you think.

### DIFF
--- a/projects/radmind.org/libsnet/package.yml
+++ b/projects/radmind.org/libsnet/package.yml
@@ -1,0 +1,33 @@
+distributable:
+  url: https://sourceforge.net/projects/libsnet/files/libsnet-{{version}}/libsnet-{{version}}.tar.gz
+  strip-components: 1
+
+display-name: libsnet
+
+versions:
+  - 1.0.0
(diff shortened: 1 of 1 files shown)


magnusviri:
Ok, I need to add a test. While doing that I started looking at this more and I realized it also uses zlib and cyrus-sasl. So I'm working on cyrus sasl and will submit that as soon as I can get it to build.

PR #5240, opened 2024-02-17
new/cyrusimap.org/cyrus-sasl

magnusviri:


magnusviri:
This is the build error (for Linux):

```
saslutil.c:280:3: error: call to undeclared function 'time'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  280 |   time(&now);
      |   ^
saslutil.c:364:34: error: call to undeclared function 'clock'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  364 |             ret[1] ^= (unsigned short) (clock() & 0xFFFF);
      |    
```

I searched online to figure out which library time.h is from and it looks like it comes with C. So I'm stumped.

jhheider:
possibly, just `build.env.linux.CFLAGS: $CFLAGS -Wno-implicit-function-declaration`?

magnusviri:
> possibly, just `build.env.linux.CFLAGS: $CFLAGS -Wno-implicit-function-declaration`?

That was it!

projects/cyrusimap.org/cyrus-sasl/package.yml
```diff
@@ -0,0 +1,35 @@
...
+
+display-name: cyrus-sasl
+
+versions:
+  github: cyrusimap/cyrus-sasl
+  strip: /^cyrus-sasl-/
+
+platforms:
+  - darwin
+  - linux/x86-64
```
jhheider: is this correct? please provide a `FIXME` comment so we know what the blocker is.

magnusviri: It's not correct. I can't remember why I put that there. Or maybe I copied it from some other project and didn't remove it. I can't remember. Anyway, I've committed a fix.

PR #5280, opened 2024-02-19
+charm.sh/skate

### LINKED ISSUE #5268: +skate
https://github.com/charmbracelet/skate

and-ri:
closes #5268


projects/charm.sh/skate/package.yml
```diff
@@ -0,0 +1,37 @@
+distributable:
+  url: git+https://github.com/charmbracelet/skate.git
```
mxcl: preferably use tarballs unless you need git for some reason or the project specifies to use git clones for installs. Is there a reason here?

rationale: I want to mirror all src tarballs again at some point and the source tarball is usually the intended way to install from source.

You've been right to use them for other go projects that embed the SHA tho for sure.

mxcl (changes requested):

projects/charm.sh/skate/package.yml
and-ri: > preferably use tarballs unless you need git for some reason or the project specifies to use git clones for installs. Is there a reason here?
> 
> rationale: I want to mirror all src tarballs again at some point and the source tarball is usually the intended way to install from source.
> 
> You've been right to use them for other go projects that embed the SHA tho for sure.

Okay, got it 👌🏻

PR #5287, opened 2024-02-20
+spin

### LINKED ISSUE #5285: +spin
https://github.com/fermyon/spin

jhheider:


jhheider:
will require using rustup to set up wasm targets; will need some work.

PR #5320, opened 2024-02-21
fix(sentencepiece)

### LINKED ISSUE #5276: ❌ build issues: google.com/sentencepiece=0.2.0
Running log of build failure for google.com/sentencepiece=0.2.0

jhheider:
complains about std::atomic_fetch_add

closes #5276


mxcl:
How about we wrap all this extremely specifically depoendent shit into a single package?

Clearly that is the only option that works anyway. Facebook are clowns.

PR #5357, opened 2024-02-24
+sourceware.org/dm

and-ri:
device-mapper => #5352


jhheider:
this line `-e "s+#LVM_PATH#+@bindir@/lvm+" \` in man/Makefile is choking because bindir has a `+` sign in it.

so, if you can't skip building the man pages, you might need to build to a `$(mktemp -d)` and move them afterward to `{{prefix}}`. :(

@mxcl this is a problem with `+brewing`, since people don't escape their regexen properly.

Issue #5359, opened 2024-02-25
remove my software from your AI generated shitshow


joeyh:
joeyh.name/code/moreutils

mxcl:
We only generate descriptions and images, nothing else is AI generated. But as you wish.

Issue #5358, opened 2024-02-24
I'm confused: what's with the project descriptions at https://pkgx.dev/pkgs/?


pawamoy:
Sorry if this has been reported elsewhere already, or if this is explained in docs somewhere, but I don't understand the contents you have in https://pkgx.dev/pkgs/. Lets take a few popular projects:

- https://pkgx.dev/pkgs/atuin.sh/

> Virtual reality mapping software.
Atuin is a software package that provides an interactive map and information on road toll rates, construction updates, and projects managed by the Central Texas Regional Mobility Authority. It also offers resources on business opportunities, financial information, and traveler information. Additionally, it allows users to pay tolls and provides information on road rules and violations.

This is... complete nonsense? Is it auto-generated by AI or something?

- https://pkgx.dev/pkgs/rye-up.com/

> Rye-up.com: Python project configuration.
Rye-up.com is a software package for Python projects that provides a user guide, community support, and a changelog. It helps with initializing search, syncing and locking projects, building, and publishing. Developed by mitsuhiko/rye, it aims to streamline the development process with its easy installation, configuration, and Python project management features.

"Provides a user guide, community support, and a changelog" :joy: Surely it cannot have been written by a human? It makes no sense either :confused: I thought AI did better than this nowadays?

- https://pkgx.dev/pkgs/github.com/squidfunk/mkdocs-material/

> Responsive documentation theme package.
The software package "squidfunk/mkdocs-material" is a documentation tool that is designed to be user-friendly and effective. It provides the ability to automate workflows, manage packages, and secure documentation. With features like code review and issue tracking, it allows for efficient collaboration and code improvement. It is suitable for enterprise teams, startups, and education purposes, offering CI/CD and DevOps solutions.

Sounds a bit better, but it's still plain wrong: there's no concept about automating workflows, managing packages, reviewing code or tracking issues in mkdocs-material... 

I really don't want to sound harsh or dismissive of any work that has been put in this package manager or its documentation, but IMO putting non-sense like this in the official project listing completely breaks any trust one could have in pkgx? Why listing everything with poorly generated contents? All these projects have a description on GitHub, can't that be used instead? Or maybe fetch metadata from the respective registries?

I get the generated pictures, waiting for maintainers to put their logo instead, but the descriptions...

I'm really confused :sweat_smile:

pawamoy:
[...]

PR #5466, opened 2024-03-07
fix(atuin): remove Atuin

ellie:
pkgx has some issues that are leading to users opening more issues on the Atuin repo

See:

1. https://github.com/atuinsh/atuin/issues/1765
2. https://github.com/atuinsh/atuin/issues/1809

Given that that's two issues almost within a week of each other, there are likely many more people having a sub-par experience and not even realising that this is due to pkgx.

I'm also concerned about three other things

1. Atuin being listed under a non-canonical name (atuin.sh, rather than just atuin)
2. Atuin having been misrepresented in the past (see https://github.com/pkgxdev/pantry/issues/5358)
3. [Some weird crypto thing](https://connortumbleson.com/2024/02/26/the-disappointing-tea-xyz/), which seems related?

It also looks like I'm not alone here:

1. https://github.com/pkgxdev/pantry/pull/5360
2. https://github.com/pkgxdev/pantry/issues/5359

I'd rather avoid Atuin being distributed by pkgx, at least for now.


jhheider:
No problem, @ellie . Sorry for any trouble `pkgx` has caused you.

In response to your 3 concerns, for posterity's sake:

1. all packages are listed under URL-style namespaces to avoid the name collisions other package managers run into over time. so far, it seems to be working well.
2. we have removed the AI-generated placeholder descriptions from the website, due to concerns. they were merely meant to be an aid to package location, and are not core to the package manager.
3. participation in the `tea` protocol is entirely separate from the `pkgx` package manager (hence the rebrand to make that clear).

Happy to merge this for you. Atuin is always welcome back if circumstances change.

PR #5468, opened 2024-03-07
+openjdk.org@21

felipecrs:


projects/openjdk.org/package.yml
```diff
@@ -144,6 +152,8 @@ provides:
   - bin/jstack
   - bin/jstat
   - bin/jstatd
+  # TODO: if: {{version.major}} >= 21
+  # - bin/jwebserver
```
felipecrs, line 156: Any ideas about this?

jhheider: just need to omit it; it would be a potential upgrade to libpkgx if desired.

jhheider:
great addition. thanks!

projects/openjdk.org/package.yml
felipecrs: https://github.com/pkgxdev/libpkgx/issues/68

Issue #5467, opened 2024-03-07
+adoptium.net/temurin/jdk +adoptium.net/temurin/jre


felipecrs:
It would be nice to have Eclipse Adoptium's Temurin Java distribution also packaged.

https://adoptium.net/temurin/releases/

And I think it'd be even greater to have JRE and JDK variants. For example, some Java-based tools may not need the full JDK to run.

Fortunately, Adoptium provides scripts to build any JDK/JRE more easily:

https://github.com/adoptium/temurin-build

They seem well documented as well.

Refs https://github.com/pkgxdev/pantry/issues/89#issuecomment-1734645955

PS: if you have no interest of packaging this in, please leave it open, I may try figuring it out someday.

felipecrs:
Refs https://github.com/docker-library/openjdk/issues/505

PR #5472, opened 2024-03-07
+openjdk.org@11

felipecrs:


jhheider:
won't be able to test-build 11 without a merge, but odds are still good. i'll link the build attempt back to this, so we can keep everything together in case changes are needed.

felipecrs:
I'm also trying to verify it locally... Will mark as ready to review when I feel it's ok.

projects/openjdk.org/package.yml
```diff
@@ -81,16 +92,18 @@ build:
...
       BOOT_JDK_DIR: $SRCROOT/boot-jdk
       JDK_DIR: build/*/images/jdk/*
       ARGS:
+        - --with-x={{deps.x.org/x11.prefix}} # from homebrew's formulae
         - --with-cups={{deps.openprinting.github.io/cups.prefix}}
         - --with-fontconfig={{deps.freedesktop.org/fontconfig.prefix}}
         - --with-freetype=system
         - --with-stdc++lib=dynamic
-        - --with-toolchain-type=clang
+        - --with-toolchain-type=clang # not in brew's formulae
```
felipecrs: what do you think about this one? not in homebrew's formulae

projects/openjdk.org/package.yml
```diff
@@ -81,16 +92,18 @@ build:
       JDK_DIR: build/*/images/jdk-bundle/jdk-{{version.marketing}}.{{version.patch}}.jdk/Contents/Home/*
       ARGS:
         - --enable-dtrace
+        # - --with-sysroot=#{MacOS.sdk_path} # from homebrew's formulae
     linux:
       STRIP: 1
       BOOT_JDK_DIR: $SRCROOT/boot-jdk
       JDK_DIR: build/*/images/jdk/*
       ARGS:
+        - --with-x={{deps.x.org/x11.prefix}} # from homebrew's formulae
```
felipecrs: I think this one was missing indeed

projects/openjdk.org/package.yml
```diff
@@ -81,16 +92,18 @@ build:
       JDK_DIR: build/*/images/jdk-bundle/jdk-{{version.marketing}}.{{version.patch}}.jdk/Contents/Home/*
       ARGS:
         - --enable-dtrace
+        # - --with-sysroot=#{MacOS.sdk_path} # from homebrew's formulae
```
felipecrs: I found this additional flag missing as compared to homebrews. Any thoughts?

@jhheider

projects/openjdk.org/package.yml
```diff
@@ -64,12 +68,19 @@ build:
...
-      --with-vendor-version-string="$(pkgx --version | grep -o '[0-9]\+\.[0-9]\+\.[0-9]\+')"
-      --with-extra-cflags="$CFLAGS"
-      --with-extra-cxxflags="$CXXFLAGS"
-      --with-extra-ldflags="$LDFLAGS"
+    - |
+      if [ {{version.major}} -gt 11 ]; then
+        version_args=--with-version-build={{version.minor}}
+      else
+        version_args="--disable-hotspot-gtest --with-conf-name=release --with-jvm-features=shenandoahgc"
+      fi
```
felipecrs: @jhheider this was the best solution I thought of to condition the args. LMK if you have a better suggestion

projects/openjdk.org/package.yml
jhheider: we try to use clang throughout; if it doesn't require that flag, then it should be ok to omit.

projects/openjdk.org/package.yml
jhheider: see `gcc`:
```yaml
    darwin:
      ARGS:
        # Reliance on CLT hard path is yuck.
        - --with-sysroot=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
```

felipecrs:
@jhheider I went through all formulaes for openjdk@11, @17 and @21 in Homebrew and I found some differences. I commented on each one just to highlight and talk about it.

Please let me know what do you think of each, and then I can remove the comments.

Also, I was not able to build on my machine even after the adjustments:

```console
$ bk build openjdk@11
...
configure: WARNING: Ignoring value of STRIP from the environment. Use command line variables instead.
checking for strip... /home/felipecrs/.pkgx/llvm.org/v16.0.6/bin/strip
checking for nm... /home/felipecrs/.pkgx/llvm.org/v16.0.6/bin/nm
checking for gobjcopy... [not found]
checking for objcopy... /home/felipecrs/.pkgx/llvm.org/v16.0.6/bin/objcopy
checking for gobjdump... [not found]
checking for objdump... /usr/bin/objdump
checking for llvm-llvm-llvm-cxxfilt... [not found]
configure: error: Could not find required tool for CXXFILT
configure exiting with result code 1
error: Uncaught (in promise) Error: UR BUILD FAILED WITH CODE 1 & SIGNAL null
if (!rv.success) throw new Error(`UR BUILD FAILED WITH CODE ${rv.code} & SIGNAL ${rv.signal}`)
                       ^
    at file:///home/felipecrs/.pkgx/pkgx.sh/brewkit/v1.14.2/bin/bk-build:123:24
    at eventLoopTick (ext:core/01_core.js:182:7)
```

I think there's some system-level dependencie not in sync with what the CI environment have.

projects/openjdk.org/package.yml
jhheider: probably use two step lines:
```yaml
  - run: version_args=--with-version-build={{version.minor}}
    if: '>=12'
  - run: version_args="--disable-hotspot-gtest --with-conf-name=release --with-jvm-features=shenandoahgc"
    if: <12
```

projects/openjdk.org/package.yml
felipecrs: Sorry, do you mean I should set it or do you mean it's already set by pkgx's gcc and therefore it shouldn't be repeated here?

(I don't use darwin)

projects/openjdk.org/package.yml
felipecrs: oh wow. I guess this means pkgx compiles a single shell script with all the commands before running for this to work then. Will try, thanks.

jhheider:
> @jhheider I went through all formulaes for openjdk@11, @17 and @21 in Homebrew and I found some differences. I commented on each one just to highlight and talk about it.
>
> checking for llvm-llvm-llvm-cxxfilt... [not found]
> configure: error: Could not find required tool for CXXFILT
> configure exiting with result code 1
> error: Uncaught (in promise) Error: UR BUILD FAILED WITH CODE 1 & SIGNAL null
>
> I think there's some system-level dependencie not in sync with what the CI environment have.

probably just make the `llvm.org` dep for all platforms and, if it's still not happy, set the env var `CXXFILT: llvm-cxxfilt`.

projects/openjdk.org/package.yml
felipecrs: let's keep it then. Looks like it defaults to gcc on Linux.

projects/openjdk.org/package.yml
jhheider: if it finds it, best to omit it; if it starts panicking about the sdk, then add it in.

projects/openjdk.org/package.yml
jhheider: yep, script-arrays were added last fall, and allow for better control of steps.

felipecrs:
Is there any way to access what's in the fourth digit of the version? I could not find any example in the pantry. Example: `{{version.buildnumber}}` or whatever.

felipecrs:
I addressed the `CXXFILT` issue in the last commit. Also, I tweaked the version parameters a little bit.  I'm pretty sure even Homebrew were not passing these parameters correctly.

I'm basing on https://github.com/adoptium/temurin-build/wiki/A-Conversation-about-versions-and-setting-them-in-build-configuration, and if I'm correct it should now output like:

```console
$ java -version
openjdk 21.0.3 2024-04-16
OpenJDK Runtime Environment pkgx@1.1.6 (build 21.0.3+6)
OpenJDK 64-Bit Server VM pkgx@1.1.6 (build 21.0.3+6, mixed mode, sharing)
```

But I was not able to test it yet. Locally the compilation is now starting, but it's failing.

felipecrs:
Compiling jdk@17 and @21 errors with:

```
configure: Toolchain type clang is not valid on this platform.
configure: Valid toolchains: microsoft.
configure: error: Cannot continue.
configure exiting with result code 1
```

I suspect it has something to do with the fact I'm on WSL.

Compiling @11 fails differently though:

```
Compiling 4 files for BUILD_JIGSAW_TOOLS
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f936d44ffc0, pid=203466, tid=203472
#
# JRE version:  (11.0.23+6) (build )
# Java VM: OpenJDK 64-Bit Server VM (11.0.23+6, mixed mode, tiered, compressed oops, serial gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0xe33fc0]  ObjectSynchronizer::inflate(Thread*, oopDesc*, ObjectSynchronizer::InflateCause)+0xf0
#
# No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /home/felipecrs/repos/pantry/builds/openjdk.org-11.0.23.6/make/hs_err_pid203466.log
#
#
/usr/bin/bash: line 1: 203466 Aborted                 /home/felipecrs/repos/pantry/builds/openjdk.org-11.0.23.6/build/linux-x86_64-normal-server-release/jdk/bin/java -Xms64M -Xmx1600M -XX:ThreadStackSize=1536 -XX:+UseSerialGC -Xms32M -Xmx512M -XX:TieredStopAtLevel=1 -cp /home/felipecrs/repos/pantry/builds/openjdk.org-11.0.23.6/build/linux-x86_64-normal-server-release/buildtools/tools_jigsaw_classes --add-exports java.base/jdk.internal.module=ALL-UNNAMED build.tools.jigsaw.AddPackagesAttribute /home/felipecrs/repos/pantry/builds/openjdk.org-11.0.23.6/build/linux-x86_64-normal-server-release/jdk > >(/usr/bin/tee -a /home/felipecrs/repos/pantry/builds/openjdk.org-11.0.23.6/build/linux-x86_64-normal-server-release/jdk/_packages_attribute.done.log) 2> >(/usr/bin/tee -a /home/felipecrs/repos/pantry/builds/openjdk.org-11.0.23.6/build/linux-x86_64-normal-server-release/jdk/_packages_attribute.done.log >&2)
gmake[3]: *** [ExplodedImageOptimize.gmk:41: /home/felipecrs/repos/pantry/builds/openjdk.org-11.0.23.6/build/linux-x86_64-normal-server-release/jdk/_packages_attribute.done] Error 134
gmake[2]: *** [make/Main.gmk:411: exploded-image-optimize] Error 2
```

Crazy stuff. When you get a chance, would you mind trying on your machine?

jhheider:
> I suspect it has something to do with the fact I'm on WSL.

WSL2, hopefully? I would hope but would never be sure that it should detect linux correctly. You can always use `docker` to ensure you're getting a "real" linux environment.

I can try it out locally. It'll probably be tomorrow.

felipecrs:
Yeah, WSL2 for sure. The `bk docker build` is unfortunately broken too. Probably coming from brewkit.

```console
❯ bk docker build java@11 --clean
+ exec docker run --name brewkit.pkgx.dev --rm --platform linux/amd64 --volume /home/felipecrs/.pkgx/pkgx.sh/brewkit:/brewkit --volume /home/felipecrs/repos/pantry:/work --volume /home/felipecrs/Library/Caches/pkgx:/root/.cache/pkgx --volume brewkit.pkgx.dev.linux-amd64:/root/.pkgx --env DENO_DIR=/root/.pkgx/.cache/deno --env PKGX_PANTRY_PATH=/work --env GITHUB_TOKEN=* --env CLICOLOR_FORCE=1 --workdir /work pkgxdev/pkgx /brewkit/bin/bk build java@11 --clean
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/brewkit/bin/bk": stat /brewkit/bin/bk: no such file or directory: unknown.
```

Shouldn't be difficult to fix, but anyway.

I'll let you know if I have any news. No rush at all. Thanks.

felipecrs:
BTW it's mounting `/home/felipecrs/.pkgx/pkgx.sh/brewkit:/brewkit`, it should probably mount `/home/felipecrs/.pkgx/pkgx.sh/brewkit/v*`.

- https://github.com/felipecrs/brewkit/commit/7682ddbd811060684a5e9b39859b6e10008d93ae

felipecrs:
Ok, I tried with docker here. Same problem unfortunately. Chances are you'll face the same issue.

jhheider:
@felipecrs note that https://jdk.java.net/archive/ has the oldest linux/aarch64 jdk as v15 and the oldest darwin/aarch64 as v17. perhaps openjdk.org/v11 should be its own package (like pcre.org/v2) to exclude those two environments?

i can try and push that change to this branch to see if it gets us closer.

felipecrs:
Oh good point, I had not noticed.

But this is not a problem, we can download the boot JDK for such platforms from different vendors.

Homebrew is set to download them from zulu, for example.

https://github.com/Homebrew/homebrew-core/blob/ef4f9890a315195fd09884fa0f035bf226eb7f66/Formula/o/openjdk%4011.rb

Also Homebrew builds ARM64 variants for JDK11 for both Linux and Mac. So it should not be a problem for pkgx to do too.

felipecrs:
I see you pushed some changes, but if you don't mind I can revert them and make the fixes for the boot jdk depending on the platform here.

I would super prefer to keep a single package if possible, to make it user-facing consistent with other pkgs like node, python, etc.

felipecrs:
nevertheless, let's see how CI builds v11. :D

jhheider:
definitely; if there's an alternate boot jdk, let's use it.

felipecrs:
Error in CI was the same in my local btw. That's good news I guess. lol

jhheider:
possibly 11 doesn't play nice with `clang`. can try `gcc` in its place.

felipecrs:
Will try. Thanks.

I noticed this btw:

```
clang-16: warning: argument unused during compilation: '-pie' [-Wunused-command-line-argument]
```

jhheider:
should be harmless. it's a linker flag, so it should be in the linking step but won't affect compilation.

felipecrs:
Ok, compiling with llvm.org@12 works. :(

Will check the highest version that still works.

But any ideas to keep using a newer llvm for newer JDK but use an older for 11?

jhheider:
> But any ideas to keep using a newer llvm for newer JDK but use an older for 11?

So, the only trick we could currently is something like this:
```yaml
  - run: |
      pkgx +llvm.org^12
      PATH="{{deps.llvm.org.prefix}}/../v12/bin:$(echo $PATH | tr ':' '\n' | sed '/llvm.org/d' | tr '\n' ':')"
    if: <17
```
roughly.

too many such tricks, and we really should split the package off.

felipecrs:
The version params worked nice btw:

image

felipecrs:
> So, the only trick we could currently is something like this:
> 
> ```yaml
>   - run: |
>       pkgx +llvm.org^12
>       PATH="{{deps.llvm.org.prefix}}/../v12/bin:$(echo $PATH | tr ':' '\n' | sed '/llvm.org/d' | tr '\n' ':')"
>     if: <17
> ```
> 
> roughly.

llvm.org <16 works. For now I suggest we keep it for all versions, what do you think? Will probably be a problem again when pkging jdk8. But that's for later.

> too many such tricks, and we really should split the package off.

I would suggest instead to make brewkit support conditioning build dependencies... https://github.com/pkgxdev/brewkit/issues/319

jhheider:
indeed, but it means package resolution becomes more complex (which is why we haven't done it yet).

if it's happy with `llvm^15`, i'm happy there for now.

jhheider:
great work, @felipecrs ! building old software is sometimes even more challenging.

felipecrs:
BTW I think I'm not gonna do openjdk@8, the differences between it and 11 in Homebrew scared me.

Also Homebrew doesn't build Linux ARM not Mac ARM.

Perhaps I'll give it a try when pkging https://github.com/pkgxdev/pantry/issues/5467, since that one is still ~maintained.

felipecrs:
@jhheider would it be possible to trigger a build for openjdk@11? I suppose it has to be done manually.

jhheider:
Looks like it failed the audit due to a missing binary:

https://github.com/pkgxdev/pantry/actions/runs/8210262716/job/22457399093

PR #5500, opened 2024-03-09
-openjdk@20

felipecrs:
- No one uses it. It's like Node.js 19, people only "use" it when they want to test something, but it's immediately forgotten once a new LTS version comes out -- which happened already with java@21.
- Waste of machine time and resources
- Not even sdkman.io (a jdk package manager) bothers to support it anymore
- It may raise a red flag for people looking into https://pkgx.dev/pkgs/openjdk.org/ (like "why would they package 11, 17, 21 and 20? do they know what they are doing?" lol)
- May make version specifiers more complicated. I.e., if I want to specify an older version of java 21, I can't just use <21 because that would include 20 although the user probably meant 17.

I know these are all weak arguments, but I am just opening the PR in case you agree. If you do I'd suggest to even unpublish the existing java@20 package previously built. If you don't agree, fell 100% free to just close this PR.

### DIFF
--- a/projects/openjdk.org/package.yml
+++ b/projects/openjdk.org/package.yml
@@ -2,8 +2,6 @@ distributable:
   # TODO: add jdk8
   - url: https://github.com/openjdk/jdk21u/archive/{{version.tag}}.tar.gz
     strip-components: 1
-  - url: https://github.com/openjdk/jdk20u/archive/{{version.tag}}.tar.gz
-    strip-components: 1
   - url: https://github.com/openjdk/jdk17u/archive/{{version.tag}}.tar.gz
     strip-components: 1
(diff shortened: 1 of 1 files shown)


jhheider:
Well, our ideal is build *every* version of everything. Unrealistic, of course, but it would be the meaningful counterargument.

What do you think, @mxcl ? Removing non-lts versions seems harmless, until someone wants one for some niche reason.

mxcl:
I think with our current system this is probs the right choice (for node as well) since users get the latest version if they use pkgx normally and usually the don't want the dev/beta releases.

felipecrs:
> I think with our current system this is probs the right choice (for node as well) since users get the latest version if they use pkgx normally and usually the don't want the dev/beta releases.

Btw it isn't a problem right now because the current latest version provided by pkgx is already 21 LTS.

PR #5505, opened 2024-03-09
Remove `bin/jpackage` for now to fix openjdk@11

felipecrs:
I think this is the best compromise for the time being. `bin/jpackage` is only available for jdk17+.

              Looks like it failed the audit due to a missing binary:

https://github.com/pkgxdev/pantry/actions/runs/8210262716/job/22457399093

_Originally posted by @jhheider in https://github.com/pkgxdev/pantry/issues/5472#issuecomment-1986928440_

Refs https://github.com/pkgxdev/libpkgx/issues/68

### DIFF
--- a/projects/openjdk.org/package.yml
+++ b/projects/openjdk.org/package.yml
@@ -140,6 +140,8 @@ build:
       - --with-zlib=system
 
 provides:
+  # TODO: https://github.com/pkgxdev/libpkgx/issues/68
+  # - bin/jaotc # <=11
   - bin/jar
   - bin/jarsigner
   - bin/java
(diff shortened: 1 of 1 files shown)


felipecrs:
One sec, I think there are more differences. I'm double checking.

felipecrs:
Done documenting all bins diffs.

jhheider:
it looks like clang 15 is still erroring for linux:

https://github.com/pkgxdev/pantry/actions/runs/8216510165/job/22471169048

felipecrs:
I will check this better tomorrow. The only explanation is that perhaps my bk build command was using some cached llvm.org.

jhheider:
possibly cached build products. `bk build --clean openjdk.org^11` is slower (since it cleans everything), but won't have old intermediate products.

felipecrs:
I was actually using --clean every time. However, my suspicious is that, since I first tried with llvm<14, all following builds would be reusing it since 13 satisfies llvm<16 for example.

jhheider:
ah, very possible. `^{{major}}` will help with that when testing.

PR #5507, opened 2024-03-09
+openshift.com

and-ri:


jhheider:
looks like it needs -buildmode=pie on linux too.

PR #5509, opened 2024-03-09
+github.com/zaach/jsonlint

and-ri:


projects/github.com/zaach/jsonlint/package.yml
```diff
@@ -0,0 +1,36 @@
...
+  script:
+    - npm i $ARGS .
+    - run: ln -s ../libexec/bin/jsonlint jsonlint
+      working-directory: '{{prefix}}/bin'
+  env:
+    ARGS:
+      - -ddd
+      - --global
+      - --build-from-source
+      - --prefix={{prefix}}/libexec
```
jhheider: why not just use `{{prefix}}`?

and-ri: yes, we can definitely use the prefix without libexec 👍🏻 

i changed it

PR #5510, opened 2024-03-09
fix(kubectx.dev)

felipecrs:
This fixes multiple problems with this package:

1. Since 0.9.0 the package is now made in Go, and requires no kubectl nor bash to run
2. It always required fzf to be installed for full functionality, but it was not set as a dependency

Looks like [Homebrew is outdated](https://github.com/Homebrew/homebrew-core/blob/af831d4c8a82c0789f9447e16033c6e1675181eb/Formula/k/kubectx.rb), and this pkg was based on it.

### DIFF
--- a/projects/kubectx.dev/package.yml
+++ b/projects/kubectx.dev/package.yml
@@ -1,6 +1,6 @@
 distributable:
-  url: https://github.com/ahmetb/kubectx/archive/refs/tags/v{{version}}.tar.gz
-  strip-components: 1
+  url: git+https://github.com/ahmetb/kubectx.git
+  ref: ${{version.tag}}
 
 versions:
   github: ahmetb/kubectx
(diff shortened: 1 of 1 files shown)


jhheider:
it seems like the build script should use `if: <0.9`/`if: '>=0.9'` tags to separate the logic out. otherwise, good catch.

felipecrs:
I honestly think it's not worth adding support for older versions of this pkg.

Also, runtime dependencies would be different... So I suggest to keep it like this, only support for 0.9+.

jhheider:
should be able to add an `ignore:` stanza to the versions then, to prevent the old ones from being seen as available.

felipecrs:
Makes sense. I did it then, based on https://github.com/felipecrs/pantry/blob/694800830cc272bb9dcc63d77e6259f0e19ab278/projects/pip.pypa.io/package.yml#L8.

I'm not sure how to test the `ignore` though (`bk build kubectx.dev@0.8` do attempt to build 0.8 btw, I was hoping that it would error out).

felipecrs:
Thank you!

PR #5588, opened 2024-03-14
+nodejs.org@10

felipecrs:
Refs https://github.com/pkgxdev/brewkit/issues/319
Refs https://github.com/pkgxdev/pantry/issues/5582

PR #5605, opened 2024-03-14
Fix import zlib for older versions of python

felipecrs:
Needed by https://github.com/pkgxdev/pantry/pull/5588

### DIFF
--- a/projects/python.org/package.yml
+++ b/projects/python.org/package.yml
@@ -34,7 +34,6 @@ interprets:
   args: python
 
 dependencies:
-  # recommended (but none are actually required)
   zlib.net: 1
   sourceware.org/bzip2: 1
   openssl.org: ^1.1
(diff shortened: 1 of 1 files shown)


felipecrs:
@jhheider `bk test python@2` works on my machine with `import zlib`.

But I added LDFLAGS for all versions so let's see if it is going to affect newer versions of Python.

jhheider:
lgtm; i'll try a rebuild of python^2, and if that works, i'll merge and restock our pythons

jhheider:
er, no. order of operations. i'll merge, restock, then complain if it didn't work ;).

jhheider:
restocking: https://github.com/pkgxdev/pantry/actions/runs/8286005182

jhheider:
hm, that doesn't seem to have resolved the import zlib issue :/

felipecrs:
That's super odd. I ran `bk test python@2` right now just to confirm and it passed. I'm investigating it. Sorry about that.

image


Building python takes ages though. It's hard to iterate. :(

jhheider:
no worries. things are nice and clean in CI (to the best of our ability), so sometimes it catches things that aren't found locally due to left-over artifacts.

felipecrs:
Update: I was able to reproduce the issue now, through docker.

PR #5611, opened 2024-03-15
Fix import zlib for older versions of python (again)

felipecrs:


felipecrs:
It worked on `bk docker test` now. But I will perform at least:

1. One more fully clean try
2. One more fully clean try without this PR to make sure I can reproduce the issue again

felipecrs:
@jhheider the test described above succeeded. I was able to reproduce the issue again and then confirm the fix.

jhheider:
you're the man, @felipecrs . such a little thing to be broken.

felipecrs:
https://github.com/pkgxdev/pantry/actions/runs/8297499542/job/22709806129#step:6:145

OMG, do you think it broke other stuff?

jhheider:
it looks like that doesn't fix the _ctypes import issue, at least on darwin+aarch64. if it fixes zlib, it might be sufficient to keep the other version disabled.

PR #5616, opened 2024-03-15
Skip known failing python@2 test on macOS

felipecrs:


felipecrs:
https://github.com/pkgxdev/pantry/actions/runs/8297499542/job/22709806129#step:6:145

OMG, do you think it broke other stuff?

_Originally posted by @felipecrs in https://github.com/pkgxdev/pantry/issues/5611#issuecomment-1999832115_

This is the only thing I can think of. `LDFLAGS` comes with `-pie` by default on `bk build`. None was set for `CPPFLAGS` though.

I have no idea whether this will fix anything, but it's probably good.

Since the build apparently passed on Linux but failed on Mac, I can also consider only setting these vars on Linux. But I don't know if that will mean `import zlib` will fail on Mac. I have nowhere to test it unfortunately.

jhheider:
it looks like what failed was `import _ctypes`, so it's possible that isn't fixed by the zlib fixes. maybe? it was only darwin+aarch64, which is newer than python2 by a bit, so that might be contributing. it might be ok to still gate that test.

felipecrs:
> it might be ok to still gate that test.

You mean skipping the ctypes import test for now for darwin? (without a way to reproduce I'm unlikely to be able to fix it)

jhheider:
yes, the original before 0910eeea8cdf97da7e1124ffb9daa5dd2e0beb55 had both _ctypes and zlib marked out for prior to 3.6.4. if you want to just do that for _ctypes, we're still in improvement range.

felipecrs:
The only problem is that I don't know whether `import zlib` works on python@2@darwin or not...

And there's currently no easy way to `if: ^3.6.4 || linux`.

Anyway, let me know what you think about the way I left.

jhheider:
i think that's the solution for now. the other option could be:
```yaml
  dependencies:
    crates.io/semverator: '*'
  script:
    # FIXME: darwin <3.6.4 fails
    - run: |
        if test "{{hw.platform}}" = "linux"; then
          python -v -c "import _ctypes"
        elif semverator gt {{version}} 3.6.3; then
          python -v -c "import _ctypes"
        fi
```
which is more precise. the gains are small, though.

felipecrs:
That's a cool project btw. I use [npm's semver cli](https://github.com/npm/node-semver), which provides a very similar CLI, but being a npm package makes it a lot harder to get access to. Having to install node.js and npm just to get access to it is... sometimes too much.

jhheider:
Thanks. It was very useful in the early days, before we got script arrays with `if:` stanzas. Still useful for some of these complex edge cases.

jhheider:
fingers crossed: https://github.com/pkgxdev/pantry/actions/runs/8310629532

PR #5685, opened 2024-03-22
fix simd rust versioning issues

jhheider:


jhheider:
nope, looks like we're waiting for 1.78.

PR #5705, opened 2024-03-25
+org-stats -- Get the contributor stats summary from all repos of any given organization

kevinmichaelchen:
https://github.com/caarlos0/org-stats

### DIFF
--- a/projects/github.com/caarlos0/org-stats/package.yml
+++ b/projects/github.com/caarlos0/org-stats/package.yml
@@ -30,4 +30,4 @@ build:
       LDFLAGS:
         - -buildmode=pie
 
-test: org-stats --version
+test: org-stats version
--- a/projects/github.com/caarlos0/org-stats/package.yml
+++ b/projects/github.com/caarlos0/org-stats/package.yml
@@ -10,17 +10,18 @@ provides:
 
 build:
   dependencies:
-    go.dev: ^1.19
-  script: |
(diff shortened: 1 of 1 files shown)


kevinmichaelchen:
cc @caarlos0

kevinmichaelchen:
```
$ org-stats version

https://github.com/caarlos0/org-stats (devel) @ none
```

Not sure if there's a way to set the version.

- https://pkg.go.dev/runtime/debug#BuildInfo
- https://github.com/caarlos0/org-stats/blob/master/cmd/version.go#L19

jhheider:
according to https://github.com/golang/go/issues/29228 (which is, admittedly, old) `BuildInfo.Main.Version` _always_ returns `"(devel)"`. so, the best we can do is update the source.

caarlos0:
you can pass -ldflags to `go build` with something like `-s -w -X main.version={{ .Version }} -X main.commit={{ .Commit }} -X main.date={{ .CommitDate }}`

caarlos0:
this is how its built btw: https://raw.githubusercontent.com/caarlos0/goreleaserfiles/main/build.yml

jhheider:
hm. i think `main.version` is the only capitalization i _didn't_ try. i'll update it.

jhheider:
hm, interestingly, not working for me. i'll keep playing with it.

PR #5732, opened 2024-03-28
+tailwindcss.com

devuo:
Add [tailwindcss](https://tailwindcss.com/)


jhheider:
are we able to build this from source, @devuo ? it looks like it should only need pnpm, node, and rust, possibly. if we're not, then we should add the top-level:
```
warnings:
  - vendored
```
key so we know to re-examine it in the future.

devuo:
We are able to "build it" but that will require node runtime to run tailwind. I opted for this approach as it's a [self-contained executable](https://tailwindcss.com/blog/standalone-cli) that does not require a separate node runtime, but if you think its better to run it that way I'm happy to change it.

jhheider:
hm, can we emulate their build process to produce the same thing? our first preference is naturally to build it, but if that's cost- or time-prohibitive, vendored binaries is the next best thing.

happy to take this as a first pass. we can just add the `warnings` key and a comment about why we went this way.

devuo:
comment and vendored warning added

jhheider:
looks good!

PR #5738, opened 2024-03-29
+joeyh.name/moreutils

tannevaled:


jhheider:
removed at the request of the author: https://github.com/pkgxdev/pantry/issues/5359

PR #5750, opened 2024-03-30
Drop dependency on unzip for deno, use pkgd denort by default

felipecrs:


felipecrs:
I tested locally (linux x86-64) with deno@1.41 (since CI will test deno@1.42), and it's working.

jhheider:
`DENORT_BIN` is a good find.

jhheider:
great work!

jhheider:
rebuilds triggered for 1.40.5, 1.41.*, and 1.42 to back-fill the missing linux-aarch64 versions.

PR #5751, opened 2024-03-30
Remove llvm.org as runtime dependency of deno on Linux

felipecrs:
This works on my machine (linux x86-64). I'm using this PR to trigger CI to test on other platforms.

Refs https://github.com/pkgxdev/pantry/commit/5f7af42e953e97e4b0e1a8df01254f39a7c41eaf#commitcomment-140400383

### DIFF
--- a/projects/deno.land/package.yml
+++ b/projects/deno.land/package.yml
@@ -18,10 +18,10 @@ runtime:
     DENO_NO_UPDATE_CHECK: 'true'
     DENORT_BIN: '{{prefix}}/bin/denort'
 
-# dependencies:
-#   darwin/x86-64:
-#     # FIXME try removing this after new builds are available
-#     llvm.org: 17 # libunwind
+dependencies:
+  darwin/x86-64:
+    # FIXME try removing this after new builds are available
+    llvm.org: 17 # libunwind
 
 build:
   script:
--- a/projects/deno.land/package.yml
+++ b/projects/deno.land/package.yml
@@ -18,10 +18,10 @@ runtime:
     DENO_NO_UPDATE_CHECK: 'true'
     DENORT_BIN: '{{prefix}}/bin/denort'
 
-dependencies:
-  darwin/x86-64:
-    # FIXME try removing this after new builds are available
-    llvm.org: 17 # libunwind
+# dependencies:
+#   darwin/x86-64:
+#     # FIXME try removing this after new builds are available
+#     llvm.org: 17 # libunwind
 
 build:
   script:
--- a/projects/deno.land/package.yml
+++ b/projects/deno.land/package.yml
@@ -37,6 +37,12 @@ build:
 
     - |
       rust_version=$(yq -er .toolchain.channel rust-toolchain.toml)
+      if semverator lt "$rust_version" 1.67.0; then
+        # older versions of pkgx's rustc --print=split-debuginfo fails with:
+        # error: unknown print request `split-debuginfo`
+        # maybe due to newer cargo?
+        rust_version=1.67.0
+      fi
       set -o allexport
       source <(pkgx "+rust-lang.org~${rust_version}" +rust-lang.org/cargo^0)
       set +o allexport
@@ -65,6 +71,7 @@ build:
     cmake.org: ^3 # deno.land>=1.36.1 requires cmake
     protobuf.dev: '*' # deno.land>=1.36.4 requires protoc
     github.com/mikefarah/yq: ^4
+    crates.io/semverator: ^0
 
 test:
   dependencies:
--- a/projects/deno.land/package.yml
+++ b/projects/deno.land/package.yml
@@ -25,16 +25,6 @@ dependencies:
 
 build:
   script:
-    # deno does not need llvm to build on darwin+x86-64, and if it is present,
-    # deno will be linked to its libunwind, which then causes deno to need llvm
-    # in runtime.
-    - |
-      if test "{{hw.platform}}+{{ hw.arch }}" != "darwin+x86-64"; then
-        set -o allexport
-        source <(pkgx +llvm.org@17)
-        set +o allexport
-      fi
-
     - |
       rust_version=$(yq -er .toolchain.channel rust-toolchain.toml)
       if semverator lt "$rust_version" 1.67.0; then
@@ -67,6 +57,7 @@ build:
     - cargo install --locked --path cli --root "{{ prefix }}"
   dependencies:
     git-scm.org: 2 # to build tinycc
+    llvm.org: 17
     curl.se: '*' # required to download v8 (python is another option)
     cmake.org: ^3 # deno.land>=1.36.1 requires cmake
     protobuf.dev: '*' # deno.land>=1.36.4 requires protoc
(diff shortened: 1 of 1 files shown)


projects/deno.land/package.yml
```diff
@@ -18,12 +18,19 @@ runtime:
...
+#     llvm.org: 17 # libunwind
 
 build:
   script:
+    - |
+      if test "{{hw.platform}}+{{ hw.arch }}" = "darwin+aarch64"; then
+        set -a
+        source <(pkgx +llvm.org@17)
+        set +a
+      fi
```
jhheider: great if this works; should be able to be:
```yaml
  - run: |
      set -a
      source <(pkgx +llvm.org@17)
      set +a
    if: darwin/aarch64
```

felipecrs:
So far, the only evidence I found was on darwin+x86-64:

https://github.com/pkgxdev/pantry/actions/runs/8492104954/job/23264754342?pr=5751#step:6:36

But I wonder if this is not caused by the fact that the deno binary used in this environment was previously built with `llvm.org` installed since `llvm.org` was previously a build dependency for all platforms.

felipecrs:
Renewing the evidence that darwin+aarch64 indeed needs llvm.org to build:

https://github.com/pkgxdev/pantry/actions/runs/8492104954/job/23264754385?pr=5751#step:6:921

jhheider:
Possibly we could do something like:
```
  # link xcode linker to ld64.lld
  - run: |
      mkdir -p .bin
      ln -s /usr/bin/ld .bin/ld64.lld
      export PATH=$(pwd)/.bin:$PATH
    if: darwin
```

felipecrs:
> Possibly we could do something like:
> 
> ```
>   # link xcode linker to ld64.lld
>   - run: |
>       mkdir -p .bin
>       ln -s /usr/bin/ld .bin/ld64.lld
>       export PATH=$(pwd)/.bin:$PATH
>     if: darwin
> ```

https://github.com/pkgxdev/pantry/actions/runs/8492487518/job/23265547171?pr=5751#step:6:768

projects/deno.land/package.yml
```diff
@@ -19,11 +19,17 @@ runtime:
...
+  darwin/x86-64:
     llvm.org: 17 # libunwind
 
 build:
   script:
+    - run: |
+        set -o allexport
+        source <(pkgx +llvm.org@17)
+        set +o allexport
+      if: darwin/aarch64
```
jhheider: so, this should be a build-time dependency on llvm.org^17 for darwin, right? since we're introducing it here, and run-time for x86-64?

felipecrs: I'm still doing some tests, but I had to step away from the computer. Will resume later today.

jhheider: no rush; i always appreciate more eyes on this stuff. it's complicated, and it's more complicated when compiling compilers.

felipecrs: > so, this should be a build-time dependency on llvm.org^17 for darwin, right?

Building on darwin+aarch64 needs llvm otherwise: https://github.com/pkgxdev/pantry/actions/runs/8492104954/job/23264754385?pr=5751#step:6:921

> and run-time for x86-64?

Running on darwin+x86-64 needs llvm otherwise: https://github.com/pkgxdev/pantry/actions/runs/8492104954/job/23264754342?pr=5751#step:6:36

My theory however is that the deno used by brewkit only needs llvm in runtime for darwin+x86-64 because it had llvm set during build time.

Perhaps after this PR, deno will be rebuilt and it will no longer need llvm as runtime dependency on darwin+x86-64, but that's just a guess.

jhheider: It appears that deno will link in libunwind on x86-64 if present. So, yes, perhaps excluding it from build will solve it.

felipecrs:
I was able to reproduce this though https://github.com/pkgxdev/pantry/issues/3638 while building 1.37.2. I'll check what versions exactly need this environment variable.

felipecrs:
Deno's documentation is very clear that llvm is needed to compile on Linux (even though compilation passed without): https://docs.deno.com/runtime/manual/references/contributing/building_from_source

So, I brought it back. Except on darwin x86-64. Even the documentation does not mention llvm is needed there (only xcode).

jhheider:
Yes, we bring it in by default. The question is whether it links libunwind without explicit injection.

felipecrs:
> The question is whether it links libunwind without explicit injection.

I believe the CI tests proved already that this only happens on darwin+x86-64.

felipecrs:
Ok, I realized that deno 1.37.2 was failing to build not because I had reverted https://github.com/pkgxdev/pantry/pull/3639, but because of the Rust version.
Deno's build manual is explicit that the Rust version should be specific to their toolchain config:

https://docs.deno.com/runtime/manual/references/contributing/building_from_source#rust

felipecrs:
I tested compiling down to deno@1.20 and it's working (at least on Linux x64). :D

Probably some older versions as well.

projects/deno.land/package.yml
```diff
@@ -19,17 +19,41 @@ runtime:
...
   script:
+    # deno does not need llvm to build on darwin+x86-64, and if it is present,
+    # deno will be linked to its libunwind, which then causes deno to need llvm
+    # in runtime.
+    - |
+      if test "{{hw.platform}}+{{ hw.arch }}" != "darwin+x86-64"; then
+        set -o allexport
+        source <(pkgx +llvm.org@17)
+        set +o allexport
+      fi
```
jhheider: ok, but this makes llvm17 a build dep for everything but darwin+x86-64, and it's already brought in to darwin+x86-64 by the run-time requirement...

projects/deno.land/package.yml
```diff
@@ -62,6 +84,11 @@ test:
     - deno compile test.ts
     - test "$(./test)" = "Hello, world!"
     # human readable size of ./test should be equal to the size of denort to ensure our denort is being used
-    - test "$(stat -c %s ./test | numfmt --to=iec-i)" = "$(stat -c %s "{{prefix}}/bin/denort" | numfmt --to=iec-i)"
+    - |
+      if semverator gt {{version}} 1.40.4; then
+        test "$(stat -c %s ./test | numfmt --to=iec-i)" = "$(stat -c %s "{{prefix}}/bin/denort" | numfmt --to=iec-i)"
+      fi
```
jhheider: this can be simplified with:
```yaml
  - run: test "$(stat -c %s ./test | numfmt --to=iec-i)" = "$(stat -c %s "{{prefix}}/bin/denort" | numfmt --to=iec-i)"
    if: '>=1.40.5'
```

jhheider (commented):
i don't think this quite makes sense as written;

further, if you want to test for `libunwind` in the test phase, something like:
```yaml
  - run: ! (otool -l {{prefix}}/bin/deno | grep libunwind)
    if: darwin
  - run: ! (ldd {{prefix}}/bin/deno | grep libunwind)
    if: linux
```

projects/deno.land/package.yml
felipecrs: I didn't realize runtime requirements were added to build environment implicitly.

Anyway, if I remove the runtime dependency, the build won't even start on darwin+x86-64.

felipecrs: Let's see if tests passes now on darwin+x86-64 with the libunwind check.

felipecrs: Test is passing: https://github.com/pkgxdev/pantry/actions/runs/8502469906/job/23286685906?pr=5751#step:9:133

felipecrs: Ok, @jhheider would you like me to do anything else? Worst case it's better to have it only as a runtime dependency on darwin+x86-64 then on all platforms.

jhheider:
I'll try rebuilding the version we're using against this tomorrow and ensure it works. If we do it out of order, I believe it'll break since deno will still need libunwind but it won't be present.

felipecrs:
For some reason when adding rust later, llvm also needs to be added later.

jhheider:
alright, [deno~1.39](https://github.com/pkgxdev/pantry/actions/runs/8511342030) rebuilt, so we can try merging this.

PR #5752, opened 2024-03-30
Use DENO_INSTALL_ROOT to make deno install to ~/.local

felipecrs:
This dates back from Deno 0.42.0:

- https://github.com/denoland/deno/releases/tag/v0.42.0
- https://github.com/denoland/deno/pull/4787


felipecrs:
PS: this is doing the same as cargo:

https://github.com/pkgxdev/pantry/blob/1ebbfc1b7619ad89140373776bdc6b749574cf07/projects/rust-lang.org/cargo/package.yml#L63

jhheider:
@mxcl what do you think about this one? seems like a step back to use runtime environment (thereby requiring `pkgx`) over having the binary work as intended even if invoked directly?

felipecrs:
PS: I have no strong opinion, except by keeping consistency (meaning if we decide that patching is better, we should do it for cargo as well).

jhheider:
agree re: consistency. strongly.

mxcl:
I do feel we should patch. env vars are unfortunate control mechanisms since they are meant for the user to control.

PR #5753, opened 2024-03-30
+github.com/brucedom/bruce

tannevaled:


jhheider:
might need:
```yaml
companions:
  linux:
    systemd.io: '*'
```

jhheider:
thanks, @tannevaled !

tannevaled:
> might need:
> 
> ```yaml
> companions:
>   linux:
>     systemd.io: '*'
> ```

is there some docs or samples explaining this point?

jhheider:
it probably needs to be added to the wiki; the short answer is a dependency is required for the binary to run, which a companion is used during normal operation of the tool. so, in this instance, the binary will execute without systemctl present, but it expects to find it for its use on linux.

PR #5755, opened 2024-03-31
+ocaml.org

isworking:
Adds a recipe for [OCaml](https://ocaml.org) WITHOUT the Compiler Libraries, which are often unnecessary for everything but Opam. For Opam, we need that, btw.


isworking:
@jhheider @mxcl Any idea what's wrong? It builds and tests just fine locally. And seems like the bug isn't related to the recipe itself. So, I'm clueless.

jhheider:
It can't find ocaml. Are you sure there's a bin/ocaml at the end, and it's not a broken symlink? Adding an ls should show it.

isworking:
> It can't find ocaml. Are you sure there's a bin/ocaml at the end, and it's not a broken symlink? Adding an ls should show it.

Yep, that's the problem. OCaml has so many binaries, that I just copy-pasted all the bin names, and added a `-` at the beginning. But I forgot to add the `bin/` part. `bk test` worked because I already had OCaml built before with PKGX (I just added `bin/ocaml` and `bin/ocamlc` as provides back then).

jhheider:
Problem was prefix in the build twice:
`/opt/ocaml.org/v5.1.1+brewing/opt/ocaml.org/v5.1.1+brewing/lib/ocaml/compiler-libs`

jhheider:
it looks like some of the bins are probably shellscripts that need surgery for relocatability.

jhheider:
this fixes the file-not-found issue:
```yaml
    - run: |
        LENGTH=$[$(head -n1 ocaml | wc -c)-1]
        NEW_SHEBANG=$(printf "%-${LENGTH}s" "#!/usr/bin/env ocamlrun")
        for BIN in $SHEBANGD; do
          sed -i "1s|.*|$NEW_SHEBANG|" $BIN
        done
      working-directory: '{{prefix}}/bin'
  env:
    SHEBANGD:
      - ocaml
      - ocamlc.byte
      - ocamlcmt
      - ocamlcp
      - ocamldebug
      - ocamldep.byte
      - ocamldoc
      - ocamllex.byte
      - ocamlmklib
      - ocamlmktop
      - ocamlobjinfo.byte
      - ocamlopt.byte
      - ocamloptp
      - ocamlprof
```
but it still has relocatability issues. Homebrew's recipe suggests build paths are baked into the output, and https://ocaml.org/docs/arm64-fix suggests the only fix is to match install path to build path. if that's true, then this isn't going to work without potentially massive changes to the underlying binaries.

jhheider:
superceded by #13542

PR #5853, opened 2024-04-10
+android.com/cmdline-tools

and-ri:
A very convenient tool for developing Android applications. Especially for installing the SDK without using Android Studio


and-ri:
`SDK manager package "emulator" is missing on ARM`
https://stackoverflow.com/questions/71545278

I need to do another test 🤔

PR #5857, opened 2024-04-10
Build jfrog-cli for linux

felipecrs:


jhheider:
that look good to you, @felipecrs ? usually we catch quite early that linux go projects need `-buildmode=pie`.

felipecrs:
Looks great! Thanks a ton!

PR #5867, opened 2024-04-12
fix(openjdk)

and-ri:
I get this error when building an app with `eas-cli`

```bash
~/.pkgx/openjdk.org/v20.0.2.9/include/jni.h:45:10: fatal error: 'jni_md.h' file not found
```


jhheider:
i'll rebuild all the versions: https://github.com/pkgxdev/pantry/actions/runs/8665318390/job/23763737631

PR #6042, opened 2024-05-02
+hyfetch

### LINKED ISSUE #6038: +github.com/hykilpikonna/hyfetch
[neofetch](https://github.com/dylanaraps/neofetch) is dead: https://www.omgubuntu.co.uk/2024/04/neofetch-system-info-tool-is-dead

hyfetch fortunately continues it.

https://github.com/hykilpikonna/hyfetch

hyfetch provides the `hyfetch` command and also `neowofetch` to mimic `neofetch`'s behavior.

Maybe in the future we may also consider providing `neowofetch` by default when users call `neofetch`.

and-ri:
closes #6038

### DIFF
--- a/projects/github.com/hykilpikonna/hyfetch/package.yml
+++ b/projects/github.com/hykilpikonna/hyfetch/package.yml
@@ -1,0 +1,42 @@
+distributable:
+  url: https://github.com/hykilpikonna/hyfetch/archive/refs/tags/{{version.tag}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: hykilpikonna/hyfetch
+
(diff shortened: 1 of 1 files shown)


felipecrs:
Just some idea, hyfetch itself is indeed Python-based. But the neowofetch program itself is just a bash script with almost no dependencies.

Perhaps it would be a good idea to create a different package like `github.com/hykilpikonna/hyfetch/neowofetch` and remove `bin/neowofetch` from this one, so that people running `pkgx neowofetch` can execute for the first time a lot quicker without having to download python.

jhheider:
a fine idea

PR #6046, opened 2024-05-02
fix(curve-dalek.rust)

jhheider:
with the release of rust 1.78 we can undo these... tweaks.


jhheider:
note that deno still uses 1.77.2, but it requires this tweak. once they move beyond it, we can remove that.

PR #6098, opened 2024-05-09
fix(werf): remove runtime dependencies

felipecrs:
They were surely taken from Homebrew's formulae, but I'm not sure if they are really required for anything. Downloading the werf binaries from the GitHub Releases also works just fine, so I'm trying to remove them. Let's see what CI tests says.

They make installing werf take way longer than otherwise.


felipecrs:
Hm... I'll probably need to tweak some of these flags now.

PR #6160, opened 2024-05-16
+cowsay

nakajima:
Adds cowsay.

### DIFF
--- a/projects/github.com/cowsay-org/cowsay/package.yml
+++ b/projects/github.com/cowsay-org/cowsay/package.yml
@@ -11,6 +11,9 @@ platforms:
   - darwin
   - linux/x86-64
 
+dependencies:
+  perl.org: '*'
+
 build:
   script:
     - run: cpan File::Find
--- a/projects/github.com/cowsay-org/cowsay/package.yml
+++ b/projects/github.com/cowsay-org/cowsay/package.yml
@@ -1,5 +1,5 @@
 distributable:
-  url: https://github.com/cowsay-org/cowsay/archive/refs/tags/v{{version}}.tar.gz
+  url: https://github.com/cowsay-org/cowsay/archive/refs/tags/{{version.tag}}.tar.gz
   strip-components: 1
 
 display-name: cowsay
@@ -7,10 +7,9 @@ display-name: cowsay
 versions:
   github: cowsay-org/cowsay
 
-dependencies:
-  cpanmin.us: '*'
-
 build:
+  dependencies:
+    cpanmin.us: '*'
   script:
     - run: cpanm -l {{prefix}} File::Find
       if: linux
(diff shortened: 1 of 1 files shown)


projects/github.com/cowsay-org/cowsay/package.yml
```diff
@@ -0,0 +1,29 @@
...
+  ref: ${{version.tag}}
+
+display-name: cowsay
+
+versions:
+  github: cowsay-org/cowsay
+
+platforms:
+  - darwin
+  - linux/x86-64
```
jhheider: is this intentional? we should either have a comment as to why it doesn't work on linux/aarch64, or remove this stanza if it does.

projects/github.com/cowsay-org/cowsay/package.yml
```diff
@@ -0,0 +1,29 @@
...
+
+dependencies:
+  perl.org: '*'
+
+build:
+  script:
+    - run: cpan File::Find
```
jhheider: we should use `cpanmin` to install it in the prefix. see https://github.com/pkgxdev/pantry/blob/main/projects/gnu.org/help2man/package.yml for example.

projects/github.com/cowsay-org/cowsay/package.yml
```diff
@@ -0,0 +1,29 @@
+distributable:
+  url: git+https://github.com/cowsay-org/cowsay.git
+  ref: ${{version.tag}}
```
jhheider: is it the case that using the packaged sources doesn't work? i'm fine if that's the case, but i'd be surprised for a perl package.

projects/github.com/cowsay-org/cowsay/package.yml
```diff
@@ -0,0 +1,29 @@
...
+  perl.org: '*'
+
+build:
+  script:
+    - run: cpan File::Find
+      if: linux
+    - make install prefix={{prefix}}
+
+provides:
+  - cowsay
```
jhheider: this should be `- bin/cowsay`

jhheider (changes requested):

projects/github.com/cowsay-org/cowsay/package.yml
nakajima: Nope, updated.

projects/github.com/cowsay-org/cowsay/package.yml
nakajima: Nope, I think that's just what was there in the generated file.

nakajima:
@jhheider Thanks for the review, it's my first PR here so I still don't quite have the lay of the land. I think I took care of everything!

jhheider (approved):
looks good!

PR #6306, opened 2024-06-03
Fix argbash

felipecrs:
First it was failing because missing `autom4te`.

Second it was failing with:

```
The directory '//opt/argbash.dev/v2.10.0+brewing/lib/argbash' with files needed by Argbash is not browsable. Check whether it exists and review it's permissions.
```

This is because of:

https://github.com/matejak/argbash/blob/0d046ad6a208a8ba2820fa7bdd13d21eb211e51e/resources/Makefile#L171

So I believe our best option is not use their `Makefile` for now. Fortunately, things are simple enough.

I also added a test to cover this.

### DIFF
--- a/projects/argbash.dev/package.yml
+++ b/projects/argbash.dev/package.yml
@@ -6,22 +6,23 @@ versions:
   github: matejak/argbash
 
 dependencies:
-  gnu.org/bash: '>=3.0'
+  gnu.org/bash: '>=3'
+  gnu.org/autoconf: '*'
 
 build:
-  dependencies:
-    gnu.org/autoconf: '*'
   script:
-    - make install PREFIX={{prefix}}
-  working-directory: resources
-  
+    - mkdir -p '{{prefix}}'
+    - cp -r bin src '{{prefix}}'
+
 provides:
   - bin/argbash
   - bin/argbash-init
   - bin/argbash-1to2
 
 test:
-  - argbash-init --pos positional-arg --opt option --opt-bool print minimal.m4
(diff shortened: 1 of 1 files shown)


jhheider:
Great catch! Thanks!

felipecrs:
@jhheider would you please trigger a re-build of argbash to include this fix? It looks like they will never release a new version.

jhheider:
It looks like the version we have (2.10.0) was built by this merge. So, you might need to remove your local version and reinstall, but this looks correct:
```shell
ls -R ~/.pkgx/argbash.dev/v2.10.0/
bin src

/Users/jacob/.pkgx/argbash.dev/v2.10.0//bin:
argbash      argbash-1to2 argbash-init

/Users/jacob/.pkgx/argbash.dev/v2.10.0//src:
argbash-1to2.m4              constants.m4                 output-bash-script.m4        output-strip-all.m4          value_validators.m4
argbash-init.m4              default_settings.m4          output-completion.m4         output-strip-none.m4         version
argbash-lib.m4               docopt.m4                    output-docopt.m4             output-strip-user-content.m4
argbash.m4                   env_vars.m4                  output-manpage-defs.m4       progs.m4
argument_value_types.m4      function_generators.m4       output-manpage.m4            stuff.m4
collectors.m4                list.m4                      output-posix-script.m4       utilities.m4
```

felipecrs:
You're right. Thank you. The following did the trick:

```
rm -rf ~/.cache/pkgx ~/.local/share/pkgx ~/.pkgx
```

PR #6377, opened 2024-06-09
chore(oapi-codegen): update repository location

mycrEEpy:
oapi-codegen is now in it's own org, decupled from deepmap.
Don't know if we should leave the name in the pantry as github.com/deepmap/oapi-codegen?


jhheider:
in the past we've handled renames by deploying at the new name and removing the `provides:` at the old name, but keeping the package. i added that.

PR #6484, opened 2024-06-21
+ctop

sanchitram1:
`ctop -v` isn't giving me the desired result, but the package works as expected

UPDATE: I checked the source installation file for ctop, and it included a Makefile that performed the two steps I've specified in the script. You need to tell ctop what its current sha and version are when triggering the build.


projects/github.com/bcicen/ctop/package.yml
```diff
@@ -0,0 +1,36 @@
...
+    go.dev: ^1.18
+  script: |
+    go mod download
+    go build -v -trimpath -ldflags="$LDFLAGS" -o $BUILDLOC
+  env:
+    BUILDLOC: '{{prefix}}/bin/ctop'
+    LDFLAGS:
+      - -s
+      - -w
+      - -X=main.Version={{version}}
```
jhheider: lowercase `version`: https://github.com/bcicen/ctop/blob/59f00dd6aaebf48f3bc501e174f75f815c2619f0/Makefile#L4

projects/github.com/bcicen/ctop/package.yml
```diff
@@ -0,0 +1,36 @@
...
+
+versions:
+  github: bcicen/ctop
+
+build:
+  dependencies:
+    go.dev: ^1.18
+  script: |
+    go mod download
+    go build -v -trimpath -ldflags="$LDFLAGS" -o $BUILDLOC
```
jhheider: in general, i've been using `GO_LDFLAGS`, in case go triggers a C-compiler which gets confused by weird `LDFLAGS` variables.

projects/github.com/bcicen/ctop/package.yml
sanchitram1: fixed

projects/github.com/bcicen/ctop/package.yml
sanchitram1: ✅

projects/github.com/bcicen/ctop/package.yml
```diff
@@ -0,0 +1,31 @@
...
+  github: bcicen/ctop
+
+build:
+  dependencies:
+    go.dev: ^1.18
+    git-scm.org: "*"
+
+  script: |
+    go mod download
+    CGO_ENABLED="$CGO_ENABLED" go build -tags release -ldflags="$GO_LDFLAGS" -o "$BUILDLOC"
```
jhheider: `CGO_ENABLED="$CGO_ENABLED"` shouldn't do anything, unless something is very weird in the shell.

projects/github.com/bcicen/ctop/package.yml
```diff
@@ -0,0 +1,31 @@
...
+  env:
+    CGO_ENABLED: 0
+    BUILDLOC: "{{prefix}}/bin/ctop"
+    BUILD: "$(git rev-parse --short HEAD)"
```
jhheider: Technically, this isn't _really_ right, since it'll be the non commit in the fake git checkout. Better to just set it to `pkgx`

projects/github.com/bcicen/ctop/package.yml
```diff
@@ -0,0 +1,30 @@
+distributable:
```
jhheider: package name should be `ctop.sh` (the homepage)

projects/github.com/bcicen/ctop/package.yml
```diff
@@ -0,0 +1,30 @@
...
+  strip-components: 1
+
+display-name: ctop
+
+versions:
+  github: bcicen/ctop
+
+build:
+  dependencies:
+    go.dev: ^1.18
```
jhheider: go is kind of picky about versions; generally using `~1.18` is safer

projects/github.com/bcicen/ctop/package.yml
```diff
@@ -0,0 +1,30 @@
...
+    git-scm.org: "*"
```
jhheider: obsolete, right?

projects/github.com/bcicen/ctop/package.yml
```diff
@@ -0,0 +1,30 @@
...
+    go mod download
+    go build -tags release -ldflags="$GO_LDFLAGS" -o "$BUILDLOC"
+  env:
+    BUILDLOC: "{{prefix}}/bin/ctop"
+    BUILD: "pkgx"
+    GO_LDFLAGS:
+      - -s
+      - -w
+      - -X main.version={{version}}
+      - -X main.build="$BUILD"
```
jhheider: probably don't need an intermediate variable

projects/github.com/bcicen/ctop/package.yml
```diff
@@ -0,0 +1,30 @@
...
+
+versions:
+  github: bcicen/ctop
+
+build:
+  dependencies:
+    go.dev: ^1.18
+    git-scm.org: "*"
+
+  script: |
```
jhheider: i usually always use script arrays these days, since they're easier to alter and tweak, but this is just a suggestion

projects/github.com/bcicen/ctop/package.yml
```diff
@@ -0,0 +1,30 @@
...
+    GO_LDFLAGS:
+      - -s
+      - -w
+      - -X main.version={{version}}
+      - -X main.build="$BUILD"
+
+provides:
+  - bin/ctop
+
+test: test "$(ctop -v | awk '{print $3}')" = {{version}},
```
jhheider: linux is segfaulting; add:
```yaml
build:
...
  env:
    linux:
      GO_LDFLAGS:
        - -buildmode=pie
```

projects/github.com/bcicen/ctop/package.yml
sanchitram1: can comment on script arrays!

projects/github.com/bcicen/ctop/package.yml
```diff
@@ -0,0 +1,31 @@
+distributable:
+  url: https://github.com/bcicen/ctop/archive/refs/tags/v{{version}}.tar.gz
+  strip-components: 1
+
+display-name: ctop.sh
```
jhheider: I didn't mean this. I meant the actual name of the package should be ctop.sh.

sanchitram1: so, bin/ctop.sh?

sanchitram1: if so, I'd disagree, since it's conventionally called using `ctop`, and we'd want to preserve that, no? your original comment was on the distributable line -- did you mean that we should be fetching the source from ctop.sh as well?

jhheider: no, `projects/ctop.sh/package.yml`

jhheider (approved):
LG!

PR #6518, opened 2024-06-26
Create 0xMinister

therealminister:
distributable:
  url: https://github.com/pinterest/minister/releases/download/{{version}}/minister-{{version}}.zip

versions:
  github: pinterest/minister

warnings:
  - vendored

dependencies:
  openjdk.org: '*'

build:
  - install -D minister-{{version}}/bin/minister {{prefix}}/bin/minister

provides:
  - bin/minister

test:
  - minister --version | grep {{version}}
  - echo 'fun main(   )  {}' > Main.kt
  - minister -F Main.kt
  - cat Main.kt | grep 'fun main() {}'

### DIFF
--- a/projects/0xMinister
+++ b/projects/0xMinister
@@ -1,0 +1,23 @@
+distributable:
+  url: https://github.com/pinterest/minister/releases/download/{{version}}/minister-{{version}}.zip
+
+versions:
+  github: pinterest/minister
+
+warnings:
+  - vendored
+
+dependencies:
+  openjdk.org: '*'
+
+build:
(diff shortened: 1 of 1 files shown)


jhheider:
I can't see what this is (the github 404s: https://github.com/pinterest/minister).

CI couldn't check it because the project is improperly named. It should either be `projects/github.com/pinterest/minister/package.yml`, or some more specific URL if the project has a real home page.

Further, this looks like it's just the package.yml for `ktlint` with the project name replaced by `minister`. So, unfortunately, this looks like a spam submission. I'm going to close this at this time. If there's a real project you'd like to package, obviously we can try again.

PR #6596, opened 2024-07-02
fix(dagger): correctly inject version number during build

leonardoce:
The version number was not properly injected, resulting in a failure when starting the engine.


leonardoce:
See: https://github.com/dagger/dagger/blob/7900eb5ea40dc52602d76d9414afc06ae287ef48/.goreleaser.common.yml#L14C9-L14C52

jhheider:
great! thanks for this!

PR #6607, opened 2024-07-02
+sqruff

sanchitram1:
wasn't sure about the rustup tool chain bit, but I saw it in another crate on pkgx, and thought I might need to include it. however, it worked without the additional lines as wlel


jhheider:
will it build with rust 1.79? nightly is a last resort.

jhheider:
looks like yes!

sanchitram1:
that's a far better test than the version check

jhheider:
yeah, version check is last resort as well, when no other reasonably simple test is possible. lots of things that use curses, for example, or complicated libraries, can be hard to test.

PR #6609, opened 2024-07-03
+wavpack.com

ivabus:
Adding wavpack codec utilities (wavpack, wvunpack, wvtag, wvgain) may be useful for automation.


jhheider:
looks great! thanks for the contribution!

PR #6621, opened 2024-07-04
+aux4.io/aux4

dgouvea:
Add package for CLI generator aux4

https://aux4.io

### DIFF
--- a/projects/aux4.io/package.yml
+++ b/projects/aux4.io/package.yml
@@ -1,11 +1,10 @@
+warnings:
+  - vendored
+
 display-name: aux4
 
 versions:
-  github: aux4/aux4/releases/tags
-
-platforms:
-  - darwin
-  - linux/x86-64
+  github: aux4/aux4
 
 build:
   dependencies:
@@ -19,7 +18,7 @@ build:
  
   env:
     linux/aarch64:
-      DISTRIBUTABLE_URL: 'https://github.com/aux4/aux4/releases/download/v{{version}}/aux4-linux-amd64'
+      DISTRIBUTABLE_URL: 'https://github.com/aux4/aux4/releases/download/v{{version}}/aux4-linux-arm64'
     linux/x86-64:
       DISTRIBUTABLE_URL: 'https://github.com/aux4/aux4/releases/download/v{{version}}/aux4-linux-amd64'
     darwin/aarch64:
@@ -29,7 +28,7 @@ build:
 
 test:
   script:
-    - aux4 aux4 version
+    - aux4 aux4 version | grep {{version}}
 
 provides:
   - bin/aux4
--- a/projects/aux4.io/package.yml
+++ b/projects/aux4.io/package.yml
@@ -1,34 +1,52 @@
-warnings:
-  - vendored
-
-display-name: aux4
+distributable:
+  url: https://github.com/aux4/aux4/archive/refs/tags/{{version.tag}}.tar.gz
+  strip-components: 1
 
 versions:
   github: aux4/aux4
 
 build:
   dependencies:
-    curl.se: '*'
-
-  working-directory: ${{prefix}}
-  script:
-    - 'mkdir bin'
-    - 'curl -L $DISTRIBUTABLE_URL -o "{{prefix}}"/bin/aux4'
-    - 'chmod +x "{{prefix}}"/bin/aux4'
- 
+    go.dev: ~1.21.5
+  script: go build $ARGS -ldflags="$GO_LDFLAGS" .
   env:
-    linux/aarch64:
-      DISTRIBUTABLE_URL: 'https://github.com/aux4/aux4/releases/download/v{{version}}/aux4-linux-arm64'
-    linux/x86-64:
-      DISTRIBUTABLE_URL: 'https://github.com/aux4/aux4/releases/download/v{{version}}/aux4-linux-amd64'
-    darwin/aarch64:
-      DISTRIBUTABLE_URL: 'https://github.com/aux4/aux4/releases/download/v{{version}}/aux4-darwin-arm64'
-    darwin/x86-64:
-      DISTRIBUTABLE_URL: 'https://github.com/aux4/aux4/releases/download/v{{version}}/aux4-darwin-amd64'
+    GO_LDFLAGS:
+      - -s
+      - -w
+      - -X main.version={{version}}
+    linux:
+      GO_LDFLAGS:
+        - -buildmode=pie
+    ARGS:
+      - -v
+      - -trimpath
+      - -o "{{prefix}}/bin/aux4"
 
 test:
+  fixture: |
+    {
+      "profiles": [
+        {
+          "name": "main",
+          "commands": [
+            {
+              "name": "hello",
+              "execute": [
+                "echo 'Hello World'"
+              ],
+              "help": {
+                "text": "say hello world"
+              }
+            }
+          ]
+        }
+      ]
+    }
   script:
     - aux4 aux4 version | grep {{version}}
+    - cp $FIXTURE .aux4
+    - aux4
+    - test "$(aux4 hello)" = "Hello World"
 
 provides:
   - bin/aux4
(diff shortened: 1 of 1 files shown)


projects/aux4.io/package.yml
```diff
@@ -0,0 +1,35 @@
+display-name: aux4
+
+versions:
+  github: aux4/aux4/releases/tags
+
+platforms:
```
jhheider: is this intentional, or just left in? most go will build for linux/aarch64 fine.

projects/aux4.io/package.yml
```diff
@@ -0,0 +1,35 @@
...
+ 
+  env:
+    linux/aarch64:
+      DISTRIBUTABLE_URL: 'https://github.com/aux4/aux4/releases/download/v{{version}}/aux4-linux-amd64'
+    linux/x86-64:
+      DISTRIBUTABLE_URL: 'https://github.com/aux4/aux4/releases/download/v{{version}}/aux4-linux-amd64'
+    darwin/aarch64:
+      DISTRIBUTABLE_URL: 'https://github.com/aux4/aux4/releases/download/v{{version}}/aux4-darwin-arm64'
+    darwin/x86-64:
+      DISTRIBUTABLE_URL: 'https://github.com/aux4/aux4/releases/download/v{{version}}/aux4-darwin-amd64'
```
jhheider: can we build from source? that's always preferred. if not, we should add a top-level
```yaml
warnings:
  - vendored
```

i suppose this is why linux/aarch64 is omitted, but it will probably build correctly. look at `crazymax.dev/diun` for a recent go project.

projects/aux4.io/package.yml
```diff
@@ -0,0 +1,35 @@
...
+
+test:
+  script:
+    - aux4 aux4 version
```
jhheider: if possible, we should test functionality here. if there's no good way, confirming version is an ok fallback.

projects/aux4.io/package.yml
```diff
@@ -0,0 +1,35 @@
+display-name: aux4
+
+versions:
+  github: aux4/aux4/releases/tags
```
jhheider: `aux4/aux4/releases/tags` is the same as `aux4/aux4`.

jhheider (changes requested):

jhheider:
just as a note: https://github.com/pkgxdev/pantry/actions/runs/9742005002/job/27018828297 will probably be tying up the darwin+x86-64 builder for another 2ish hours. so, we won't see that build output until then (but it's nearly always the same as darwin+aarch64).

projects/aux4.io/package.yml
dgouvea: i removed that as it can build for all platforms

projects/aux4.io/package.yml
dgouvea: i included that

projects/aux4.io/package.yml
dgouvea: i added a check if the version matches

projects/aux4.io/package.yml
dgouvea: fixed

dgouvea:
@jhheider thanks for your comments, they were very helpful. it's okay if takes longer to build. i am more concerned if everything is properly configured.

jhheider:
thank *you* for your contribution. always happy to have new contributors.

jhheider:
aux4 looks _really cool_. i'm a big fan of `just` and similar tools.

btw, i think what we want is probably:

```yaml
distributable:
  url: https://github.com/aux4/aux4/archive/refs/tags/{{version.tag}}.tar.gz
  strip-components: 1

versions:
  github: aux4/aux4

build:
  dependencies:
    go.dev: ~1.21.5
  script: go build $ARGS -ldflags="$GO_LDFLAGS" .
  env:
    GO_LDFLAGS:
      - -s
      - -w
      - -X main.version={{version}}
    linux:
      GO_LDFLAGS:
        - -buildmode=pie
    ARGS:
      - -v
      - -trimpath
      - -o "{{prefix}}/bin/aux4"

test:
  fixture: |
    {
      "profiles": [
        {
          "name": "main",
          "commands": [
            {
              "name": "hello",
              "execute": [
                "echo 'Hello World'"
              ],
              "help": {
                "text": "say hello world"
              }
            }
          ]
        }
      ]
    }
  script:
    - aux4 aux4 version | grep {{version}}
    - cp $FIXTURE .aux4
    - aux4
    - test "$(aux4 hello)" = "Hello World"

provides:
  - bin/aux4
```

dgouvea:
thank you again @jhheider! i am glad to know you found aux4 interesting. i hope it might be useful for your daily tasks.

i accepted your changes. i tested them look great. however, i am not a big fan of building from source because I always have to "maintain" the build in multiple places. i think my repository should be the source of truth for that, and the distribution should rely on that.

but again, thanks a lot for your contribution! huge help!

jhheider:
> i am not a big fan of building from source because I always have to "maintain" the build in multiple places. i think my repository should be the source of truth for that, and the distribution should rely on that.

this is a perfectly valid concern. i'm usually on top of build failures on our infrastructure, so you've got backup at least. in general, we try to build from source by default.

of course, once `aux4` is built, it could depend on itself, if you wanted more external control. `go` programs tend to be easy to build, but people sometimes change e.g. the path to the `version` variable, and we need to add that to the recipe. this is where things like makefiles or rust's (i'm a big rust fan) cargo + build.rs scheme comes in handy. the build command stays the same. so, if you like using your own tooling, `aux4 install --prefix={{prefix}}` looks like the `aux4` way. and you retain control.

you have to accept the circular dependency, but after we have one version, that's not a big deal. we have a small number of things (mpdecimal, llvm, and a couple others) that have circular bootstrapping requirements.

jhheider (approved):

dgouvea:
i am going to leave as it's for now. i may change at some point.

i love how pkgx works! it's amazing how it finds the dependencies and temporarily installs them to use.

i think aux4 combined with that will be a very powerful tool.

thanks for approving my PR.

jhheider:
thanks for submitting it! i was looking through all your documentation and thinking about how it all works. obviously, you can just execute pkgx from aux4 if you have it; i was thinking if there was any sense in it getting it's own executor. there is potential for a `math:` executor, if you want to do high-end math without `let` or `bc`. another possibility i considered was a `url:` executor that looks for curl/wget/aria2/etc to handle http operations directly.

there's so many options to simplify workflows by wrapping common tasks in executors. it's an exciting project.

dgouvea:
This is probably not the right place for this conversation, but it's a great discussion. Thank you so much for sharing your thoughts.

Let's consider this scenario: you have to download a CSV file from an S3 bucket, parse it, make some adjustments, and insert the data into your database.

You can create a command on `aux4` to orchestrate everything. But to get the file from S3, you will have to have the `aws` CLI installed. Usually, the hack I use is to create a command called `setup,` where the user can run once to install the dependencies.

This is where pkgx comes into place. If I run that integrated, I can let `pkgx` know and manage the dependencies.

But the most valuable thing from `aux4` is not in its documentation yet. I have tens of modules (CLI tools) to integrate with `aux4`. Most are [npm packages](https://www.npmjs.com/search?q=@aux4/).

For example, I can use the [adapter](https://www.npmjs.com/package/@aux4/adapter) to parse and transform the CSV into any JSON structure I need. Then, I can use [db](https://www.npmjs.com/package/@aux4/db) to run the query to insert the data into the database.

I am working on that right now. I am studying how to use `pkgx` to make it effortless for the users.

dgouvea:
thanks for merging that!

jhheider:
agree that this is more of an aux4 discussion, but, man, that's all really cool.

so, taken as read the caveat about space usage for pkgx, what might make the most sense for your initial integration is the `super portable scripts` block from https://docs.pkgx.sh/run-anywhere/scripts. that'll keep everything isolated to $TMPDIR if the user doesn't otherwise have `pkgx` installed. then you really can integrate almost anything (and "pkgx: clang ${filename} -o ${exename}" becomes trivial, except for downloading `llvm.org`).

PR #7226, opened 2024-09-10
fix(watchman)

Dunedubby:


Dunedubby:
Judging from the logs, we may need to update folly to a more recent version. The correct version (2024.09.09.00) has not yet been released on pkgx.

jhheider:
yeah, they update every week, but have to be built in a specific order. i'll get them finished tomorrow. though watchman hasn't built for our linux targets in a few months.

Dunedubby:
Gotcha. Thanks so much for maintaining these. By the way, what's the best
way to set the build-time WATCHMAN_STATE_DIR so it points to the .pkgx
directory?

jhheider:
that should probably be an `runtime.env.WATCHMAM_STATE_DIR: ${{prefix}}/var` unless there's a reason we shouldn't hijack it.

projects/facebook.com/watchman/package.yml
```diff
@@ -46,7 +46,7 @@ build:
     WATCHMAN_STATE_DIR: "{{prefix}}/var/run/watchman"
     CMAKE_ARGS:
       - -DCMAKE_INSTALL_PREFIX="{{prefix}}
-      - -DCMAKE_PREFIX_PATH=$(echo $LIBRARY_PATH | sed 's/\/lib:/:/g; s/\/lib$//')
+      - -DCMAKE_PREFIX_PATH=$(echo $LIBRARY_PATH | sed 's/\/lib:/;/g; s/\/lib$//')
```
jhheider: CMAKE_PREFIX_PATH should be fixed in pkgx 1.2.1 again. sorry for the breakage!

Dunedubby: Ah okay haha I was wondering what was happening...

PR #7229, opened 2024-09-11
+moonrepo.dev/moon

iyefrat:
adapted from https://formulae.brew.sh/formula/moon and https://github.com/pkgxdev/pantry/blob/main/projects/crates.io/bat/package.yml


jhheider:
thanks for the contribution, @iyefrat !

PR #7274, opened 2024-09-13
+x.org/libxres

sebst:
https://gitlab.freedesktop.org/xorg/lib/libxres

`libXres` is packaged in brew and needed as dependency for several other x11 related packages


jhheider:
thanks, @sebst !

Issue #7290, opened 2024-09-15
cairographics.org/pycairo does not create `pycairo.pc` for `pkg-config`


sebst:
'pkg-config --libs py3cairo' failed

sebst:
see: https://packages.debian.org/buster/all/python3-cairo-dev/filelist

sebst:
see: https://github.com/pygobject/pycairo/issues/270

sebst:
See: https://github.com/pygobject/pycairo/issues/270

PR #7291, opened 2024-09-15
+xpra.org

sebst:
Blocked by #7290


sebst:
#7292

jhheider:
looks like this had some kind of issue finding cython.

PR #7294, opened 2024-09-15
Pycairo: Create `py3cairo.pc` file for pkg-config (wip)

sebst:
Related: #7290


jhheider:
pycairo doesn't appear to ship include/lib files (at least the one we have); it might need different compile files.

jhheider:
handled in #7366. thanks for catching this.

PR #7361, opened 2024-09-20
+kafka.apache.org

### LINKED ISSUE #653: +kafka (387/548)
Formulae: https://formulae.brew.sh/formula/kafka

sebst:


sebst:
- [x] Tick kafka in #99

projects/kafka.apache.org/package.yml
```diff
@@ -0,0 +1,73 @@
...
+build:
+  dependencies:
+    rsync.samba.org: '*'
+  script:
+    - rm -rf bin/windows
+    - sed -i 's|log.dirs=/tmp/kafka-logs|log.dirs={{prefix}}/kafka-logs|' config/server.properties
+    - sed -i 's|log.dirs=/tmp/kraft-combined-logs|log.dirs={{prefix}}/kraft-combined-logs|' config/server.properties
+    - sed -i 's|log.dirs=/tmp/kraft-controller-logs|log.dirs={{prefix}}/kraft-controller-logs|' config/controller.properties || true
+    - sed -i 's|log.dirs=/tmp/kraft-broker-logs|log.dirs={{prefix}}/kraft-broker-logs|' config/broker.properties || true
+    - sed -i 's|dataDir=/tmp/zookeeper|log.dirs={{prefix}}/zookeper|' config/zookeeper.properties
```
jhheider: this probably needs some work to be relocatable. `{{prefix}}` on the build machine won't exist on install machines. if this can be controlled by runtime.env vars (unlikely) we should do that, otherwise, we'll need to replace the scripts with a shim that fixes paths before launching. see https://github.com/pkgxdev/brewkit/blob/main/share/brewkit/python-venv-stub.py for the most complicated version.

sebst: Wouldn't it be easier to have variable {{installed_prefix}}? That would need some changes in brewkit, but could be worth the effort? Or do I miss a conceptual point here?

jhheider: Yeah, there's no way to know the installed prefix, since it depends on where the user installs pkgx. Ideally, programs are runnable directly without pkgx once installed (not always true, but often true). Hence using a shim, so you can correct at runtime. Other tricks abound, such as using $(dirname $(dirname $0)) in shell scripts.

jhheider: actually, looking at what you're changing, it's probably fine to just remove the sed commands. using /tmp is probably optimal.

sebst: I see, makes sense. Will have a look later

jhheider: we'll try it as /tmp for now, unless that breaks anything.

PR #7419, opened 2024-09-26
+smallstep.com/certificates

sebst:


jhheider:
this is `go` and should be easy to build. was there some blocker that required vendoring?

sebst:
> this is `go` and should be easy to build. was there some blocker that required vendoring?

It's notarized and for such security relevant package it's probably more trustworthy to use the vendored packages

jhheider:
that's an interesting argument. we've built everything else in the stack in the open, with fully-transparent builds, and properly signed apple binaries, as do most other packagers. homebrew, for example, builds it here: https://github.com/Homebrew/homebrew-core/blob/f45ca9f767b050e0f9a16a155ba0d50baf49e865/Formula/s/step.rb#L32-L37. i'll think about this a little more.

sebst:
> that's an interesting argument. we've built everything else in the stack in the open, with fully-transparent builds, and properly signed apple binaries, as do most other packagers. homebrew, for example, builds it here: https://github.com/Homebrew/homebrew-core/blob/f45ca9f767b050e0f9a16a155ba0d50baf49e865/Formula/s/step.rb#L32-L37. i'll think about this a little more.

Let me know what you think. Should be relatively easy to build it from scratch

sebst:
Any new thoughts?

jhheider:
yeah, i think we should probably maintain consistency. if someone is concerned, we can always provide a vendored version, or, more likely, they get the "official" binary themselves.

PR #7420, opened 2024-09-26
+smallstep.com/cli

sebst:


jhheider:
this looks easy enough to build; was there some reason it's vendored?

https://github.com/Homebrew/homebrew-core/blob/5276d35fcd730351c64c32d51a0977e326034077/Formula/s/step.rb

PR #7436, opened 2024-09-27
Fix/wezterm

sebst:


jhheider:
are these supposed to be symlinks, or are there other binaries that need to be built?

sebst:
> are these supposed to be symlinks, or are there other binaries that need to be built?

Binaries

Should be those:
```
-rwxr-xr-x. 2 root root 1.3M Jan  1  1970 strip-ansi-escapes
-rwxr-xr-x. 2 root root  24M Jan  1  1970 wezterm
-rwxr-xr-x. 2 root root  70M Jan  1  1970 wezterm-gui
-rwxr-xr-x. 2 root root  25M Jan  1  1970 wezterm-mux-server
```

jhheider:
i can see you're still working, but this works for me to build them:
```yaml
distributable:
  url: https://github.com/wez/wezterm/releases/download/{{version.tag}}/wezterm-{{version.tag}}-src.tar.gz
  strip-components: 1

versions:
  github: wez/wezterm/tags
  transform: v => v.replace(/^(\d{4})(\d{2})(\d{2})-.*$/, '$1.$2.$3')

dependencies:
  zlib.net: ^1.3
  linux:
    freetype.org: '*'
    freedesktop.org/fontconfig: '*'
    openssl.org: ^1.1

build:
  dependencies:
    rust-lang.org: '>=1.71<1.78'
    rust-lang.org/cargo: '*'
  script:
    - cargo install $ARGS --path strip-ansi-escapes
    - cargo install $ARGS --path sync-color-schemes
    - cargo install $ARGS --path wezterm
    - cargo install $ARGS --path wezterm-gui
    - cargo install $ARGS --path wezterm-mux-server
  env:
    ARGS:
      - --locked
      - --root {{prefix}}

provides:
  - bin/strip-ansi-escapes
  - bin/sync-color-schemes
  - bin/wezterm
  - bin/wezterm-gui
  - bin/wezterm-mux-server

test: test "$(wezterm --version)" = "wezterm {{version.tag}}"
```

jhheider:
```shell
$ cat ~/.pkgx/x.org/x11/v1.8.10/lib/pkgconfig/x11.pc 
prefix=${pcfiledir}/../..
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

xthreadlib=-lpthread

Name: X11
Description: X Library
Version: 1.8.10
Requires: xproto kbproto
Requires.private: xcb >= 1.11.1
Cflags: -I${includedir} 
Libs: -L${libdir} -lX11
Libs.private: -lpthread
```

looks like you need x.org/xcb as a dep as well.

jhheider:
looks like they only provide a cask for macos on homebrew, so it'll be some trial-and-error to get the full list of linux libraries if wezterm doesn't specify them: https://github.com/Homebrew/homebrew-cask/blob/7db552eae14ba123c69e5e9a55f3ca395d8cfbec/Casks/w/wezterm.rb

jhheider:
hm, xcb-util.pc is _not_ in x.org/xcb. looks like it's its own package: https://xcb.freedesktop.org/dist/

edit: this should be ready shortly.

sebst:
> looks like they only provide a cask for macos on homebrew, so it'll be some trial-and-error to get the full list of linux libraries if wezterm doesn't specify them: https://github.com/Homebrew/homebrew-cask/blob/7db552eae14ba123c69e5e9a55f3ca395d8cfbec/Casks/w/wezterm.rb

https://wezfurlong.org/wezterm/install/source.html

PR #7442, opened 2024-09-27
add krampus package

idleberg:
I just discovered pkgx and start seeing the potential for my own use cases. That said, I'd be delighted if you accept this Go project of mine.


jhheider:
looks great! thanks for the contribution, @idleberg . `krampus` looks like it would be very useful in _other_ pantry packages for testing (to clear known ports).

PR #7449, opened 2024-09-27
upgrade(buster,bullseye)

jhheider:
buster-* images no longer published on docker hub by debian

tagging @mxcl for visibility


mxcl:
yikes, we need immutable decentralized containers

jhheider:
reverted (docker hub was the issue), but this tag is no longer updated (since `bullseye` is now `oldstable`). shouldn't affect us, but it was an interesting find.

PR #7473, opened 2024-09-29
fix: pc file for libxfont2

sebst:


sebst:
```
libtool: warning: library '/opt/x.org/libfontenc/v1.1.8/lib/pkgconfig/../../lib/libfontenc.la' was moved.
libtool: warning: library '/opt/x.org/libfontenc/v1.1.8/lib/pkgconfig/../../lib/libfontenc.la' was moved.
ld: warning: duplicate -rpath '/opt' ignored
+ make install
```

jhheider:
those changes are handled automatically by brewkit. i think what we need is to remove the *.la files from `libfontenc` (those aren't relocatable, but also not necessary).

jhheider:
closing this for now, since i think its need was otherwise satisfied (by removing .la files).

PR #7474, opened 2024-09-29
+x11.org/xserver (wip)

sebst:


tannevaled:
Drive-by triage — this is open for 20 months with no recent activity. Bringing it up to ship-ready state needs substantial rework rather than small fixes:

**Build system has moved on.** Modern `xorg-server` (>= 21.1.x) is meson-built, not autotools. arch's [xorg-server PKGBUILD](https://gitlab.archlinux.org/archlinux/packaging/packages/xorg-server/-/blob/main/PKGBUILD) uses `meson setup` + `meson compile`, and the source is now fetched from `gitlab.freedesktop.org/xorg/xserver` rather than the legacy `www.x.org/archive/individual/xserver/` tarballs. The current PR's `./configure $ARGS` flow won't apply to any version published in the last ~3 years.

**Hardcoded multiarch path.** Lines like
```
cp {{deps.x.org/libcvt.prefix}}/lib/x86_64-linux-gnu/pkgconfig/libxcvt.pc …
```
break on darwin and on linux/aarch64 — only Debian-derivative distros put pkg-config files under `lib/x86_64-linux-gnu/`. Pkgx bottles ship `pkgconfig` at `lib/pkgconfig/`. The `cp` is also a workaround for an apparently-misnamed `.pc` file; root-cause would be to fix the libcvt recipe or the project's `find_package` query, not to symlink shim it here.

**Test target mismatch.** `provides: bin/Xorg` but `test: xserver --version` — different binary names. Should be `Xorg -version` (note: single dash, that's how Xorg's flag parser works).

**Missing/wrong deps for current xserver.** arch's `makedepends` runs to ~30 entries (mesa, libxkbfile, libxfont2, libxv, libxcvt, libxmu, libepoxy, xcb-util-*, libxshmfence, libunwind, systemd, meson, …). The current PR's deps list is partial and pre-meson.

**Recommendation:** treat this as a fresh attempt rather than a fix-up. The closest viable path is to mirror arch's recipe shape (meson-based, fetch from gitlab.fd.o, x86_64-only platforms, full dep list) — or borrow from nixpkgs which also drives the meson build. Happy to take this forward as a new PR superseding #7474 if a maintainer can confirm there's interest in shipping an Xorg server bottle; otherwise probably best to close as superseded-by-time.

For reference:
- arch PKGBUILD: <https://gitlab.archlinux.org/archlinux/packaging/packages/xorg-server/-/blob/main/PKGBUILD>
- nixpkgs xorg-server: <https://github.com/NixOS/nixpkgs/blob/master/pkgs/servers/x11/xorg/default.nix>

PR #7504, opened 2024-10-02
fix(watchman)

Dunedubby:
Created a new pull request because my older one was bogged down by a lot of commits to externally fix a problem (CMake prefix) which hopefully is now fixed in the pkgx framework. This pull request has a different scope: getting the WATCHMAN_STATE_DIR variable to be relocatable.

### DIFF
--- a/projects/facebook.com/watchman/package.yml
+++ b/projects/facebook.com/watchman/package.yml
@@ -42,11 +42,11 @@ build:
         - mainfile=$(find . -name "main.cpp")
         - 'echo "Found main.cpp at: $mainfile"'
         - oldline='auto state_dir = computeWatchmanStateDirectory(user);'
-        - 'newline=\'const char* env_state_dir = getenv("WATCHMAN_STATE_DIR"); auto state_dir = env_state_dir ? env_state_dir : computeWatchmanStateDirectory(user);\''
+        - "newline='const char* env_state_dir = getenv(\"WATCHMAN_STATE_DIR\"); auto state_dir = env_state_dir ? env_state_dir : computeWatchmanStateDirectory(user);'"
         - sed -i "s/$oldline/$newline/" "$mainfile"
-        cat $mainfile
+        - cat $mainfile
       working-directory: watchman
-      
+
     # version/path mismatch
     - run: |
         if test -f Cargo.toml; then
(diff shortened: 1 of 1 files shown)


Dunedubby:
Just wanted to point out that this seems somewhat relevant: https://aur.archlinux.org/packages/edencommon (the comments).

jhheider:
i'm not surprised others are having issues. these package don't follow any versioniong, other than weekly releases, whether or not they're stable, and the versions are all tightly-coupled.

Dunedubby:
Yeah it's so messy. I'm thinking it might be necessary to switch fb303 and edencommon to gcc instead of clang as well. Do you think that'll be fine? If so, is it pkgx best practices to create a new branch for each, or is it fine to edit them in this pull request as well?

jhheider:
Fine to do in a single branch, but the rebuilds won't be available since they're done in parallel.

jhheider:
#7539 moves fbthrift to gcc, and fixes the recent build errors. it may help. i'll look at fb303 too.

Dunedubby:
Hmm it's so weird... from the errors it's like the edencommon package wants an old version of folly (<2024.09.16, not even just 2024.09.23), but obviously the edencommon source itself only references the most recent version of folly. My hunch is that there's some weird dependency daisy chain where one of the facebook packages has a dependency on the previous version, which in turn has a package that depends on the previous version...

Dunedubby:
I'm starting to strongly suspect the problem may be wangle... https://github.com/pkgxdev/pantry/actions/runs/11218670141/job/31182979376

jhheider:
the fact that the watchman releases hadn't built a distributable binary on their GitHub since May the last I checked suggests that it releases pro-forma on Monday like the others, but isn't really in a 'release' state.

jhheider:
i think after moving everything to `gcc` on linux, i finally got it. merging #6389

Dunedubby:
Oh that's awesome!!! Yeah I was getting stuck on linking Python, so I'm really impressed that you figured it out! Does that build fix the state directory not in userspace problem?

jhheider:
> Oh that's awesome!!! Yeah I was getting stuck on linking Python, so I'm really impressed that you figured it out! Does that build fix the state directory not in userspace problem?

If I didn't that part should be easy, I hope. Let me know.

jhheider:
ah, it builds. does that fix the issue for you?

Dunedubby:
After you fixed everything (thank you so much) the state directory started pointing to the original, backup state directory location in /tmp. This PR now would just put the state directory in .pkgx instead of /tmp ... but I don't know if it's worth the rather hacky/manual script edit. Maybe it's better just to close without merging

jhheider:
no, i think patching is the right solution. we do it for other packages for relocatability. if this solves your problem, just mark it ready and I'll merge it.

Dunedubby:
Thanks for closing this! Sorry, got pulled into some tight deadlines at work 😅

PR #7600, opened 2024-10-12
+github.com/mycreepy/pakku

mycrEEpy:
pakku is a declarative approach to system package managers (still under development but happy to share)


jhheider:
pretty cool. thanks for supporting `pkgx`!

PR #7864, opened 2024-11-09
+modernc.org/goyacc

tannevaled:


tannevaled:
is there an equivalent to {{prefix}} like {{testbed}} to do not ship the tests files?

jhheider:
> is there an equivalent to {{prefix}} like {{testbed}} to do not ship the tests files?

no... it's $PWD in the test script. if you want to run `go test` in the source directory, you need to do it in the build step (build files aren't available at test time).

you can also use `curl` or `git` to retrieve test files, like `gobject-introspection` and others do.

tannevaled:
> > is there an equivalent to {{prefix}} like {{testbed}} to do not ship the tests files?
> 
> no... it's $PWD in the test script. if you want to run `go test` in the source directory, you need to do it in the build step (build files aren't available at test time).
> 
> you can also use `curl` or `git` to retrieve test files, like `gobject-introspection` and others do.

ok. is it possible to have something like {{distributable.url}} to not repeat something already done?

```bash
...
+ cd /root/work/pkgxdev-pantry-goyacc/testbeds/modernc.org__goyacc-1.0.3
+ git clone '{{distributable.url}}' test
fatal: repository '{{distributable.url}}' does not exist
```

jhheider:
> ok. is it possible to have something like {{distributable.url}} to not repeat something already done?

It could be possible, that would be something like reading in the yaml itself into https://github.com/pkgxdev/libpkgx/blob/main/src/hooks/useMoustaches.ts .

Generally, tests that are built into the build source are tested at build time, rather than test time (so, run `go test` in the build script, e.g. `make check` in `libpaper`). And the `test` script should do its best to ensure that the installed binary is working correctly, ideally.

PR #7904, opened 2024-11-13
fix(skopeo)

### LINKED ISSUE #7903: ❌ build issues: github.com/containers/skopeo=1.17.0
Running log of build failure for github.com/containers/skopeo=1.17.0

jhheider:
closes #7903

### DIFF
--- a/projects/github.com/containers/skopeo/package.yml
+++ b/projects/github.com/containers/skopeo/package.yml
@@ -9,33 +9,24 @@ provides:
   - bin/skopeo
 
 dependencies:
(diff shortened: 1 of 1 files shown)


felipecrs:
@jhheider, I updated the patch to work with all versions of skopeo:

- https://github.com/felipecrs/skopeo-bin/pull/1

Do you think it's ok if I introduce it again?

jhheider:
I've lost context on the broader issue in 9 months but patches are always welcome. :)

PR #7990, opened 2024-11-22
fix(cocoapods)

waliid:


jhheider:
i think this is a worse name. the redundant path doesn't add anything. thanks for the suggestion.

PR #7992, opened 2024-11-22
update(fastlane)

waliid:


jhheider:
this is likely a companion (works with) rather than a dependencies (requires), right? Otherwise, it wouldn't work at all.

PR #8343, opened 2025-01-07
fix(pkgx.dev)

jhheider:


mxcl:
lol my double bad

jhheider:
might be a problem to resolve with brewkit (which is locked to ~1.39)

jhheider:
but using pkgx in the prop might work. testing.

mxcl:
```
/opt/deno.land/v2.1.4/bin/deno: error while loading shared libraries: libffi.so.8: cannot open shared object
```

ah, deno is no longer standalone so presumably this hack is not going to work on linux, might work if I just absorb the env without an export actually…

jhheider:
erm, no, using pkgx in the shim gives the dirty env issue. hm. might be able to just hard-path the libs.

mxcl:
I'll need a proper solution in `dev` proper by the looks of it. dang.

jhheider:
we could see if we can statically link deno. i mean, we should if we can anyway.

jhheider:
rock

PR #8419, opened 2025-01-15
+pkgm

### LINKED ISSUE #8405: +pkgm
Why not? lol

https://github.com/pkgxdev/pkgm

jhheider:
closes #8405

### DIFF
--- a/projects/pkgx.sh/pkgm/package.yml
+++ b/projects/pkgx.sh/pkgm/package.yml
@@ -23,8 +23,8 @@ test:
   - pkgx --version
   - head -n1 {{prefix}}/bin/pkgm
   - pkgm install dua
-  # TODO: this will work eventually
-  # - pkgm ls | grep dua
-  - command -v dua
-  - dua --version
+  - pkgm ls | grep dua
+  - PATH=/usr/local/bin:$PATH command -v dua
+  # ^^ brewkit PATH removes /usr/local/bin
+  - PATH=/usr/local/bin:$PATH dua --version
   - pkgm uninstall dua
--- a/projects/pkgx.sh/pkgm/package.yml
+++ b/projects/pkgx.sh/pkgm/package.yml
@@ -9,11 +9,19 @@ provides:
 
 dependencies:
   pkgx.sh: ^2
+  curl.se/ca-certs: '*'
 
-build: install -Dm755 pkgm {{prefix}}/bin/pkgm
+build:
+  - sed -i 's|--allow-run=pkgx,/usr/bin/sudo|--allow-run|' pkgm
+  - head -n1 pkgm
+  - install -Dm755 pkgm {{prefix}}/bin/pkgm
 
 test:
   - pkgm --version
+  - command -v pkgm
+  - command -v pkgx
+  - pkgx --version
+  - head -n1 {{prefix}}/bin/pkgm
   - pkgm install dua
   # TODO: this will work eventually
   # - pkgm ls | grep dua
--- a/projects/pkgx.sh/pkgm/package.yml
+++ b/projects/pkgx.sh/pkgm/package.yml
@@ -16,7 +16,11 @@ build:
 
 test:
   - pkgm --version | grep {{version}}
-  - pkgm install dua
+  - run: pkgm $INSTALL_CMD dua
+    env:
+      linux: { INSTALL_CMD: 'install' }
+      darwin: { INSTALL_CMD: 'local-install' }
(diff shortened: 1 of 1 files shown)


jhheider:
@mxcl odd. i wonder if this is because both v1 and v2 are present?
```shell
  ▌ pantry script start
+ cd /Users/jacob/pkgx/pantry/testbeds/pkgx.sh__pkgm-0.3.0
+ pkgm --version
pkgm 0.3.0
+ pkgm install dua
┏ ⚠️  Deno requests run access to "pkgx".
┠─ Requested by `Deno.Command().spawn()` API.
┠─ To see a stack trace for this prompt, set the DENO_TRACE_PERMISSIONS environmental variable.
┠─ Learn more at: https://docs.deno.com/go/--allow-run
┠─ Run again with --allow-run to bypass this prompt.
┗ Allow? [y/n/A] (y = yes, allow; n = no, deny; A = allow all run permissions) > 
```
^local. on CI it just fails.

jhheider:
adding:
```yaml
  - sed -i 's|--allow-run=pkgx,/usr/bin/sudo|--allow-run|' pkgm
```
"fixes" that, but then i get:
```shell
  ▌ pantry script start
+ cd /Users/jacob/pkgx/pantry/testbeds/pkgx.sh__pkgm-0.3.0
++ pkgm --version
+ test 'pkgm 0.3.0' = 'pkgm 0.3.0'
+ pkgm install dua
 × usage error: no such arg: --json
                       ^
    at file:///Users/jacob/.pkgx/pkgx.sh/brewkit/v1.14.3/bin/bk-test:106:24
    at eventLoopTick (ext:core/01_core.js:182:7)
```
which means it's finding v1 first. so... maybe this can't be done until we move brewkit to v2?

felipecrs:
Maybe we should `deno compile pkgm.ts` during build time to avoid a hard runtime dependency on pkgx?

jhheider:
possibly, but a) deno compile binaries are _huge_, and b) it requires pkgx to functional in any case.

felipecrs:
I see. Maybe you should remove `pkgx` from the list of dependencies? There's no way this will be executed without pkgx installed in the system anyway, lol.

mxcl:
there's a bug somewhere if it finds v1 when the package depends on v2.

jhheider:
that's... what it seems like, even though path inspection shows v2. and the darwin version gets `can't spawn`. so, maybe it just needs to be on-hold.

felipecrs:
What's the benefit of keeping pkgx in the dependencies list? It would trigger pkgx to redundantly install another pkgx. Also, if users installed pkgx in their machine, they would expect pkgm to reuse that installation. It doesn't make sense to break that expectation, does it?

jhheider:
it strictly requires pkgx^2. and may, in the future, strictly require higher versions. so, same benefits of mapping dependencies.

felipecrs:
Regardless of which version of pkgx pkgm requires, I would expect pkgm to use the version of pkgx which I installed to my system, even if it is not compatible and fails.

Maybe pkgm should be an exception, as it manages stuff installed at system level, not at pkgx dir level.

Well, maybe I'm going too crazy. lol

felipecrs:
Maybe this is like `mvn`, which I would expect to use my system's Java, but instead it downloads its own Java.

felipecrs:
Maybe my concern will be "fixed" by: https://github.com/pkgxdev/pkgx/issues/971#issuecomment-1936978538

I.e. would allow my system installed pkgx to satisfy the (companion) dependency.

mxcl:
`pkgm` expects `pkgx` to be in a sanitized path which for this packaging scenario is not true. I can fix on the `pkgm` end.

felipecrs:
@mxcl what do you think about making `pkgm` provide `install` and `uninstall` commands?

This would "restore" backwards compatibility with v1 by allowing `pkgx install deno` and `pkgx uninstall deno` to work.

mxcl:
`pkgm` is a bit spaghetti code. Can probably fix the above by setting `PKGX` to the `pkgx` path before re-executing for the install. Fix coming up

edit: https://github.com/pkgxdev/pkgm/pull/32

mxcl:
k fails differently now with v0.5

on mac because `sudo` doesn't work in CI. Really pkgm should check if it *needs* sudo or not before trying to run it. Since on CI it doesn’t.

The other issue I dunno about. `command -v dua` fails after `pkgm i dua`

mxcl:
```sh
$ pkgx pkgm shim pkgm
# ^^ 🤣
```

PR #8433, opened 2025-01-16
If we have pkgx^2.1 use --quiet

mxcl:
I wonder how much this slows down execution and if we can do it with bash primitives

### DIFF
--- a/projects/pkgx.sh/dev/package.yml
+++ b/projects/pkgx.sh/dev/package.yml
@@ -14,13 +14,25 @@ build:
   - echo 'export default "{{version}}"' > ./src/app-version.ts
   - mkdir -p {{prefix}}/bin {{prefix}}/share/pkgx/dev
   - cp -r ./app.ts src deno.json deno.lock {{prefix}}/share/pkgx/dev
+
   - run:
       - cp $PROP {{prefix}}/bin/dev
       - chmod +x {{prefix}}/bin/dev
+      - args="$(sed -n '1s/.*pkgx //p' ./app.ts)"
+      - echo exec pkgx \$_PKGX_ARG "$args" '"$_PKGX_DEV_ROOT/share/pkgx/dev/app.ts" "$@"' >> {{prefix}}/bin/dev
     prop: |
       #!/bin/sh
-      d="$(cd "$(dirname $0)"/.. && pwd)"
-      exec "$d/share/pkgx/dev/app.ts" "$@"
+
+      _PKGX_DEV_ROOT="$(cd "$(dirname $0)"/.. && pwd)"
+
+      if echo "$PATH" \\
+        | grep -o 'pkgx\.sh/v[0-9]\+\.[0-9]\+\.[0-9]\+/bin' \\
+        | sed -E 's/.*v([0-9]+\.[0-9]+\.[0-9]+)\/bin/\1/' \\
+        | awk -F. '{if ($1 > 2 || ($1 == 2 && $2 >= 1)) exit 0; else exit 1;}'
+      then
+        _PKGX_ARG=--quiet
+      fi
+
 provides:
   - bin/dev
 
@@ -35,4 +47,3 @@ test:
     - dev
     - deno --version
     - test "$(dev --version)" = 'dev {{version}}'
-


projects/pkgx.sh/dev/package.yml
```diff
@@ -14,13 +14,25 @@ build:
...
       #!/bin/sh
-      d="$(cd "$(dirname $0)"/.. && pwd)"
-      exec "$d/share/pkgx/dev/app.ts" "$@"
+
+      _PKGX_DEV_ROOT="$(cd "$(dirname $0)"/.. && pwd)"
+
+      if echo "$PATH" \\
+        | grep -o 'pkgx\.sh/v[0-9]\+\.[0-9]\+\.[0-9]\+/bin' \\
+        | sed -E 's/.*v([0-9]+\.[0-9]+\.[0-9]+)\/bin/\1/' \\
+        | awk -F. '{if ($1 > 2 || ($1 == 2 && $2 >= 1)) exit 0; else exit 1;}'
```
jhheider, line 31: `test "$string" = "*pkgx.sh/v2*" && test "$string" != "*pkgx.sh/v2.0*"`?

jhheider: or just use variable substitution and a case statement with wild cards.

jhheider: lol, or write it in amber-lang and see what it compiles to ;)

mxcl:
I'm really messing around to have `dev` able to use `--quiet` here (dev itself needs to know too).

Maybe I can just re-release 2.0.0…? How bad is that gunna be really? I dunno if anyone else is packaging it yet. I know Arch AUR has it. Not sure anyone else does.

jhheider:
is 2.0.1 so bad?

mxcl:
I have revlock in that I have to detect if `pkgx` supports `--quiet` or not since `pkgx` itself will not ensure a dep bump to pkgx^2.1 for `dev` happens until a `--sync` and thus all this other logic to detect it. As well as weird scenarios like pkgx 2.0.0 existing for some reason and then the hardcoded `--quiet` in your `~/.zshrc` will break randomly.

Or… I can re-release pkgx 2.0.0 with this flag I should have added already.

mxcl:
Maybe I'm just overconcerned for something that is only going to effect a couple people as long as I get a new release of pkgx out promptly.

jhheider:
break stuff and move fast.

jhheider:
for the record:
```amber
import { match_regex } from "std/text";

let pkgx_path = $ command -v pkgx $ failed {
    echo "pkgx not found"
    exit 1
}

echo "pkgx found at {pkgx_path}"

if match_regex(pkgx_path, "pkgx\.sh/v2") {
  if match_regex(pkgx_path, "pkgx\.sh/v2\.0") {
    echo "pkgx is v2.0"
  } else {
    echo "pkgx is ^2.1"
  }
} else {
    echo "pkgx is not v2"
}
```
gives
```shell
#!/usr/bin/env bash
# Written in [Amber](https://amber-lang.com/)
# version: 0.4.0-alpha
# date: 2025-01-16 14:24:33
replace__0_v0() {
    local source=$1
    local search=$2
    local replace=$3
    __AF_replace0_v0="${source//${search}/${replace}}";
    return 0
}
match_regex__17_v0() {
    local source=$1
    local search=$2
    local extended=$3
            replace__0_v0 "${search}" "/" "\/";
        __AF_replace0_v0__130_18="${__AF_replace0_v0}";
        search="${__AF_replace0_v0__130_18}"
        local output=""
        if [ ${extended} != 0 ]; then
            # GNU sed versions 4.0 through 4.2 support extended regex syntax,
            # but only via the "-r" option; use that if the version information
            # contains "GNU sed".
             re='\bCopyright\b.+\bFree Software Foundation\b'; [[ $(sed --version 2>/dev/null) =~ $re ]] ;
            __AS=$?
            local flag=$(if [ $(echo $__AS '==' 0 | bc -l | sed '/\./ s/\.\{0,1\}0\{1,\}$//') != 0 ]; then echo "-r"; else echo "-E"; fi)
            __AMBER_VAL_0=$( echo "${source}" | sed "${flag}" -ne "/${search}/p" );
            __AS=$?;
            output="${__AMBER_VAL_0}"
else
            __AMBER_VAL_1=$( echo "${source}" | sed -ne "/${search}/p" );
            __AS=$?;
            output="${__AMBER_VAL_1}"
fi
        if [ $([ "_${output}" == "_" ]; echo $?) != 0 ]; then
            __AF_match_regex17_v0=1;
            return 0
fi
    __AF_match_regex17_v0=0;
    return 0
}
__AMBER_VAL_2=$( command -v pkgx );
__AS=$?;
if [ $__AS != 0 ]; then
    echo "pkgx not found"
    exit 1
fi;
__0_pkgx_path="${__AMBER_VAL_2}"
echo "pkgx found at ${__0_pkgx_path}"
match_regex__17_v0 "${__0_pkgx_path}" "pkgx\.sh/v2" 0;
__AF_match_regex17_v0__10_4="$__AF_match_regex17_v0";
if [ "$__AF_match_regex17_v0__10_4" != 0 ]; then
    match_regex__17_v0 "${__0_pkgx_path}" "pkgx\.sh/v2\.0" 0;
    __AF_match_regex17_v0__11_6="$__AF_match_regex17_v0";
    if [ "$__AF_match_regex17_v0__11_6" != 0 ]; then
        echo "pkgx is v2.0"
else
        echo "pkgx is ^2.1"
fi
else
    echo "pkgx is not v2"
fi
```

PR #8486, opened 2025-01-23
+projen.io

mycrEEpy:
Adding [projen.io](https://github.com/projen/projen)

~~Starting with only JS here since I think it's better to distinguish between the various languages in multiple packages.~~

~~But this would also mean we would have multiple packages which provide `bin/projen` in the end, seems wrong?~~

Not sure about the copy of the peer dependency & having git & yarn as runtime dependencies.
Is that the best way to do it?


projects/projen.io/js/package.yml
```diff
@@ -0,0 +1,38 @@
+distributable:
+  url: https://github.com/projen/projen/archive/refs/tags/v{{version.raw}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: projen/projen
+  strip: /^v/
+
+dependencies:
+  nodejs.org: ^22
```
jhheider: does it strictly require node^22 to run? or can this be relaxed?

projects/projen.io/js/package.yml
```diff
@@ -0,0 +1,38 @@
...
+  classic.yarnpkg.com: ^1
```
jhheider: is this a runtime or a build-time dep?

projects/projen.io/js/package.yml
```diff
@@ -0,0 +1,38 @@
...
+  git-scm.org: "*"
```
jhheider: is this required to run? or just to test? or is it a companion?

projects/projen.io/js/package.yml
```diff
@@ -0,0 +1,38 @@
...
+    - yarn version --new-version {{version.raw}} --no-git-tag-version
+    - yarn install --check-files --frozen-lockfile
+    - yarn package:js
+    - tar xfz dist/js/*.tgz -C {{prefix}}
+    - cp -R node_modules/constructs "{{prefix}}"/package/node_modules # peer dependency
+    - run: ln -s ../package/bin/projen projen
+      working-directory: ${{prefix}}/bin
+
+provides:
+  - bin/projen
```
jhheider: should the package just be `projen.io`? rather than `projen.io/js`?

projects/projen.io/js/package.yml
mycrEEpy: I can relax it I think.

projects/projen.io/js/package.yml
mycrEEpy: Projen is used to bootstrap & maintain projects, it executes yarn during runtime for JS/TS projects.

projects/projen.io/js/package.yml
mycrEEpy: Same as with yarn, projen executes git during runtime when creating or mainting projects. Can you clarify what's the difference between companion and dependency?

projects/projen.io/js/package.yml
mycrEEpy: This is the tricky question i think. Projen can manage various projects next to JS/TS, e.g. Java or Go. For these projects it needs the respective language (and buildtools like Maven) as runtime dependency. I feel like this would clutter the runtime dependencies if one only needs to use Projen on e.g. JS/TS projects.

projects/projen.io/js/package.yml
jhheider: Generally, a program requires a dependency to execute, most often libraries. A companion is usually another program, which may be optional, that provides functionality.

Companion functionality is newer than dependency functionality, and there are, guaranteed, many violations of this split in the repo.

projects/projen.io/js/package.yml
jhheider: Generally, a program requires a dependency to execute, most often libraries. A companion is usually another program, which may be optional, that provides functionality.

Companion functionality is newer than dependency functionality, and there are, guaranteed, many violations of this split in the repo.

projects/projen.io/js/package.yml
jhheider: Hm. But it's the same binary? I'd do it in one of two ways, then:

- all one package, all feasible companions
- all one package, no companions (user includes as needed)

mycrEEpy: Thanks for the valuable input. I'll rename the package to projen.io and change all dependencies but nodejs.org itself to companions.

projects/projen.io/js/package.yml
mycrEEpy: Encountered a warning when relaxing the constraint.

```
This software is currently running on node v23.6.1.
As of the current release of this software, supported node releases are:
  - ^22.0.0 (Planned end-of-life: 2027-04-30)
  - ^20.0.0 (Planned end-of-life: 2026-04-30)
  - ^18.0.0 (Planned end-of-life: 2025-04-30)
```

jhheider: that should be `nodejs.org: ^22 || ^20 || ^18`

PR #8592, opened 2025-02-04
fix(tcl)

jhheider:
config scripts have hard-code paths. breaking new python builds.


jhheider:
skipping fix-machos to get 9.x to build was a mistake. i think we need to accept that 9 isn't ready for macOS yet. we can't even update the libraries by hand with install_name_tool.

PR #8609, opened 2025-02-06
+liburcu.org

tannevaled:


jhheider:
this appears to be the most idiomatic fix:
```yaml
distributable:
  url: https://lttng.org/files/urcu/userspace-rcu-{{version}}.tar.bz2
  strip-components: 1

display-name: liburcu

versions:
  github: urcu/userspace-rcu/tags

build:
  - ./configure $ARGS --prefix="{{prefix}}"
  - make --jobs {{ hw.concurrency }} install
  - run: find . -name '*.la' -exec rm {} \;
    working-directory: '{{prefix}}/lib'

test:
  - run: cc $FIXTURE -lurcu
    fixture:
      extname: c
      content: |
        // SPDX-FileCopyrightText: 2013 Mathieu Desnoyers <mathieu.desnoyers@efficios.com>
        //
        // SPDX-License-Identifier: MIT

        /*
        * This example shows how to add into a linked-list safely against
        * concurrent RCU traversals.
        */

        #include <stdio.h>

        #include <urcu/urcu-memb.h>	/* Userspace RCU flavor */
        #include <urcu/rculist.h>	/* RCU list */
        #include <urcu/compiler.h>	/* For CAA_ARRAY_SIZE */

        /*
        * Nodes populated into the list.
        */
        struct mynode {
          int value;                 /* Node content */
          struct cds_list_head node; /* Linked-list chaining */
        };

        int main(void)
        {
          int values[] = { -5, 42, 36, 24, };
          CDS_LIST_HEAD(mylist);		/* Defines an empty list head */
          unsigned int i;
          int ret = 0;
          struct mynode *node;

          /*
            * Adding nodes to the linked-list. Safe against concurrent
            * RCU traversals, require mutual exclusion with list updates.
            */
          for (i = 0; i < CAA_ARRAY_SIZE(values); i++) {
              node = malloc(sizeof(*node));
              if (!node) {
                ret = -1;
                goto end;
              }
              node->value = values[i];
              cds_list_add_rcu(&node->node, &mylist);
          }

          /*
            * Just show the list content. This is _not_ an RCU-safe
            * iteration on the list.
            */
          printf("mylist content:");
          cds_list_for_each_entry(node, &mylist, node) {
              printf(" %d", node->value);
          }
          printf("\n");
        end:
          return ret;
        }

  - 'test "$(./a.out)" = "mylist content: 24 36 42 -5"'
```

thanks for finding a really good test. that's often the most strenuous part.

tannevaled:
this test comes from the doc folder of the liburcu repository. maybe better to get it from the sources.

jhheider:
if there's substantial reason to think it'll change as the code evolves, you can probably curl it from the repo at the specific tag. but otherwise, I think this is fine.

projects/liburcu.org/package.yml
```diff
@@ -0,0 +1,84 @@
...
+display-name: liburcu
+
+versions:
+  github: urcu/userspace-rcu/tags
+build:
+  dependencies:
+    gnu.org/gcc: '*'
+    gnu.org/autoconf: '*'
+    gnu.org/make: '*'
+
```
jhheider: in my testing you can omit all of this. including llvm.org explicitly messes up `configure`, but leaving all of these out works fine. _most_ sources come already properly ready to `configure`.

tannevaled: i was thinking it was necessary/better to explicit all the dependencies.
fine to remove.

PR #8612, opened 2025-02-07
fix(open-mpi)

jhheider:
- don't build 3rd party libraries we have
- force Xcode clang on darwin

hopefully this will speed up builds (since both the x86-64s are timing out at 6 hours).


jhheider:
Brings the build, often exceeding 6 hours, under 4.

PR #8636, opened 2025-02-09
+kernel.org/libcap

tannevaled:


tannevaled:
i do not understand how to get rid of this error 'ldd: /usr/bin/ld: No such file or directory'

jhheider:
maybe set `build.env.linux.LD: clang`?

tannevaled:
ok, will try. it build on my own box without this.

jhheider:
probably you _have_ something at `/usr/bin/ld`. the builder environments are intentionally minimal. it might be that something is hard-coded somewhere (grepping the sources could help).

tannevaled:
good point
$ grep -R '/usr/bin/ld' srcs/kernel.org__libcap-1.2.73/*
srcs/kernel.org__libcap-1.2.73/Make.Rules:lib=$(shell ldd /usr/bin/ld|grep -E "ld-linux|ld.so"|cut -d/ -f2)

jhheider:
that'll be it. simple `sed` to fix that path might be enough.

tannevaled:
not so simple
.pkgx/pkgx.sh/brewkit/v1.14.3/share/toolchain/bin/ld is not a dynamic executable so the logic must be changed in the Make.Rules file

jhheider:
maybe check a different binary? this looks like it works:
```makefile
lib=$(shell ldd /bin/*|grep -E "ld-linux|ld.so"|cut -d/ -f2|head -n1)
```

tannevaled:
thank you for the last mile

jhheider:
no problem. i look at these _a lot_ O_O

PR #8766, opened 2025-02-19
+code-cli

### LINKED ISSUE #8745: +code.visualstudio.com/cli
VS Code CLI can be installed standalone on headless servers to enable functions like [remote tunnels](https://code.visualstudio.com/).

It's offered to be downloaded in their website as well:

Image

I honestly don't know if this will be possible, but I noticed this `compile-cli` script in the vscode project:

- https://github.com/microsoft/vscode/blob/7923f036c990433adf840e22e9ba0881ab93790a/package.json#L50

Consider I'm just opening this issue for exploration purposes.

jhheider:
closes #8745

### DIFF
--- a/projects/microsoft.com/code-cli/package.yml
+++ b/projects/microsoft.com/code-cli/package.yml
@@ -1,0 +1,31 @@
+distributable:
+  url: https://github.com/microsoft/vscode/archive/refs/tags/{{version.tag}}.tar.gz
+  strip-components: 1
+
+displaty-name: vscode cli
+
+versions:
+  github: microsoft/vscode
+
+dependencies:
+  openssl.org: 1.1
+  zlib.net: 1
+
+build:
+  dependencies:
+    rust-lang.org: ^1.81
+    rust-lang.org/cargo: '*'
+  working-directory: cli
+  script: cargo install --locked --path . --root {{prefix}}
+  env:
+    OPENSSL_NO_VENDOR: 1
+    OPENSSL_DIR: '{{deps.openssl.org.prefix}}'
+    VSCODE_CLI_NAME_LONG: Code OSS
+    VSCODE_CLI_VERSION: '{{version}}'
+
+provides:
+  - bin/code
+
+test:
+  - test "$(code --version)" == "code-oss {{version}} (commit unknown)"
+  - code tunnel prune | grep 'Successfully removed all unused servers'


projects/microsoft.com/code-cli/package.yml
```diff
@@ -0,0 +1,31 @@
+distributable:
+  url: https://github.com/microsoft/vscode/archive/refs/tags/{{version.tag}}.tar.gz
+  strip-components: 1
+
+displaty-name: vscode cli
```
felipecrs, line 5: Oops, typo?

projects/microsoft.com/code-cli/package.yml
```diff
@@ -0,0 +1,31 @@
...
+build:
+  dependencies:
+    rust-lang.org: ^1.81
+    rust-lang.org/cargo: '*'
+  working-directory: cli
+  script: cargo install --locked --path . --root {{prefix}}
+  env:
+    OPENSSL_NO_VENDOR: 1
+    OPENSSL_DIR: '{{deps.openssl.org.prefix}}'
+    VSCODE_CLI_NAME_LONG: Code OSS
```
felipecrs, line 23: Just wondering... Why did you change from the default _Code - OSS_?

https://github.com/microsoft/vscode/blob/f14abad992a026eef3c83491c59b3e2628b99f66/cli/src/constants.rs#L58

jhheider: Generally observe homebrew's methodology. You're correct that it shouldn't be needed.

projects/microsoft.com/code-cli/package.yml
jhheider: Yep.

mxcl:
neat. I think. Does this function like the `code` command you install via VSCode.app or is it different?

jhheider:
that is precisely what it is. afaik.

felipecrs:
Too bad it didn't work for my intended use case, but it's not pkgx's fault.

- https://github.com/microsoft/vscode/issues/241257

felipecrs:
> neat. I think. Does this function like the `code` command you install via VSCode.app or is it different?

It's a bit more complicated because the code-cli when compiled from source will look for an installation of Code - OSS, not an installation of VS Code (which is closed source).

But I think it's configurable.

```console
❯ pkgx code --new-window
No installation of Code oss was found.

If you already installed Code and we didn't detect it, run `code version use oss --install-dir /path/to/installation`
```

However, this other feature here is cool:

```console
❯ pkgx code serve-web

[2025-02-20 15:07:45] warn error getting latest version: Updates are are not available: no configured quality
Web UI available at http://127.0.0.1:8000?tkn=18022fed-6cd1-4073-9d66-bf4e613befdd
[2025-02-20 15:07:49] error error getting latest version: Updates are are not available: no configured quality
[2025-02-20 15:07:49] error error getting latest version: Updates are are not available: no configured quality
```

It was supposed to serve you vscode like `vscode.dev` in your browser, but it also doesn't work for a similar reason of the one linked above lol.

PR #8963, opened 2025-03-19
+qsv

sanchitram1:


sanchitram1:
```
--- stderr
  error: The `auto-initialize` feature is enabled, but your python installation only supports embedding the Python interpreter statically. If you are attempting to run tests, or a binary which is okay to link dynamically, install a Python distribution which ships with the Python shared library.

  Embedding the Python interpreter statically does not yet have first-class support in PyO3. If you are sure you intend to do this, disable the `auto-initialize` feature.

  For more information, see https://pyo3.rs/v0.24.0/building-and-distribution.html#embedding-python-in-rust
```

@jhheider from the [linked pyo3 webpage](https://pyo3.rs/v0.24.0/building-and-distribution.html#embedding-python-in-rust), it looks like you can either statically or dynamically attach python. from the error message, it looks like we are attaching the Python package statically and cannot add it dynamically.  The pyo3 webpage cites many more issues with statically attaching it, the way we are now.

I've removed the pyo3 feature for now, maybe a TODO for later?

jhheider:
looks like homebrew went with a very small feature set: `system "cargo", "install", *std_cargo_args, "--features", "apply,luau,feature_capable"`. i often like to start where they are, since they've got a lot of experience/usage.

PR #9087, opened 2025-04-06
Install pre-reqs

mxcl:
My intention is to remove the pre-reqs installation from the installer (and thus also the `pkgxdev/setup` github action). Because:

* It slows down pkgxdev/setup a lot (minutes vs seconds)
* These deps aren't needed (99%)
  * Mostly they are only needed compiling C++ bar the `netbase` dep which is trickier.
* Once we merge manifests they won't be needed at all

So we install them here instead.


jhheider:
Seems fine. Obviously once the `sudo` detection works.

jhheider (approved):

PR #9152, opened 2025-04-13
fix(ruby)

### LINKED ISSUE #8718: ❌ build issues: ruby-lang.org=3.4.2
Running log of build failure for ruby-lang.org=3.4.2

### LINKED ISSUE #9157: ❌ build issues: ruby-lang.org=3.4.3
Running log of build failure for ruby-lang.org=3.4.3

### LINKED ISSUE #9365: Please build Ruby 3.4.3
It appears there's only Ruby 3.3.x versions available. Would be great to start with version 3.4.3

thank you!

scottjg:
closes #8718

### DIFF
--- a/projects/ruby-lang.org/package.yml
+++ b/projects/ruby-lang.org/package.yml
@@ -26,7 +26,7 @@ build:
   script:
     - patch -p1 -F5 < props/mkconfig.rb.diff
 
-    - ./configure $ARGS
+    - CC=cc CXX=c++ ./configure $ARGS
 
     # clang17 doesn't like [[maybe_unused]] on types
     # fix cribbed from https://github.com/ruby/ruby/pull/4603/commits/fbdff085ef3c8a56e0a33404e8795879e3167549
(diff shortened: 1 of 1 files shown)


projects/ruby-lang.org/package.yml
```diff
@@ -79,10 +79,21 @@ build:
...
-        mv *-{{hw.platform}}*/* .
-        rmdir *-{{hw.platform}}*
+        if test -d *-{{hw.platform}}* ; then
+          mv *-{{hw.platform}}*/* .
+          rmdir *-{{hw.platform}}*
+        fi
       working-directory: ${{prefix}}/lib
       if: '>=2.6'
+    - run: |
+        if test *-{{hw.platform}}*/bin/ruby ; then
```
jhheider, line 89: this shouldn't ever run, because of line 83, i think? should line 87 be changed to `'>=2.6<3.4'`?

jhheider:
assuming this needs https://github.com/pkgxdev/brewkit/pull/330, i'm optimistic that'll pass review. i appreciate the help!

not to be greedy, but if you wanted to look at the recent rubygems failures (i'm no rubynista), help is _always_ welcome and appreciated.

projects/ruby-lang.org/package.yml
scottjg: they're actually a different `working-directory` in each clause. one is moving stuff in `lib/` and one is moving stuff at the prefix root

scottjg: the main reason i added the `test` clauses was because the bug here seems like ruby might have some mishandling of the `--disable-multiarch` option in configure.ac, which necessitates the need for those workarounds. if it got fixed upstream it would break the formula without those new conditionals.

i don't totally have the history here why that option is being used (maybe there's a good reason), but if it was possible to just not use those options it might make the formula a little less brittle. something for another PR, maybe.

jhheider: Ah, good catch. Yeah, some of these formulas get complicated.

scottjg:
> not to be greedy, but if you wanted to look at the recent rubygems failures (i'm no rubynista), help is _always_ welcome and appreciated.

took an initial look, but might need some guidance: https://github.com/pkgxdev/pantry/issues/9150#issuecomment-2800339467

projects/ruby-lang.org/package.yml
```diff
@@ -26,7 +26,7 @@ build:
   script:
     - patch -p1 -F5 < props/mkconfig.rb.diff
 
-    - ./configure $ARGS
+    - CC=cc CXX=cxx ./configure $ARGS
```
scottjg: i think this should fix the rubygems package native compilation

scottjg:
now that the brewkit fix is merged, does it need to be released for ci to pick up the fix?

jhheider:
yup, it'll require a release.

scottjg:
i guess brewkit hasn't had a release in some time. any chance i could trouble you to do another one with the fix?

jhheider:
ydm, @scottjg ! thanks for this!

PR #9208, opened 2025-04-21
fix hexedit test

sanchitram1:


jhheider:
I don't think this test helps. If hexedit -h succeeded, the test would fail. And it doesn't understand why it failed. So a broken binary would pass this test, and an update that generated output without curses would fail it.

sanchitram1:
fair enough. the real question is how do you test interactive terminal tools like this one?

jhheider:
With rust programs, I usually add a --version flag using sed that doesn't launch the GUI.

sanchitram1:
An example? We (the package manager) add a patch and allow it to pass?

jhheider:
See https://github.com/pkgxdev/pantry/blob/main/projects%2Famp.rs%2Fpackage.yml

PR #9250, opened 2025-04-27
+github.com/withered-magic/starpls

amartani:
[starpls](https://github.com/withered-magic/starpls) is a language server for the starlark language.

Note: I am not the author / maintainer of starpls; I'm just a user of it.


amartani:
Thank you!

PR #9251, opened 2025-04-27
fix(npm): update npmrc

branchv:
Since https://github.com/npm/cli/pull/8071, `npm` warns this was not recognized:

```
$ pkgx npm --version
npm warn Unknown builtin config "update_notifier". This will stop working in the next major version of npm.
11.2.0
```


jhheider:
good catch; i targeted the fix (though if we find out how long ago that changed, we should change it further back) and put in a test to catch unknown config warnings.

PR #9369, opened 2025-05-09
+github.com/Diniboy1123/usque

Derad6709:
Usque is an open-source reimplementation of the Cloudflare WARP client's MASQUE protocol.


jhheider:
thanks, @Derad6709 !

PR #9387, opened 2025-05-12
+ty

jhheider:


sanchitram1:
approve, thumbs up -- we love astral.sh/ty

jhheider:
`astral-sh` is _killing it_ in python tooling.

PR #9436, opened 2025-05-15
+github.com/git-ecosystem/git-credential-manager

ralfschulze:
Add git-credential-manager package

### DIFF
--- a/projects/github.com/git-ecosystem/git-credential-manager/package.yml
+++ b/projects/github.com/git-ecosystem/git-credential-manager/package.yml
@@ -13,6 +13,9 @@ dependencies:
   unicode.org: ^71
   zlib.net: ^1.3
 
+companions:
+  git-scm.org: '*'
+
 build:
   dependencies:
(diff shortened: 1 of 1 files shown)


jhheider:
looks like homebrew is just vendoring the binaries: https://github.com/Homebrew/homebrew-cask/blob/dcc16c139ce1a44d78ac43afec926484423cd65c/Casks/g/git-credential-manager.rb

i assume there's a way to build from local source using `dotnet` (preferred), but this seems very close.

ralfschulze:
Yes, building from source is nicer. Was not too difficult to implement.

ralfschulze:
Cannot get it to work for Arm and Darwin and I also have not test systems available for that, so I disabled these platforms for now.

A "temporary" fix would be to install from binary for these platforms and only compile from source for linux/x86-64 until someone with appropriate hardware can fix it.

jhheider:
dotnet doesn't seem to install things, other than on linux/x86-64. fixed it, at least. thanks for the contribution!

PR #9438, opened 2025-05-15
rye is an astral thingy

sanchitram1:


jhheider:
i'm torn on this, because a) changing names is risky, and b) it's a true homepage. we could have it appear twice, like `rtx` and `mise` which was kept after the rename.

sanchitram1:
(b) is correct, so I think you are correct. Also, if people use `dev` (like me), this would be a breaking change?

jhheider:
exactly.

jhheider:
closing pending a compelling interest to change.

PR #9517, opened 2025-05-26
Добавлен манифест для bybit_to_tradingview

VanoBond:


jhheider:
I don't know what this is, but it appears to be spam. Since there is no description I can read in English, I will close it and give you the benefit of the doubt.

If you would like to reopen this, please provide an English justification, and a working public repo. Any other actions will result in a ban. Thank you.

PR #9519, opened 2025-05-26
Add package manifest for bybit_to_tradingview Python script

VanoBond:
This pull request adds a new package `bybit_to_tradingview`, which is a simple Python script that sends market data from Bybit to TradingView.

📦 Source repository: https://github.com/vanobond/dtmain  
🔖 Version: 1.0.0  
🔑 License: MIT  
🧪 Tested on: Python 3.11+

Let me know if anything needs to be adjusted — happy to improve the manifest!

### DIFF
--- a/projects/github.com/vanobond/dtmain/package.yml
+++ b/projects/github.com/vanobond/dtmain/package.yml
@@ -32,6 +32,23 @@ fetch:
 
 build:
   script: |
+    install -Dm755 bybit_to_tradingview.py $PREFIX/bin/bybit_to_tradingviewprovides:
+  - bybit_to_tradingview
+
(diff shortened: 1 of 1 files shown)


jhheider:
This is a brand new single script with no stars. Nothing is capable of depending on it, and it seems like a simple one-off, rather than a large open-source initiative. I recommend you instead host a mash script (https://github.com/pkgxdev/mash) for it, rather than attempt to distribute it this way. It will be more discoverable, and you'll have greater room to innovate your script as time passes.

I *do* recommend English translation for the widest audience, but don't require it, and mash as a distributed network has no such enforcement.

Thanks for your contribution.

PR #9584, opened 2025-05-30
fix: Add fuse dependency for borgfs

ralfschulze:


ralfschulze:
Have to package libfuse (https://github.com/libfuse/libfuse) and maybe macFUSE (https://macfuse.github.io/) first.

jhheider:
looks like it needs the v2 library; that'll take more work, since it's a different build script.

ralfschulze:
- The Linux version compiles now, but the mount test fails. Need to get my local build environment working for further testing
- Darwin build still has issues. Borgbackup build instructions say the MacOS version needs libfuse2 (llfuse), but macfuse now seems to require v3 in the newest version.

PR #9610, opened 2025-06-03
update sftpgo.com

tannevaled:


jhheider:
good addition. thanks!

PR #9636, opened 2025-06-05
wip: make the app work as a service on linux and darwin

tannevaled:


jhheider:
i'm going to merge this as probably an improvement, even if incomplete.

PR #9754, opened 2025-06-22
+github.com/Virviil/oci2git

amartani:


jhheider:
thanks for this, @amartani !

PR #9876, opened 2025-07-04
Remove >1GB of unused files from helix package

godalming123:


jhheider:
thanks, @godalming123 !

PR #9893, opened 2025-07-08
+apache.org/serf

tannevaled:
work in progress.

dependency for this https://serverfault.com/questions/522646/svn-e170000-unrecognized-url-scheme-for-httpxxxx


tannevaled:
i do not understand why, on the linux runner, it seems gcc is not found, despite beeing a dependency

jhheider:
i've got this almost working locally; i'll push in a minute.

jhheider:
i'll merge this so you can progress, but if you want to compile a deeper test, i'll happily accept it.

PR #9918, opened 2025-07-11
werf: remove linux dependencies

felipecrs:
I don't know why these deps are needed, but they make werf installation a lot heavier than otherwise.

I also refreshed the go version and added the cni go build tag as found [here](https://github.com/werf/werf/blob/081b6e5d884f55504f93e41b87eb3160a526ed41/Taskfile.dist.yaml#L25).

### DIFF
--- a/projects/werf.io/package.yml
+++ b/projects/werf.io/package.yml
@@ -35,7 +35,8 @@ build:
         - netgo
         - no_devmapper
         - static_build
-        - cni
+        # test without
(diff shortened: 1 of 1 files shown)


jhheider:
wfm.

felipecrs:
Thanks for merging, @jhheider. This is working like a charm, at least so far. :D

jhheider:
Love to hear it!

PR #10024, opened 2025-07-25
Disfavor `kaspad` for `rusty-kaspa`

davdroman:
As you can see in https://github.com/kaspanet/kaspad, this version of `kaspad` written in Go has been discontinued for over a year in favor of the one written in Rust https://github.com/kaspanet/rusty-kaspa.

Currently, 99.87% of nodes run the Rust version (source: https://www.kaspalytics.com/).

The problem is the pantry holds both, which makes invocation annoying by having to disambiguate with a + prefix every time.

Can we please remove the old version from the pantry to make life easier?


jhheider:
I hate removing software; a better solution is comment out the `provides:` in `kaspad`, so the provided binary is unambiguous. Go, KAS!

davdroman:
@jhheider no problem, I don't mind that way either. Done 👍 

(not sure if `test:` needs to be commented out too)

jhheider:
Lgtm!

PR #10186, opened 2025-08-09
fix devcontainer config

amartani:
Fix devcontainer config:

- Specify an image; current versions of devcontainer (eg. the VSCode extension) requires an explicit base image. Picked a base debian image, which is a lightweight one with no language-specific additions, since everything else can come from pkgx.
- Fix postAttachCommand to work with newer versions of `dev` (as `pkgx --integrated` doesn't work anymore) and to work on non-bash shells (as debian image uses dash as its `sh` shell).


jhheider:
looks good!

PR #10337, opened 2025-08-28
+elixir-lang.org/otp-27

gaggle:
This adds Elixir specifically compiled and linked against Erlang OTP 27.

Why?

Elixir has tight build-time coupling with Erlang OTP versions. `pkgx`'s main `elixir-lang.org` package remains as `erlang.org: '*'` for latest everything, but production systems often need specific OTP versions for compatibility and reproducibility. For example, [Elixir 1.18.4](https://github.com/elixir-lang/elixir/releases/tag/v1.18.4) supports OTP 25-27.

- This new `package.yml` is identical to main package except the `erlang.org: ^27` dependency
- No `provides` key to avoid conflicting with main package
- Test verifies binary is built against expected OTP version
- Usage: `pkgx +elixir-lang.org/otp-27 elixir --version`

### DIFF
--- a/projects/elixir-lang.org/otp-27/package.yml
+++ b/projects/elixir-lang.org/otp-27/package.yml
@@ -1,0 +1,35 @@
+distributable:
+  url: https://github.com/elixir-lang/elixir/archive/refs/tags/v{{version}}.tar.gz
+  strip-components: 1
+
+versions:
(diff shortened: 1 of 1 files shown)


jhheider (approved):
Love it. Easy to extend if others need other pairs.

PR #10830, opened 2025-10-23
+q

### LINKED ISSUE #10818: +aws.amazon.com/q
https://github.com/aws/amazon-q-developer-cli
https://docs.aws.amazon.com/amazonq
https://formulae.brew.sh/cask/amazon-q

program: `q`

jhheider:
closes #10818

### DIFF
--- a/projects/aws.amazon.com/q/package.yml
+++ b/projects/aws.amazon.com/q/package.yml
@@ -1,0 +1,21 @@
+distributable:
+  url: https://github.com/aws/amazon-q-developer-cli/archive/refs/tags/{{ version.tag }}.tar.gz
+  strip-components: 1
+
+provides:
+  - bin/q
+
+versions:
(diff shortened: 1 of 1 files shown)


projects/aws.amazon.com/q/package.yml
```diff
@@ -0,0 +1,22 @@
+distributable:
+  url: https://github.com/aws/amazon-q-developer-cli/archive/refs/tags/{{ version.tag }}.tar.gz
+  strip-components: 1
+
+provides:
+  - bin/chat_cli
```
felipecrs: I don't think you should provide `chat_cli`, that's more like an internal build-time name.

They never refer to it in documentation.

jhheider: i could see that. might be too generic.

PR #10855, opened 2025-10-26
+tuxpaint

EarldridgeJazzedPineda:
New packages:
* ferzkopp.net/SDL2_gfx
* github.com/markuskimius/SDL2_Pango
* libsdl.org/SDL_mixer
* libsdl.org/SDL_ttf
* tuxpaint.org

My linker insists on using GCC libraries, and it won't use the system-provided ones regardless of `LDFLAGS`. I added a dependency on `gnu.org/gcc` to all of the packages in this pull request, but it's not really feasible in the long term. I have to be careful not to remove the `gnu.org/gcc` dependency or else it will stop working permanently. I am using Solus 4.7 "Endurance" if that helps.

### DIFF
--- a/projects/ferzkopp.net/SDL2_gfx/package.yml
+++ b/projects/ferzkopp.net/SDL2_gfx/package.yml
@@ -1,0 +1,49 @@
+distributable:
+  url: http://www.ferzkopp.net/Software/SDL2_gfx/SDL2_gfx-{{version}}.tar.gz
+  strip-components: 1
+
+display-name: SDL2_gfx
+
+versions:
+  url: https://www.ferzkopp.net/wordpress/2016/01/02/sdl_gfx-sdl2_gfx/
+  match: /SDL2_gfx-\d+\.\d+\.\d+.tar.\gz/
+  strip:
+    - /^SDL2_gfx-/
+    - /\.tar\.gz$/
+
+dependencies:
+  libsdl.org: 2
+
+build:
+  dependencies:
+    freedesktop.org/pkg-config: '*'
+    gnu.org/autoconf: '*'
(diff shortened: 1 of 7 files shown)


jhheider:
> My linker insists on using GCC libraries, and it won't use the system-provided ones regardless of `LDFLAGS`. I added a dependency on `gnu.org/gcc` to all of the packages in this pull request, but it's not really feasible in the long term. I have to be careful not to remove the `gnu.org/gcc` dependency or else it will stop working permanently. I am using Solus 4.7 "Endurance" if that helps.

that sounds like something to solve in `brewkit`. it should be using only pkgx-provided tools, as much as possible.

EarldridgeJazzedPineda:
> My linker insists on using GCC libraries, and it won't use the system-provided ones regardless of `LDFLAGS`.

I figured it out. I had to create a symbolic link at /usr/lib/gcc/x86_64-unknown-linux-gnu that points to /usr/lib/gcc/x86_64-solus-linux.

EarldridgeJazzedPineda:
`pkgx tuxpaint` doesn't work for some reason after building, but here's a screenshot (I'm using Windows Subsystem for Linux):

image

jhheider:
yeah, it doesn't work because the other packages don't exist yet.

jhheider:
that patch is serious work. I'm going to commit this, and _assume_ the test will pass. we can fix the test after, once everything else is built, if it doesn't.

jhheider:
looks like a jpeg library version on linux, and a compiler issue (probably needs to use llvm.org instead of falling back to Xcode) on darwin:

https://github.com/pkgxdev/pantry/actions/runs/19711009201

PR #10871, opened 2025-10-28
update(emacs)

### LINKED ISSUE #10868: version request: gnu.org/emacs
Include native compilation and tree-sitter support?

jhheider:
closes #10868


bdf369:
Possible to use --with-imagemagick as well? Some terminals like iterm2 can display inline images.

jhheider:
looks like that requires X, which is a ball of wax.

PR #10895, opened 2025-10-31
+github.com/pete911/ipcalc

tannevaled:


jhheider:
quick and simple. love it.

PR #11082, opened 2025-11-19
Add needs package

nmb4:
Multi-threaded tool to check if binary(ies) are installed and get their versions.


jhheider:
cool tool. does `needs` need rust as a runtime-companion? if that's just for the tests, it should be under `tests.dependencies`.

nmb4:
it actually doesn't, thx for noting.

PR #11207, opened 2025-12-02
Add pkgx manifest for just-mcp

elasticdotventures:
## Summary
- add the pkgx pantry manifest so pkgx can install the just-mcp binary per host/arch
- document the release artifacts the manifest expects so CI upload remains compatible

## Testing
- not run (packaging change only)


jhheider:
thanks for this, @elasticdotventures ! i _love_ just, and this is a great idea.

elasticdotventures:
> thanks for this, @elasticdotventures ! i _love_ just, and this is a great idea.

Thanks @jhheider -- thats great, I really love pkgx.  

As a rust dev it's certainly the easiest way to publish mcp servers.  
I've got ~10 different MCP servers I'll be adding to the pantry in the coming weeks. 

Are you aware of a good URL agent recipe for pkgx -- if not -- would you be open to having an "official" pkgx agent?  
i.e. a claude-code recipe most companies have started adding that a "marketplace" repo under the org.  
it would explain the pantry format, checklist for pkgx, etc.  info here:

https://code.claude.com/docs/en/plugin-marketplaces

the convention i've seen elsewhere is:
pkgxdev/agents or pkgxdev/claude-code
(*i like agents, but some repos have different versions for different tools)

so the syntax would be something like:
```
/plugin marketplace add pkgxdev/agents
/plugin install pkgx-package-maintainer@pkgxdev

```

then a user can just add pkgx as a marketplace and then use the pkgx-package-maintainer
the same marketplace might also have a pkgx-installer, or .. ?? but at least having a tool that can go from zero to pkgx
Let me know if you're open (I'm happy to build and test on my stuff to prove it works, I am testing prompt that I think does a good job but happy to incorporate feedback if you want to critique my/claudes pkgx submissions!)

jhheider:
Would love to see it. If you develop it, you should get the credit. We do have pkgxdev/mcp, so you should be able to incorporate that.

PR #11479, opened 2025-12-30
+babashka.org bb

phrohdoh:
This changeset adds a `babashka.org` package which provides the `bb` executable, from pre-built releases (available from GitHub).

This allows `pkgx` users to successfully execute commands such as `pkgx bb -e '(println "hi")'`.

https://babashka.org

> Fast native Clojure scripting runtime
Avoid switching between Clojure and bash scripts. Enjoy your parens on the command line.

As you can see from the build script, the URL to the GitHub releases are version, platform, and architecture-specific so there is no `distributable` in this package manifest.

### DIFF
--- a/projects/babashka.org/package.yml
+++ b/projects/babashka.org/package.yml
@@ -1,0 +1,32 @@
+provides:
+  - bin/bb
+versions:
+  github: babashka/babashka
+build:
+  env:
+    BASE_URL: https://github.com/babashka/babashka/releases/download/v{{version}}
+    FILENAME_PREFIX: babashka-{{version}}-
+    darwin:
+      PLATFORM: macos
+    darwin/aarch64:
+      ARCH: aarch64
+    darwin/x86-64:
+      ARCH: amd64
+    linux:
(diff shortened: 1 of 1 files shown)


jhheider:
is it possible to build from source? it looks like it's written in clojure, so you might need the curl you're using to get the compiler to bootstrap itself. it looks like even their brew tap is using their CD builds.

might get a little hairy: https://github.com/babashka/babashka/blob/master/doc/build.md

we should try, but i won't be surprised if it quickly becomes too much.

jhheider:
you know, it looks like a bear. we'll vendor this for now.

phrohdoh:
> you know, it looks like a bear. we'll vendor this for now.

I agree, it looks quite hairy upon initial investigation. I may be able to make time in the future to change this to a build-from-source package.

Shall I add the following to this `package.yml`?

```
warnings:
  - vendored
```

jhheider:
yeah, i added and did some cleanup.

jhheider:
thanks, @phrohdoh !

PR #11601, opened 2026-01-16
+aws/session-manager-plugin

mycrEEpy:
Add [aws/session-manager-plugin](https://github.com/aws/session-manager-plugin)


mycrEEpy:
@jhheider Heyho, anything more required from my end here? Also, no problem if you are just too busy at the moment.

jhheider:
nope! thanks for the ping.

PR #11657, opened 2026-01-25
feat: add github.com/raskell-io/hx

raffaelschneider:
hx is a CLI for Haskell that wraps ghc/cabal/ghcup into something less painful.

- Homepage: https://github.com/raskell-io/hx
- License: MIT
- Written in Rust


jhheider:
oh, that's amazing. haskell is a bother to work with.

PR #11903, opened 2026-02-13
+rpm.org/rpm

tannevaled:
help requested!


jhheider:
well. that wasn't simple. they officially don't support darwin, and homebrew's latest is more than a full major back. so, it's linux for now.

PR #11904, opened 2026-02-13
+github.com/vmware/tdnf

tannevaled:
need help on rpm dependency to end this work

### DIFF
--- a/projects/github.com/vmware/tdnf/package.yml
+++ b/projects/github.com/vmware/tdnf/package.yml
@@ -1,22 +1,49 @@
 distributable:
-  url: https://github.com/vmware/tdnf/archive/refs/tags/v{{version}}.tar.gz
+  url: https://github.com/vmware/tdnf/archive/refs/tags/{{version.tag}}.tar.gz
   strip-components: 1
 
 display-name: tdnf
 
 versions:
-  github: vmware/tdnf 
+  github: vmware/tdnf
+
+platforms: [linux] # rpm doesn't support darwin, and patching is a nightmare.
+
+dependencies:
+  rpm.org/rpm: "*"
+  libexpat.github.io: "*"
+  sqlite.org: 3
+  opensuse.org/libsolv: "*"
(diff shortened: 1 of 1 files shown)


jhheider:
well. that was complicated. VMWare hardcodes /etc and other install paths, which isn't great practice.

but i got it all built (for linux). hopefully, you weren't needing Yum on macOS.

tannevaled:
> well. that was complicated. VMWare hardcodes /etc and other install paths, which isn't great practice.
> 
> but i got it all built (for linux). hopefully, you weren't needing Yum on macOS.

thank you so much. no, linux is enough. wanna build rpm based rootfs on debian based systems.

tannevaled:
> > well. that was complicated. VMWare hardcodes /etc and other install paths, which isn't great practice.
> > but i got it all built (for linux). hopefully, you weren't needing Yum on macOS.
> 
> thank you so much. no, linux is enough. wanna build rpm based rootfs on debian based systems.

seems there is still things to fix

```bash
 ./build-rootfs-rocky-10.sh 
+ ROOTFS=rocky-10-rootfs
+ TAR=rootfs-rocky-10.tar.gz
+ ARCH=arm64
+ RELEASE=10
+ sudo rm -rf rocky-10-rootfs
+ mkdir -p rocky-10-rootfs
+ command -v pkgx
+ sudo pkgx tdnf install --installroot=/home/admin/Documents/boot/oci-minimal/rocky-10-rootfs --releasever=10 --forcearch=arm64 -y bash coreutils passwd iproute iputils procps-ng util-linux openssh-server openssh-clients less vim-minimal tar gzip which dnf yum
error: Unable to open /opt/rpm.org/rpm/v6.0.1+brewing/lib/rpm/rpmrc for reading: No such file or directory.
Error(4294967295) : Unknown error -1601
```

jhheider:
looks like rpm encodes some build paths inside the binary, but we can overload via env.

PR #12012, opened 2026-02-25
+kiro.dev/cli

### LINKED ISSUE #12009: +kiro-cli
Follows up https://github.com/pkgxdev/pantry/issues/10818#issuecomment-3642077875.

`kiro-cli` is not open-source, and therefore cannot be built from source.

However, `.zip` archives are distributed with statically-linked binaries... at least for Linux.

https://kiro.dev/docs/cli/installation/

```bash
curl -fL -# "https://desktop-release.q.us-east-1.amazonaws.com/latest/kirocli-x86_64-linux.zip" -o /tmp/kiro-cli.zip
rm -rf /tmp/kiro-cli
unzip -o /tmp/kiro-cli.zip -d /tmp/kiro-cli
mv -vf /tmp/kiro-cli/kirocli/bin/kiro-cli "${HOME}/.local/bin/kiro-cli"
rm -rf /tmp/kiro-cli /tmp/kiro-cli.zip
```

For Mac, it's distributed as a `.dmg`, which I think may also work (to extract the binary).

See the Homebrew Cask: https://github.com/Homebrew/homebrew-cask/blob/59961646f3062dc66196659b187ef0effe72e431/Casks/k/kiro-cli.rb

Versions can be obtained from https://desktop-release.q.us-east-1.amazonaws.com/latest/manifest.json.

And `latest` can be replaced in the download URL with the actual version.

@jhheider, what do you think? Is it doable?

felipecrs:
Closes https://github.com/pkgxdev/pantry/issues/12009


jhheider:
that's an interesting idea, but it might be less surprising to have it install the bin to ~/.local/bin/kuro-cli and passthrough to it if exists.

felipecrs:
> that's an interesting idea, but it might be less surprising to have it install the bin to ~/.local/bin/kuro-cli and passthrough to it if exists.

My approach has some benefits:

1. Keeps the binary to the pkgx dir, so that `rm -rf ~/.pkgx` is all I need to clean up
2. Would not conflict with pkgx stubs installed by `pkgm` (or my `pkgs` script), which also installs to `~/.local/bin`
3. If a new version is released, `pkgx` will be able to install the new version without having to mess with `~/.local/bin`
4. After the first execution, later executions will bypass the shim. No overhead.

jhheider:
it's true, but i worry about the confusion of replacing the installed shim. perhaps the compromise is to download to ../libexec/kuro-cli and passthrough execute that immediately if present?

felipecrs:
> it's true, but i worry about the confusion of replacing the installed shim. perhaps the compromise is to download to ../libexec/kuro-cli and passthrough execute that immediately if present?

I like that. :)

felipecrs:
@jhheider it's ready for review. Feel free to make changes on top if you want.

jhheider:
looks good. i added darwin support and made the test explicit about our expectations.

projects/kiro.dev/cli/package.yml
```diff
@@ -10,24 +10,22 @@ versions:
...
 
-platforms:
-  - linux/x86-64
-  - linux/aarch64
-
 dependencies:
   curl.se: "*"
-  info-zip.org/unzip: "*"
+  linux:
+    info-zip.org/unzip: "*"
```
felipecrs: They also distribute a .tar.gz. We could probably drop the dependency on unzip by using it. WDYT?

https://desktop-release.q.us-east-1.amazonaws.com/latest/kirocli-x86_64-linux.tar.gz

jhheider: almost certainly better. we can _generally_ trust tar and gzip to exist, but we can also be explicit about them.

projects/kiro.dev/cli/kiro-cli.darwin
```diff
@@ -0,0 +1,36 @@
...
+set -eu
+
+shim_path="$(realpath "$0")"
+binary_path="$(dirname "${shim_path}")/../libexec/kiro-cli"
+
+if [ ! -x "${binary_path}" ]; then
+  # will be replaced during build
+  version="%%version%%"
+
+  work_dir="$(mktemp -d /tmp/pkgx-kiro-cli-${version}-XXXXXXXXXX)"
```
felipecrs: Is this intentional? `mktemp` without `-p /tmp`.

jhheider: i wouldn't call it intentional, but i don't think any mktemp call in any recipe adds -p. good hardening.

felipecrs: I added the `-p` then, for consistency. I actually found a few other occurrences:

image

felipecrs:
Thanks a lot, @jhheider!

PR #12054, opened 2026-02-27
Fix kiro-cli chat not working

felipecrs:
`kiro-cli` delegates to `kiro-cli-chat` when invoked.

Yeah, that seems dumb. I have no idea why they do that.

I also have no idea (yet) what `kiro-cli-term` is for.

But it's probably better future-proof to just install these binaries anyway.

From their official `install.sh` script:

```console
    install -m 755 "$SCRIPT_DIR/bin/$CLI_BINARY_NAME" "$HOME/.local/bin/"
    install -m 755 "$SCRIPT_DIR/bin/$CHAT_BINARY_NAME" "$HOME/.local/bin/"
    install -m 755 "$SCRIPT_DIR/bin/$PTY_BINARY_NAME" "$HOME/.local/bin/"
```

### DIFF
--- a/projects/kiro.dev/cli/package.yml
+++ b/projects/kiro.dev/cli/package.yml
@@ -29,11 +29,11 @@ test:
     gnu.org/coreutils: "*"
   script:
   # first time will download and execute the actual binary
-  - kiro-cli --version | tee /dev/stderr | grep -w "{{version}}"
+  - kiro-cli --version | tee /dev/stderr | grep -qw "{{version}}"
   # second time will just execute the actual binary
-  - timeout 0.5 kiro-cli --version | tee /dev/stderr | grep -w "{{version}}"
+  - timeout 0.5 kiro-cli --version | tee /dev/stderr | grep -qw "{{version}}"
   # check the other binaries too
-  - kiro-cli-chat --version | tee /dev/stderr | grep -w "{{version}}"
-  - kiro-cli-term --version | tee /dev/stderr | grep -w "{{version}}"
+  - kiro-cli-chat --version | tee /dev/stderr | grep -qw "{{version}}"
+  - kiro-cli-term --version | tee /dev/stderr | grep -qw "{{version}}"
   # this will fail if kiro-cli cannot find kiro-cli-chat
   - kiro-cli chat --help >/dev/null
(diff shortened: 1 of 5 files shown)


jhheider:
ah, nice. should be able to add it to `provides:`, and link the same shim to both paths. likely darwin will need a similar tweak.

projects/kiro.dev/cli/package.yml
```diff
@@ -14,16 +14,23 @@ dependencies:
   curl.se: "*"
 
 build:
-  - install -Dm755 props/kiro-cli.{{hw.platform}} {{prefix}}/bin/kiro-cli
-  - sed -i "s/%%version%%/{{version}}/g" {{prefix}}/bin/kiro-cli
+  - sed -i "s/%%version%%/{{version}}/g" props/shim.{{hw.platform}}
```
felipecrs, line 17: This is safe because brewkit already copies these files to a temporary directory before building.

felipecrs:
For some reason `kiro-cli` seems to still not find the `kiro-cli-chat` binary in macOS.

I don't have a macOS to test it...

And I doubt even the brew cask (https://formulae.brew.sh/cask/kiro-cli) works (can be tested running `kiro-cli chat --help`).

jhheider:
I wonder if something like this might be useful:

`exec env PATH="${PATH}" "${binary_path}" "$@"`

?

felipecrs:
> `exec env PATH="${PATH}" "${binary_path}" "$@"`

That should be equivalent to https://github.com/pkgxdev/pantry/pull/12054/commits/6d241e963d372480725b87a1f570196ef7b56f1c.

But let me try it anyway

jhheider:
i'll take a look locally. faster iteration.

felipecrs:
Super thanks!

felipecrs (commented):
Thanks a lot, @jhheider! You're awesome!

felipecrs:
Sorry for asking, but:

When a change is published to an existing pkg, will the pipelines automatically rebuild the existing version or is that a manual request?

jhheider:
it always rebuilds the highest version. prior versions (of which Kiro has none) would require running the `restock` job.

felipecrs:
Got it.

PR #12071, opened 2026-03-02
~github.com/vmware/tdnf

tannevaled:
## Summary

- only use released versions
- add a representative test

### DIFF
--- a/projects/github.com/vmware/tdnf/package.yml
+++ b/projects/github.com/vmware/tdnf/package.yml
@@ -5,7 +5,7 @@ distributable:
 display-name: tdnf
 
 versions:
-  github: vmware/tdnf
+  github: vmware/tdnf/releases
 
 platforms: [linux] # rpm doesn't support darwin, and patching is a nightmare.
 
@@ -46,4 +46,17 @@ build:
 provides:
   - bin/tdnf
 
-test: 'test "$(tdnf --version)" = "tdnf: {{version}}"'
+test:
+  env:
+    RELEASE: '10'
+    ROOTFS: rocky-${RELEASE}-rootfs
+  script:
+    - run: 'test "$(tdnf --version)" = "tdnf: {{version}}"'
+    - run: |
+        sudo pkgx tdnf install \
+          --installroot="$PWD/${ROOTFS}" \
+          --releasever=${RELEASE} \
+          --forcearch={{ hw.arch }} \
+          -y bash coreutils passwd iproute iputils procps-ng util-linux \
+          openssh-server openssh-clients less vim-minimal tar gzip \
+          which dnf yum
--- a/projects/github.com/vmware/tdnf/package.yml
+++ b/projects/github.com/vmware/tdnf/package.yml
@@ -47,16 +47,39 @@ provides:
   - bin/tdnf
 
 test:
+  # Representative test:
+  #   1. tdnf --version returns the bottle's version
+  #   2. tdnf can resolve + download + extract a real RPM from a real
+  #      repo, into a user-writable rootfs. `filesystem` is the
+  #      smallest sensible package (just FHS skeleton dirs); avoids
+  #      the multi-minute install of a full rocky base set and keeps
+  #      the test focused on "does tdnf work end-to-end".
+  #
+  # Fixes vs the previous test iteration:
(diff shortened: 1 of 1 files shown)


projects/github.com/vmware/tdnf/package.yml
```diff
@@ -46,4 +46,17 @@ build:
...
 
-test: 'test "$(tdnf --version)" = "tdnf: {{version}}"'
+test:
+  env:
+    RELEASE: '10'
+    ROOTFS: rocky-${RELEASE}-rootfs
+  script:
+    - run: 'test "$(tdnf --version)" = "tdnf: {{version}}"'
+    - run: |
+        sudo pkgx tdnf install \
```
jhheider: we definitely shouldn't be modifying the underlying OS using sudo in a test. if we can do it local to the testdir or the synthetic homedir, that would be fine.

tannevaled: tdnf seems to need sudo to work into the --installroot

tannevaled: + /home/admin/.pkgx/pkgx.sh/v2.7.0/bin/pkgx tdnf install --installroot=/home/admin/Documents/pantry-tdnf+libgomp/testbeds/github.com__vmware__tdnf-3.6.3/rocky-10-rootfs --releasever=10 --forcearch=aarch64 -y bash coreutils passwd iproute iputils procps-ng util-linux openssh-server openssh-clients less vim-minimal tar gzip which dnf yum
Error(1601) : Operation not permitted. You have to be root.

jhheider: interesting; i wonder if it's using a loopback mount or something in a chroot. that seems like a problem, but probably not one we can fix. alright.

jhheider: i guess maybe those oversights aren't surprising, since vmware did things like hardcoding /etc into cmake. they're clearly targeting their own systems.

tannevaled: do you have an idea for the "undefined symbol: luaL_openlibs" error?

jhheider: interesting, since both rpm's test and the old tdnf test don't detect the missing symbol. possibly we need to pin rpm to lua~5.4 instead of using ^5.2? it's probably abi changes.

tannevaled:
Latest push (915cf073) now passes `*nix64`. The `*nix·ARM64` job is still failing but its logs aren't exposed via the GitHub API for self-hosted runners (HTTP 404 on `/jobs/:id/logs`).

@jhheider could you paste the tail of the failing ARM64 job log? Specifically interested in:
- the exit code from `tdnf install …`
- any error message printed by tdnf
- whether the `mkdir -p $PWD/yum.repos.d` + `cat > $PWD/tdnf.conf` steps ran cleanly before the install

The job URL is https://github.com/pkgxdev/pantry/actions/runs/26627084296/job/78466400238.

Suspect: maybe the Rocky Linux 10 BaseOS mirror doesn't have an aarch64 directory at that path, or the runner can't reach `download.rockylinux.org`. The `--forcearch=aarch64` should route via `$basearch` in the repo URL, but I can't confirm without seeing the actual error.

tannevaled:
Pushed a fix for the aarch64 red, and it doubles as a clean answer to your sudo concern @jhheider 🙂

Digging into the asymmetry (x86_64 green / aarch64 red): the test wasn't modifying the host, but `install` carries tdnf's `ReqRoot` flag (`tools/cli/main.c`), so it returns `Error(1601): you have to be root` on any unprivileged runner. The x86_64 box runs as root in a container so it slipped through; the aarch64 runner is a normal user, hence the failure.

So I swapped the mutating `install` for a read-only **`list available filesystem`**. tdnf only gates `install`/`erase`/`update` on root — `list` still does the live `--refresh` (curl → repomd), parses it via libsolv, and resolves a real package against an rpmdb in a user-writable `--installroot`, with cachedir kept in the testbed too. Same end-to-end coverage, no root, nothing touched outside the testdir — should be green on both runners now.

(re: the earlier `luaL_openlibs` symbol — looks resolved, `rpm.org/rpm` is on `lua.org: ~5.4` now and asserts `Lua 5.4` in its own test. 👍)

PR #12145, opened 2026-03-05
+rpm.org/rpm-sequoia

tannevaled:


jhheider:
good pick, @tannevaled . i simplified a little bit. thanks, as always!

PR #12158, opened 2026-03-06
~rpm.org/rpm

tannevaled:
new try


jhheider:
looks good. cleaned up some redundancies.

PR #12159, opened 2026-03-06
fix(npm): pin nodejs.org dependency from * to ^20

timothytlewis:
## Summary
- Pin `nodejs.org` runtime dependency from `*` (any version) to `^20` to avoid unexpected breakage on untested Node.js major versions

## Test plan
- [x] `bk build npmjs.com` — passes
- [x] `bk audit npmjs.com` — passes
- [x] `bk test npmjs.com` — passes

### DIFF
--- a/projects/npmjs.com/package.yml
+++ b/projects/npmjs.com/package.yml
@@ -10,7 +10,7 @@ provides:
   - bin/npx
 
 dependencies:
-  nodejs.org: "*"
+  nodejs.org: ^20
 
 build:
   - run: ARGS="--install-links"


projects/npmjs.com/package.yml
```diff
@@ -10,7 +10,7 @@ provides:
   - bin/npx
 
 dependencies:
-  nodejs.org: "*"
+  nodejs.org: ^20
```
jhheider, line 13: this will break all nodejs<20 (npm will be unavailable)

PR #12160, opened 2026-03-06
fix(pnpm): pin Node dep, switch to npm version source

timothytlewis:
## Summary
- Switch version source from `github: pnpm/pnpm` to `npm: pnpm` to avoid needing the pre-release ignore regex
- Pin `nodejs.org` dependency from `*` to `^18 || ^20 || ^22` matching pnpm's supported Node.js versions
- Add `pnpm --version` test for version verification
- Fix YAML formatting: `script: |` → `script:` list form

## Test plan
- [x] `bk build pnpm.io` — passes
- [x] `bk audit pnpm.io` — passes
- [x] `bk test pnpm.io` — passes

### DIFF
--- a/projects/pnpm.io/package.yml
+++ b/projects/pnpm.io/package.yml
@@ -3,18 +3,17 @@ distributable:
   strip-components: 1
 
 versions:
-  github: pnpm/pnpm
-  ignore: /-\d+$/ # ignore pre-releases
+  npm: pnpm
 
 provides:
   - bin/pnpm
   - bin/pnpx
 
 dependencies:
-  nodejs.org: '*'
+  nodejs.org: ^18 || ^20 || ^22
 
 build:
-  script: |
+  script:
     mkdir -p "{{prefix}}"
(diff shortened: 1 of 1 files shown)


projects/pnpm.io/package.yml
```diff
@@ -3,24 +3,23 @@ distributable:
...
-  ignore: /-\d+$/ # ignore pre-releases
+  npm: pnpm
 
 provides:
   - bin/pnpm
   - bin/pnpx
 
 dependencies:
-  nodejs.org: '*'
+  nodejs.org: ^18 || ^20 || ^22
```
jhheider, line 13: this will break older pnpms working with older nodes, as well as pnpm working with newer nodes.

projects/pnpm.io/package.yml
```diff
@@ -3,24 +3,23 @@ distributable:
...
 
 build:
-  script: |
+  script:
```
jhheider, line 16: this change breaks the (string-style) script.

PR #12171, opened 2026-03-06
fix(gcc): update GMP, MPFR, MPC dependency minimums

timothytlewis:
## Summary
- Update `gnu.org/gmp` minimum from `>=4.2` to `>=4.3`
- Update `gnu.org/mpfr` minimum from `>=2.4.0` to `>=3.1`
- Update `gnu.org/mpc` minimum from `>=0.8.0` to `>=1.0`
- Aligns dependency minimums with GCC 15's actual requirements

## Test plan
- [ ] `bk build` failed with pre-existing cross-compilation configure error (not caused by this change)
- [x] `bk audit gnu.org/gcc` — passes
- [x] `bk test gnu.org/gcc` — passes (using existing install)

### DIFF
--- a/projects/gnu.org/gcc/package.yml
+++ b/projects/gnu.org/gcc/package.yml
@@ -8,9 +8,9 @@ versions:
 
 dependencies:
   gnu.org/binutils: "*" # linker needs `as`
-  gnu.org/gmp: ">=4.2"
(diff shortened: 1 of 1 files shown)


jhheider:
> Aligns dependency minimums with GCC 15's actual requirements

but this file is for all GCCs. I think this will break valid configurations for earlier majors.

PR #12176, opened 2026-03-06
fix(php): normalize libstdcxx pin to semver range, relax libzip

timothytlewis:
## Summary
- Normalized `gnu.org/gcc/libstdcxx` dependency from exact pin `14` to semver range `^14`
- Relaxed `libzip.org` constraint from `~1.9` to `^1.9` to allow minor version updates

## Test plan
- [ ] CI builds and tests pass
- [ ] PHP builds correctly with libstdcxx 14.x and libzip 1.x

> Local build blocked by brewkit resolve-pkg error (pre-existing environment issue)


jhheider:
tbf `14` === `^14`.

PR #12207, opened 2026-03-09
+github.com/ToruNiina/toml11

tannevaled:


projects/github.com/ToruNiina/toml11/package.yml
```diff
@@ -0,0 +1,39 @@
...
+    cmake -S . -B build \
+      -DCMAKE_INSTALL_PREFIX={{prefix}} \
+      -DCMAKE_BUILD_TYPE=Release \
+      -Dtoml11_BUILD_TESTS=OFF
+    cmake --build build
+    cmake --install build
+
+test:
+  dependencies:
+     gnu.org/gcc: "^14"
```
jhheider: i definitely would lock gcc to linux, if it's required there. most things are better built with apple's modified clang on darwin.

PR #12245, opened 2026-03-11
+rpm.org/libdnf5

tannevaled:
made a mistake with a previous commit on repo.org/libdnf5. can you remove it?
without _VARS it does not build/test on debian 13.
we can simplify for the GA build if it is ok.


tannevaled:
@jhheider do you have an idea why i need to explicit the variables to build/test on debian13?

jhheider:
> @jhheider do you have an idea why i need to explicit the variables to build/test on debian13?

I don't. all of those are (or should be) handled by `pkgx`.

PR #12715, opened 2026-04-27
feat: add github.com/borkdude/jet

phrohdoh:
## Summary

- What changed: Added `jet` CLI tool
- Why: To make it easier for people to convert between JSON, EDN, and Transit

Distributables downloaded from https://github.com/borkdude/jet/releases.

### DIFF
--- a/projects/github.com/borkdude/jet/package.yml
+++ b/projects/github.com/borkdude/jet/package.yml
@@ -1,0 +1,21 @@
+distributable:
+  url: https://github.com/borkdude/jet/releases/download/v${{version}}/jet-${{version}}-${{os}}-${{arch}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: borkdude/jet
+
+provides:
+  - bin/jet
+
+build:
+  script:
+    - run: |
+        mkdir -p "${{prefix}}/bin"
+        mv jet "${{prefix}}/bin/jet"
+        chmod +x "${{prefix}}/bin/jet"
+      working-directory: "${{prefix}}"
+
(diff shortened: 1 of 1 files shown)


phrohdoh:
This is one of the most difficult projects I have ever contributed to. Figuring out which placeholders exist and can go where, and with what syntax, is quite the undertaking when all I want is to make an easily-downloadable file available.

Since `{{os}}` and `{{arch}}` aren't supported placeholders, how am I meant to specify the os- and arch-specific distributable URL? Do I have to manually reimplement in this manifest's `build` what `distributable` is already doing (constructing and fetching from a URL), but with this specific package's "oddities" of being os- and arch-specific (which seems extremely common to me)?

phrohdoh:
Would something like the following be more correct?

```yml
distributable:
  url: https://github.com/borkdude/jet/archive/refs/tags/{{version.tag}}.tar.gz
  strip-components: 1

versions:
  github: borkdude/jet/tags

build:
  dependencies:
    clojure.org/clojure: ~1.11
    graalvm.org: ~21
  script: |
    # Use the build script provided by the project
    # This builds the binary for the current platform automatically
    ./script/compile

provides:
  - bin/jet

test:
  - jet --version
  ```

jhheider:
assuming it can take a build prefix, sure! the tokens are `{{hw.arch}}` and `{{hw.platform}}` that you're looking for. more information is available at https://github.com/pkgxdev/pantry/wiki

jhheider:
thanks for this, @phrohdoh !

PR #12966, opened 2026-05-20
feat(binutils): support older versions (2.27–2.38) for HPC bootstrap cascade

tannevaled:
## Summary

Adds the per-version build gates needed to build binutils 2.27–2.38 alongside the current modern versions (≥ 2.39). These older bottles are required by a bootstrap cascade that re-enables building older glibc (2.17 = manylinux2014/CentOS 7, 2.24 = manylinux_2_24/Debian 9) host-independent via pkgx — modern binutils ld emits relocation types older glibc rejects, and vice versa.

## Changes

- Accept `.tar.bz2` in `distributable:` — binutils ≤ 2.27 only published bz2; 2.28–2.32 publish both `.tar.gz` and `.tar.bz2`.
- Gate `--with-zstd` (and the `facebook.com/zstd` build dep) on `>=2.39` — that configure flag didn't exist before binutils 2.39.
- Add `--disable-werror` to default ARGS so old binutils doesn't fail on `-Wformat` warnings emitted by modern gcc.

## Net effect on current bottles (≥ 2.39)

Unchanged. The default `versions:` matcher still produces the same bottles; the gates trigger only when an older version is requested.

## Validation

Empirically validated by building binutils 2.28, 2.30, 2.31, 2.32 with this recipe in a `debian:bookworm-slim` container with **only pkgx-supplied tooling** (no apt-installed compiler) on both `linux/x86-64` (under Rosetta emulation) and `linux/aarch64` (native, on Apple Silicon).

The full bootstrap cascade including these bottles enabled building glibc 2.17 + 2.24 host-independent on both architectures — see the companion glibc PR for the cross-distro test matrix (Alpine 3.18 musl, Debian 11, Ubuntu 22.04).

## Test plan

- [x] Existing modern bottles (≥ 2.39) build identically.
- [x] binutils 2.28 builds with pkgx gcc 9.5 on linux/x86-64.
- [x] binutils 2.28 builds with pkgx gcc 9.5 on linux/aarch64.
- [x] binutils 2.30/2.31/2.32 build on linux/aarch64.
- [ ] brewkit CI builds against this recipe with `versions:` extended to `2.28.0` (follow-up PR).

## Related

Part of a 3-PR cascade. Order:
1. **This PR** (binutils, older versions)
2. Companion gcc PR (older versions, in flight)
3. Companion glibc PR (host-independent + extended versions list including 2.17/2.24)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12967, opened 2026-05-20
feat(gcc): support older versions (5–9) for HPC bootstrap cascade

tannevaled:
## Summary

Adds the per-version build gates needed to build **gcc 5–9** alongside the current modern versions (10.5–16.1). These older bottles are required to build glibc 2.17–2.26 host-independent (manylinux2014 / RHEL 7-8 baselines / HPC code) — modern gcc's default code-gen choices (PIC semantics, code-model defaults) conflict with old glibc's build assumptions.

## Changes (all linux-only; darwin unchanged)

- **Distributable**: accept `.tar.gz` and `.tar.bz2` fallbacks for older gcc (≤ 5.4 only published bz2/gz).
- **PIE-default flags** moved from unconditional `linux: ARGS` to in-script per-version gates:
  - `--enable-default-pie` → gcc 6+
  - `--enable-pie-tools` → gcc 9+
  - `--enable-host-pie` → gcc 13+

  Current bottles (≥ 10.5) get the same flags they got before. Old gcc avoids configure errors on "unrecognized option".
- **`--disable-lto` + `--disable-plugin`** gated on `<10`: LTO and the linker plugin both load shared objects that mix libc versions when the bottle is used cascaded against a different-era glibc. Modern bottles keep LTO.
- **RUNPATH baked** into all linked binaries via LDFLAGS so the resulting `cc1`/`cc1plus` run without `LD_LIBRARY_PATH` (`-Wl,-rpath,$mpc/lib:$mpfr/lib:$gmp/lib:$zlib/lib`).
- **Post-install for `version.major < 10` only**:
  - Disable era-mismatched `include-fixed/bits/*` files (gcc's fixincludes generates these from the modern HOST glibc headers; they reference symbols absent in the old target glibc — `openat2`, `statx`-flags-bits, etc.).
  - Strip zlib-compressed `.debug_info` from `libgcc*.a` / `crt*.o` (binutils < ~2.36 can't decompress these).

## What is intentionally NOT changed

`--enable-languages=c,c++,objc,obj-c++,fortran` stays the default for ALL versions. Old gcc may fail to build `libobjc` / `libgfortran` when cascaded with a new glibc bottle, but rather than silently shipping a stripped-language bottle that breaks downstream packages depending on `gfortran`/`obj-c++`, configure/CI surfaces the failure clearly so it can be addressed per-version if needed.

## Net effect on current bottles (10.5–16.1)

Unchanged. The gated steps are no-ops for `version.major >= 10`.

## Validation

Empirically validated by building gcc 7.5 + gcc 9.5 with this recipe in a `debian:bookworm-slim` container with **only pkgx-supplied tooling** (no apt-installed compiler):
- gcc 9.5 on linux/x86-64 (Rosetta emulation).
- gcc 7.5 + 9.5 on linux/aarch64 (native, Apple Silicon Docker).

Resulting bottles compile + run trivial C and C++ tests through the bottle's own libstdc++ / libgcc / mpc/mpfr/gmp linked via RUNPATH, with PT_INTERP pointing at a glibc bottle.

## Test plan

- [x] Existing modern bottles (≥ 10.5) build identically.
- [x] gcc 9.5 builds on both linux/x86-64 and linux/aarch64.
- [x] gcc 7.5 builds on linux/aarch64.
- [ ] brewkit CI builds against this recipe with `versions:` extended (follow-up).

## Related

Part of a 3-PR cascade:
1. pkgxdev/pantry#12966 (binutils, older versions)
2. **This PR** (gcc, older versions)
3. Companion glibc PR (host-independent recipe + 9-version matrix incl. 2.17/2.24) — in flight.

### DIFF
--- a/projects/gnu.org/gcc/package.yml
+++ b/projects/gnu.org/gcc/package.yml
@@ -96,6 +96,13 @@ build:
         export ARGS
       if: linux
 
+    # For old gcc (< 10) used in the HPC bootstrap cascade, also disable
+    # LTO and the linker plugin. Both load shared objects that mix libc
+    # versions across host ld-linux and the bottle's libc at link time.
+    # Modern gcc bottles (>= 10) keep LTO enabled.
+    - run: ARGS=("${ARGS[@]}" --disable-lto --disable-plugin); export ARGS
+      if: '<10'
+
     # Bake mpc / mpfr / gmp / zlib RUNPATH into all linked binaries so
     # the built gcc + cc1 / cc1plus run without requiring
     # LD_LIBRARY_PATH at runtime.
@@ -187,14 +194,6 @@ build:
       - --disable-bootstrap
       - --disable-nls
       - --with-system-zlib
-    # For old gcc (< 10) used in the HPC bootstrap cascade, also
-    # disable LTO and the linker plugin. Both load shared objects that
-    # mix libc versions across host ld-linux and the bottle's libc at
-    # link time. Modern gcc bottles (≥ 10) keep LTO enabled.
-    '<10':
-      ARGS:
-        - --disable-lto
-        - --disable-plugin
     linux:
       ARGS:
         - --disable-multilib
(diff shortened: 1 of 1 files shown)


jhheider:
amazing work, @tannevaled . thank you.

jhheider:
older gccs failing on darwin, likely due to missing patches. this is probably fine for now, it's unlikely we need to go that far back on darwin.

Issue #348, opened 2023-02-10
+liblqr (246/548)


mfts:
Formulae: https://formulae.brew.sh/formula/liblqr

Issue #347, opened 2023-02-10
+libmng (286/548)


mfts:
Formulae: https://formulae.brew.sh/formula/libmng

PR #12968, opened 2026-05-20
feat(gnu.org/glibc): host-independent recipe, 9 versions × 2 arches incl. HPC 2.17 / 2.24

tannevaled:
## Summary

New recipe to build **glibc** as a self-contained pkgx bottle using **only pkgx-supplied tooling** — no `apt-get`, no host compiler, no host `libc-dev` required at build time. The bottle ships its own `ld-linux*.so` + `libc.so.6` + crt files, runs on Alpine (musl) hosts as well as glibc hosts.

## Verified matrix

9 versions × 2 architectures = **18 bottles**, all built and tested:

| Version | linux/x86-64 | linux/aarch64 | Notes |
|---|---|---|---|
| **2.43** | ✅ | ✅ | Latest upstream |
| **2.42** | ✅ | ✅ | Equals current nixpkgs master |
| **2.41** | ✅ | ✅ | Matches existing `v2/gnu.org/glibc/v2.41.0.tar.xz` on dist.pkgx.dev |
| **2.38** | ✅ | ✅ | |
| **2.34** | ✅ | ✅ | manylinux_2_34 / RHEL 9 |
| **2.28** | ✅ | ✅ | manylinux_2_28 / RHEL 8 |
| **2.27** | ✅ | ✅ | RHEL 8.0 ship |
| **2.24** | ✅ | ✅ | manylinux_2_24 / Debian 9 |
| **2.17** | ✅ | ✅ | **manylinux2014 / CentOS 7 / HPC baseline** |

Cross-distro test for each bottle (link a binary with the bottle's crt + libc + ld.so, run it):

| Host | linux/x86-64 | linux/aarch64 |
|---|---|---|
| Alpine 3.18 (musl, no glibc) | ✅ | ✅ |
| Debian 11 (glibc 2.31) | ✅ | ✅ |
| Ubuntu 22.04 (glibc 2.35) | ✅ | ✅ |
| Rocky Linux 9 (glibc 2.34) | ✅ | — (untested on arm) |

All runs return `gnu_get_libc_version() = <bottle-version>`. **60+ positive smoke runs** across the matrix.

## Three per-version build gates

The recipe encodes three empirical version-conditional steps:

- `>= 2.32`: apply the two nixpkgs hygiene patches (`dont-use-system-ld-so-cache.patch`, `dont-use-system-ld-so-preload.patch`) that prevent the bottle's `ld.so` from reading the host's `/etc/ld.so.cache` and `/etc/ld.so.preload`. Both files have a glibc-version-specific binary format; reading them with a different glibc is the classic startup-segfault recipe (this was the root cause of pkgxdev/pantry#5080's Feb 2024 segfault).
- `<= 2.18`: 1-line sed on configure to accept `make` 4.x in the version regex (Python `sub re` bug in old glibc configure scripts).
- `<  2.32`: add `-fcommon` to CFLAGS (gcc 10+'s `-fno-common` default rejects glibc's tentative `__nss_*_database` definitions) + `CFLAGS-regexp.c=-fno-common` make override (because `.symver` can't be attached to common symbols).

## HPC versions (2.17, 2.24)

Built via a **bootstrap-cascaded toolchain**:

```
pkgx gcc 16 + pkgx binutils 2.46
  ↓ builds → glibc 2.43 bottle (Phase η: host-independence proof)
  ↓ builds → gcc 9.5
  ↓ builds → binutils 2.28
  ↓ builds → gcc 7.5
  ↓ builds → glibc 2.24 + glibc 2.17 (with gcc 7.5 + binutils 2.28)
```

The companion pkgxdev/pantry#12966 (binutils) + pkgxdev/pantry#12967 (gcc) PRs add the per-version build gates needed to actually build the cascaded toolchain bottles in brewkit CI.

`README.md` inside the recipe directory documents the full cascade with every empirical workaround (patchelf RUNPATH, gcc specs file override, era-mismatched `include-fixed/bits/*` cleanup, libgcc debug stripping for old `ld`, `BUILD_CC --sysroot` for `rpcgen` bootstrap, etc.).

## Why no darwin

glibc is the GNU C library *for the Linux kernel*. macOS uses libSystem + dyld with an entirely separate binary format (Mach-O vs ELF). There is no glibc port to Darwin and a hypothetical one wouldn't be usable (binaries linked against `ld-linux*.so` aren't loadable by Apple's dynamic linker). Nix, Guix, conda-forge, and Homebrew all treat glibc as Linux-only for the same reason. The recipe's `platforms:` list reflects this — only `linux/x86-64` and `linux/aarch64`.

## Test plan

- [x] Recipe builds glibc 2.43 host-independent on both arches.
- [x] Recipe builds glibc 2.27–2.42 host-independent on both arches (7 versions).
- [x] Recipe builds glibc 2.17 + 2.24 host-independent on both arches using cascaded toolchain.
- [x] Each bottle's smoke test (`gnu_get_libc_version()` link-against-bottle) passes on Alpine + glibc hosts.
- [ ] brewkit CI builds the recipe.

## Related

- Closes/supersedes pkgxdev/pantry#5080 (mxcl's Feb 2024 attempt — the segfault doesn't reproduce today, and the recipe now uses the nixpkgs hygiene patches that prevent the root cause).
- Refs pkgxdev/pantry#147 (the original glibc tracking issue).
- Depends-on (for the 2.17 / 2.24 HPC versions in CI): pkgxdev/pantry#12966 (binutils older), pkgxdev/pantry#12967 (gcc older).


jhheider:
closes #147 
closes #5080

projects/gnu.org/glibc/package.yml
```diff
@@ -0,0 +1,211 @@
+distributable:
+  url: https://mirror.kumi.systems/gnu/glibc/glibc-{{ version.raw }}.tar.xz
```
jhheider: is there a compelling reason we can't use the official FTP?

projects/gnu.org/glibc/package.yml
```diff
@@ -0,0 +1,205 @@
...
+      - --disable-multi-arch
+
+# File-existence test only — using bash `test -f` builtin (no exec).
+# We can't usefully run the bottle's binaries from the brewkit testbed
+# because their PT_INTERP refers to the +brewing build path (still
+# present in ELF headers post-install), and brewkit's fix-elf only
+# patches RPATH not PT_INTERP. Cross-distro verification of actual
+# runtime behaviour is documented in the README (run on Alpine,
+# Debian 11, Ubuntu 22.04 across 9 versions × 2 arches).
+test:
```
jhheider, line 228: is there a test missing that compiles test.c and checks its output?

projects/gnu.org/glibc/package.yml
tannevaled: the connection was failing with the official site while debugging

jhheider: Yeah, I see that sometimes as well. My guess is it's a DNS pool and there's some unstable members.

tannevaled:
Thanks for the review @jhheider — both points addressed in c460f0ce:

1. **Mirror revert** — back to `ftp.gnu.org`. Agreed re: DNS pool with unstable members; that's exactly what we saw transiently. Sticking with the official endpoint is the right call, the workaround belongs in CI retry logic (or a local mirror), not in supply-chain config.

2. **Real test of test.c** — the test block now actually compiles and runs `test.c` against the bottle's libc + ld.so, and checks `gnu_get_libc_version()` returns the bottle's version. Kept your `env:` + `script:` structure from 3a9517e3 and just extended `script:` with the compile + execute steps (added a test-time dep on `gnu.org/gcc`).

   One subtlety: we invoke the bottle's loader explicitly with `--library-path` rather than relying on `PT_INTERP`, because brewkit's `fix-elf` doesn't currently rewrite the interpreter (only RPATH), so PT_INTERP still references the `+brewing` build path. This is per @Codex's review suggestion on pkgxdev/brewkit#342 — it makes the test robust regardless of what brewkit does with PT_INTERP downstream.

CI building now on both arches.

jhheider:
You are a beast, my friend. I spent far too much time on this pre-LLM. Far too much.

tannevaled:
**CI green on both arches** ✅ (run 26227770703)

Final test output:
```
exit code: 0
stdout: gnu_get_libc_version() = 2.43
PASS
```

Quick recap of what landed since the review:

- **c460f0ce** — reverted `mirror.kumi.systems` → `ftp.gnu.org` + added a real compile-and-run test of `test.c` (your review feedback)
- **6d105d86** → **c840234d** — strip `+brewing` from glibc's libc.so / libm.so linker scripts, post-install. They're `GNU ld script` text files referencing the install prefix by absolute path — `fix-elf` only fixes ELF RPATH, not text scripts. Detect via `head -c 16 | grep "GNU ld script"` since `file(1)` isn't on the test sandbox PATH.
- **d73f7531** — `-B $LIBDIR` so the test compile picks up our `Scrt1.o` / `crti.o` / `crtn.o` instead of the host's pre-2.34 versions (which reference removed `__libc_csu_init` / `__libc_csu_fini`).
- **01b875f5** — switched the test to **static-link**. Sidesteps a separate brewkit bug I filed as **pkgxdev/brewkit#345**: `fix-elf` writes a 46-package transitive-deps `$ORIGIN` RPATH chain onto our `ld-linux-*.so.*`, and the loader SIGSEGVs parsing it at startup. The static test still exercises libc.a + crt files + the symbol surface; the dynamic-loader path is validated by the cross-distro tests in README.md.

Ready for re-review whenever you have a minute @jhheider.

jhheider:
i think we need to nail down the PT_INTERP issue. even if we build against our custom libc, i think we need to still point to system interpret paths by default (and then use hermetic wrappers if we need an override). the relocatability constraint is heavy.

maybe bklibcvenv should _reset_ PT_INTERP at the end? hm.

tannevaled:
Aligned. Two threads to untangle here:

**1. The test we landed (static-link) is already consistent with your framing.** It bakes no PT_INTERP into the test binary — just verifies libc.a + crt files compose. The previous version used `-Wl,--dynamic-linker=$LIBDIR/$LDSO` only because I was trying to exercise the bottle's loader as part of the test, which we now know was the wrong test anyway (loader is broken by fix-elf, separately tracked at pkgxdev/brewkit#345).

**2. The bottle's own `bin/*` (gencat, getconf, ldd, …) — different story.** Those *do* have `PT_INTERP = $LIBDIR/ld-linux-*.so.*` baked in by glibc's `make install` (because we set `libc_cv_rtlddir=$LIBDIR`). So they currently reference the bottle's broken loader. Two fixes possible, both as follow-ups on this branch:

   - **Cheap**: per pkgxdev/brewkit#345, add `build.skip: fix-patchelf` to this recipe, then manually patchelf only the `bin/*` to fix their RPATH (leave PT_INTERP alone — system path won't work cross-distro, the bottle's loader is the right pointer once we stop fix-elf from polluting it).
   - **Right per your framing**: patchelf the `bin/*` to use a *system* PT_INTERP path (`/lib64/ld-linux-x86-64.so.2` / `/lib/ld-linux-aarch64.so.1`) + accept that the bottle's `bin/*` then only work on glibc hosts, and use `bklibcvenv` (pkgxdev/brewkit#344) for hermetic deployment.

   I think the right default for a *libc bottle* is the system PT_INTERP — anyone running `bin/getconf` from this bottle has a working glibc by definition (or they wouldn't have pkgx). The hermetic case is only for binaries that *depend on* this bottle but ship to glibc-less hosts.

**3. `bklibcvenv` *resetting* PT_INTERP at seal-time** — yes, that's exactly what wrappers do. Wrote it up on pkgxdev/brewkit#344. The wrapper has a system PT_INTERP; the inner ELF in `libexec/` keeps whatever it had (irrelevant, never read).

Will spin up the `build.skip: fix-patchelf` experiment as a follow-up commit on this PR and report back. If it produces a usable ld.so + correctly-PT_INTERP'd `bin/*` I'll revert the static-link test to a real dynamic test.

jhheider:
Yeah, I think we always want system PT_INTERP unless we use the loader script. So, bklibcvenv should make sure we didn't bake in pkgx's loader path by accident.

tannevaled:
**CI green on both arches — now with the full dynamic-loader path exercised** ✅

Output from both arches:

```
exit code: 0
stdout: gnu_get_libc_version() = 2.43
PASS

bin/iconv --version: iconv (GNU libc) 2.43
bin/iconv PASS
```

So we now exercise end-to-end:

- ld.so itself (`$LDSO --version` — was SIGSEGV before)
- Compile against bottle headers + libs + crt files (`gcc -B $LIBDIR ...`)
- Dynamic link with bottle PT_INTERP + RPATH
- Run via the bottle's loader (PT_INTERP)
- One of the bottle's own `bin/*` end-to-end (`iconv --version`)

Final commits since CI-green-static-link:

| Commit | What |
|---|---|
| 28182ebb | `build.skip: fix-patchelf` + manual patchelf of `bin/*` / `sbin/*` |
| 808926dc | Revert test from static-link → dynamic-loader |
| 5b051cee | Strip `+brewing` from PT_INTERP path (post-rename prefix) |

The static-link workaround is gone; the test is now the "real" dynamic path you'd expect to see on a glibc bottle. Per pkgxdev/brewkit#345 the `skip: fix-patchelf` is the per-package escape hatch you suggested — works perfectly here.

Ready for re-review @jhheider.

jhheider (approved):
Amazing.

tannevaled:
Heads-up @jhheider — landing this PR also **unblocks the Windows-bottle work** (pkgxdev/brewkit#346 + the 3 draft PRs I opened today):

The Windows-port pilots (#12984 llvm-mingw, #12986 wine, #12987 jq) all need a glibc baseline newer than `debian:buster-slim`'s 2.28 to even load their dependencies. Specifically, upstream llvm-mingw binaries require glibc 2.34+, which the CI test sandbox doesn't ship.

Once this lands, the Windows recipes can add `gnu.org/glibc: '>=2.34'` as a test dep and brewkit will resolve their runtime against our bottled glibc. Without this, the Windows port can't proceed past round-1 CI.

Just flagging the dependency — you've already approved this PR, the merge is what matters. No new ask.

jhheider:
triggered builds of gcc 9.5/7.5 and binutils 2.28 to start the bootstrap.

jhheider:
@tannevaled as feared, it fails on relocation:

```
+ cd /__w/pantry/pantry/testbeds/gnu.org__glibc-2.43.0
+ export LIBDIR=/github/home/.pkgx/gnu.org/glibc/v2.43.0/lib/glibc-2.43
+ LIBDIR=/github/home/.pkgx/gnu.org/glibc/v2.43.0/lib/glibc-2.43
+ export LDSO=ld-linux-x86-64.so.2
+ LDSO=ld-linux-x86-64.so.2
+ test -f /github/home/.pkgx/gnu.org/glibc/v2.43.0/lib/glibc-2.43/libc.so.6
+ test -f /github/home/.pkgx/gnu.org/glibc/v2.43.0/lib/glibc-2.43/ld-linux-x86-64.so.2
+ test -f /github/home/.pkgx/gnu.org/glibc/v2.43.0/include/stdio.h
+ test -L /github/home/.pkgx/gnu.org/glibc/v2.43.0/bin/ld.so
+ echo '--- ld.so --version ---'
--- ld.so --version ---
+ /github/home/.pkgx/gnu.org/glibc/v2.43.0/lib/glibc-2.43/ld-linux-x86-64.so.2 --version
+ head -2
ld.so (GNU libc) stable release version 2.43.
Copyright (C) 2026 Free Software Foundation, Inc.
+ gcc -o test-bottle test.c -nostdinc -isystem /github/home/.pkgx/gnu.org/glibc/v2.43.0/include -isystem /github/home/.pkgx/gnu.org/gcc/v16.1.0/lib/gcc/x86_64-pc-linux-gnu/16.1.0/include -B /github/home/.pkgx/gnu.org/glibc/v2.43.0/lib/glibc-2.43 -L /github/home/.pkgx/gnu.org/glibc/v2.43.0/lib/glibc-2.43 -Wl,--rpath=/github/home/.pkgx/gnu.org/glibc/v2.43.0/lib/glibc-2.43 -Wl,--dynamic-linker=/github/home/.pkgx/gnu.org/glibc/v2.43.0/lib/glibc-2.43/ld-linux-x86-64.so.2
/github/home/.pkgx/gnu.org/binutils/v2.46.0/bin/ld: cannot find /opt/gnu.org/glibc/v2.43.0/lib/glibc-2.43/libc.so.6: No such file or directory
/github/home/.pkgx/gnu.org/binutils/v2.46.0/bin/ld: cannot find /opt/gnu.org/glibc/v2.43.0/lib/glibc-2.43/libc_nonshared.a: No such file or directory
/github/home/.pkgx/gnu.org/binutils/v2.46.0/bin/ld: cannot find /opt/gnu.org/glibc/v2.43.0/lib/glibc-2.43/ld-linux-x86-64.so.2: No such file or directory
collect2: error: ld returned 1 exit status
```

https://github.com/pkgxdev/pantry/actions/runs/26298446139

jhheider:
https://github.com/pkgxdev/pantry/commit/5354c73f9a6201ad95cb3a30e6d892e1d6ee7346 seems to fix it. i'll trigger the others, and we can figure out what tweaks are needed to version the script.

tannevaled:
That's beautiful — the wrapper-script approach is so much cleaner than patchelf'ing PT_INTERP. Three big wins I see:

1. **Linker scripts use basenames** — `-L $LIBDIR -lc` resolves via the search path, so the bottle survives relocation from `/opt` to `~/.pkgx` to anywhere.
2. **`bin/*` are POSIX sh wrappers**, real binaries in `libexec/glibc-bin/` — the wrapper computes its own prefix via `cd ../$bindir`, no absolute paths baked in.
3. **`ld.so --library-path` explicit invocation** — sidesteps PT_INTERP entirely (which we knew was unrewritable by brewkit's fix-elf anyway).

This is essentially **bklibcvenv (pkgxdev/brewkit#344)** implemented inline. The shape factors out cleanly:

| | inline-in-recipe (this commit) | brewkit helper |
|---|---|---|
| wrapper template | `prop:` block in package.yml | `share/brewkit/glibc-wrapper.sh` |
| seal logic | the `for dir in bin sbin` loop | `libexec/bklibcvenv seal {{prefix}}` |
| recipe call | ~20 lines | `bklibcvenv seal {{prefix}}` (~1 line) |

So if a few more recipes start needing the same pattern (other libc bottles, …), the extraction to `libexec/bklibcvenv` becomes obvious. And the **bkwinvenv** sketch in pkgxdev/brewkit#347 is now empirically validated — same exact shape works for Windows DLLs, just swap sh wrappers for .cmd wrappers.

Re. "trigger the others and we can figure out what tweaks are needed to version the script" — pretty sure the wrapper template needs no version-specific tweaks, just per-arch (`@LDSO@`) which the env block already handles. Older glibc versions install to the same `lib/glibc-X.Y/` subdir so the wrapper's relative-path math holds. Let me know if I should help test any of the older versions.

jhheider:
ok. as of https://github.com/pkgxdev/pantry/actions/runs/26409068091 we can build glibc2.17 for x86-64. that involved a bunch of turns and local docker builds, but it passes CD.

tannevaled:
Awesome — that's the HPC baseline unlocked. Run 26409068091 showing as failure in the API but you mentioned "passes CD" — assuming you mean a subsequent run that resolved whatever was caught.

Anything you'd like me to take on for the cascade? Possibilities I can see:
- review the older glibc test step (the static-link approach we converged on may need tweaking for 2.17's gcc-7.5 / binutils-2.28 cascade)
- port the wrapper-script approach to the older versions if the inline-in-recipe pattern needs duplicating per version
- something else from your turn-by-turn debugging that's worth automating

Or if you'd rather just keep iterating yourself, totally fine — just flag if there's anything blocking that I can shave off.

jhheider:
2.17 passes for x86-64. I *suspect* it's too old for aarch64, but if there are patches that can get that building, I'm happy to have them.

Otherwise, if you need older binutils or gcc, let's look into those.

tannevaled:
You're right re. 2.17 + aarch64 — it was experimental ARM support in glibc 2.17 (Dec 2012); not viable as a real ABI baseline. The manylinux project explicitly diverges arch baselines:

| arch | manylinux baseline | glibc | typical RHEL/CentOS target |
|---|---|---|---|
| x86-64 | manylinux2014 | **2.17** | RHEL 7 / CentOS 7 |
| aarch64 | manylinux_2_28 | **2.28** | RHEL 8 |
| (alt) aarch64 | manylinux_2_24 | 2.24 | older arm distros |

So pragmatic per-arch baselines for the pkgx HPC cascade:

| | x86-64 | aarch64 |
|---|---|---|
| **conservative** | 2.17 ✅ (yours) | **2.28** |
| **slightly newer** | 2.24 | 2.24 (both arches, single baseline) |

My vote: **2.17 x86-64 + 2.28 aarch64** matches what every other major distro / packaging ecosystem does. Same RHEL 7/8 split as the rest of the HPC world. Anyone needing aarch64 < 2.28 has bespoke needs and can build their own.

For older binutils/gcc — likely yes for 2.28 too, since gcc 16 + glibc 2.28 might want `-fcommon -Wno-error` (same workaround we used for 2.17/2.24). Let me know which combos you've already triggered in CI vs. need exploring; I can pick up older gcc/binutils version walks in parallel if useful.

jhheider:
Glibc/x86-64 we have, I think, all the versions you wanted: https://dist.pkgx.dev/?prefix=gnu.org/glibc/linux/x86-64

Aarch64: we have 2.27 and 2.28 and onwanrds.

So, do we need older gcc/binutils? I think our gcc goes back to 10.

tannevaled:
Confirmed from `dist.tea.xyz` listing:

```
x86-64 :  2.17  2.24  2.27  2.28  2.34  2.38  2.41  2.42  2.43   (9 versions ✓)
aarch64:              2.27  2.28  2.34  2.38  2.41  2.42  2.43   (7 versions ✓)
```

That's the **complete HPC cascade matrix** we'd want — pretty much exactly the manylinux baseline coverage. 🎉

Re. older gcc/binutils — **gcc 10 is fine for the vast majority of consumers.** The use cases that need <10 are:

| use case | typical gcc |
|---|---|
| Building manylinux2014 pip wheels (true RHEL-7-built) | gcc 4.8 (devtoolset-2) or gcc 7.5 (devtoolset-7) |
| CUDA N.x baseline (older CUDA tied to specific gcc) | varies — CUDA 11.x → gcc ≤ 11, CUDA 10.x → gcc ≤ 8 |
| Fortran compat with old academic codes | gfortran 7.x / 8.x |
| AOCC / Intel pinning to specific gcc-as-host | usually gcc 9 / 10 |

Nobody NEEDS gcc < 10 for the common case ("build a binary that runs on RHEL 7"). The HPC users who do need it are usually building their own toolchain anyway (see `tannevaled/bootstrap-tools` repo for a manylinux2014-based gcc 7.5 + binutils 2.32 sketch, if anyone ever asks).

**So: no urgent need for older gcc/binutils.** Glibc cascade is the load-bearing piece; that's done. Can revisit if a downstream HPC consumer files an issue.

Side note: the Windows-port stack still has 4 PRs awaiting review when you have a moment (pkgxdev/pantry#12984 / #12986 / #12987 + pkgxdev/brewkit#343 / #347 / #348 — with #348 already approved, just gated on `enforce-test-delta`). No rush, just flagging that they exist and that #348 specifically could be a one-click merge if the test-delta gate is overridable.

tannevaled:
Walking back my previous "no urgent need" — that was from the *project-driving* angle (the glibc cascade + Windows port I've been on are unblocked at gcc 10+). But from the *HPC-user* angle, there are real use cases that benefit from older gcc/binutils in pantry, and you offered to look into them, so let me take you up on that:

**Useful older toolchain versions, in priority order:**

| ver | rationale |
|---|---|
| **gcc 7.5 + binutils 2.30** | manylinux2014 baseline (devtoolset-7); lets HPC users build *true* RHEL-7-compat binaries with the canonical toolchain rather than a newer-gcc-targeting-2.17 emulation |
| **gcc 9.5 + binutils 2.32** | CUDA 10.x / 11.x baseline (CUDA pins host compiler version); useful for academic Fortran/MPI codes still on devtoolset-9 |
| gcc 11.x | edge case; mostly covered by gcc 10 |

I have a working source-side cascade for **gcc 7.5 + gcc 9.5 + binutils 2.28/2.30/2.32** in `tannevaled/bootstrap-tools` — that repo builds these from source inside a manylinux2014 container, producing relocatable tarballs. Used as scaffolding for the original glibc 2.17 work; now obsolete for THAT purpose but the toolchain bottles themselves are still useful.

If you want them in pantry, two paths:

1. **I port the recipes** — straightforward extraction from bootstrap-tools/build.sh to `gnu.org/gcc/v7/package.yml` + `gnu.org/gcc/v9/package.yml` + `gnu.org/binutils/v2/package.yml` (with version pins). Probably 3 PRs.
2. **You drive it** — you've got more context on pantry's current gcc cascade structure (10 → 11 → … → 16). I'd just be guessing at the right per-version directory layout. If you give a pointer to the existing structure I can match it.

WDYT? I lean (1) but happy with either.

(And separately — answer to your literal question: no, the work I've been driving doesn't NEED older gcc. The above is for downstream HPC users.)

jhheider:
happy to let you lean on the existing recipes (or create variants, if needed, as above). if they'll drop right on the existing recipes, then it should just be a triggered build.

tannevaled:
Plot twist — checked `dist.tea.xyz` and the cascade is already shipping wider than your "gcc goes back to 10" estimate:

**gcc** (both x86-64 + aarch64):
```
7.5.0  9.5.0  10.5.0  11.4.0  11.5.0  12.2…12.5  13.1…13.4  14.1…14.3  15.1  15.2  16.1
```

**binutils** (both arches):
```
2.28.0  2.39.0  2.40.0  2.41.0  2.42.0  2.43.0/1  2.44.0  2.45.0/1  2.46.0
```

So **gcc 7.5 + gcc 9.5 + binutils 2.28 are already bottled and shipping.** That covers:

- gcc 7.5 → manylinux2014 / devtoolset-7 baseline ✅
- gcc 9.5 → CUDA 10.x / 11.x baseline ✅
- binutils 2.28 → matches CentOS 7 / manylinux2014 era ✅

Existing recipe comments already document this is the empirically-validated cascade for glibc 2.17 – 2.26 — so somebody (you?) already did the trigger work, possibly during the glibc 2.17 push earlier this week.

**Nothing for me to port.** I'll close out the "older toolchain in pantry" thread — already done.

Gaps I notice but don't think anyone urgently needs:

- gcc 8.x (between 7.5 and 9.5) — devtoolset-8, mostly redundant given 7.5+9.5 cover the ends
- binutils 2.29 – 2.38 (sizable gap) — practical effect: any package pinning specifically to 2.31/2.32 is stuck; rare in the wild

If a downstream user actually needs one of those, they can file an issue. Not blocking anything I'm aware of.

Thanks for the offer — turns out it was already taken care of. Moving the conversation to the still-open pieces in my next ping.

Issue #346, opened 2023-02-10
+gnu-sed (332/548)


mfts:
Formulae: https://formulae.brew.sh/formula/gnu-sed

PR #12984, opened 2026-05-22
new(llvm.org/mingw-w64): LLVM-based mingw-w64 cross-compiler (Windows-target toolchain)

tannevaled:
## What

Adds a new pantry recipe `llvm.org/mingw-w64` — a vendored re-distribution of upstream [mstorsjo/llvm-mingw](https://github.com/mstorsjo/llvm-mingw) prebuilt toolchains.

This is a **cross-compiler**: runs on Linux/macOS hosts, produces native Windows binaries (PE/COFF) for x86-64, aarch64, i686, armv7.

## Why

Phase-0 deliverable for **pkgxdev/brewkit#346** (RFC: native Windows bottles, cross-compile from Linux CI runners). Without a Windows-targeting toolchain in the pantry, no recipe can declare `platforms: windows/*`.

Once this lands, the next step is a pilot recipe (`jq` cross-compiled to `windows/x86-64`) — that needs:
- This toolchain (✅ here)
- A handful of brewkit-side platform-handling fixes (sketched in #346)

I'm marking this **draft** because it depends on the architecture sketch in #346 being approved. Happy to promote to ready-for-review when the RFC settles.

## Shape

Same pattern as `ziglang.org` and other vendored binary redistributions:
- `warnings: vendored` — declares we're re-shipping upstream binaries
- `distributable.url` → upstream source archive (small, mostly unused)
- `build.script` → curls the host-specific prebuilt tarball + extracts into `{{prefix}}`

## Host platforms

| host | upstream prebuilt |
|---|---|
| linux/x86-64 | `llvm-mingw-VER-ucrt-ubuntu-22.04-x86_64.tar.xz` (78 MB) |
| linux/aarch64 | `llvm-mingw-VER-ucrt-ubuntu-22.04-aarch64.tar.xz` (73 MB) |
| darwin/x86-64 | `llvm-mingw-VER-ucrt-macos-universal.tar.xz` (116 MB) |
| darwin/aarch64 | `llvm-mingw-VER-ucrt-macos-universal.tar.xz` (116 MB) |

UCRT variant is the default (Win10+); we can add MSVCRT variant later if we need legacy targets.

## Self-test

Cross-compiles a trivial `hello.c` to both x86-64 and aarch64 Windows targets, verifies the output is a valid PE/COFF binary (DOS `MZ` magic at offset 0). Doesn't try to *run* the .exe — that's a Windows binary on a Linux test sandbox.

## Open questions

(Same as #346, surfacing here):

- MinGW vs MSVC ABI default — keeping flexible via clang-cl since llvm-mingw can do both
- Whether to bundle `vcruntime140.dll` or rely on Win10+ shipping UCRT
- Naming: `llvm.org/mingw-w64` vs `github.com/mstorsjo/llvm-mingw`? I picked the former for findability; happy to rename.

Refs: pkgxdev/brewkit#346, pkgxdev/pkgx#607, pkgxdev/libpkgx#48


tannevaled:
**Big architectural finding — this PR is actually blocked on pkgxdev/pantry#12968 merging.**

Round-2 CI failure surfaced the underlying issue:

```
clang: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found
   (required by /opt/.../mingw-w64/v.../lib/libLLVM.so.22.1)
```

Upstream llvm-mingw binary tarballs are built against ubuntu-22.04 (glibc 2.35). The brewkit test sandbox is `debian:buster-slim` (glibc 2.28). Clang can't load — it needs symbols that don't exist in the host glibc.

This is **exactly the host-independence problem we just solved in pkgxdev/pantry#12968** (currently APPROVED, awaiting merge). Once #12968 lands and `gnu.org/glibc` is in pantry main, the fix here is simply:

```yaml
test:
  dependencies:
    winehq.org: '*'
    gnu.org/glibc: '>=2.34'   # llvm-mingw clang needs glibc 2.34+
```

brewkit will pull our bottled glibc 2.43 into the test environment; clang resolves against it via LD_LIBRARY_PATH; debian:buster's ancient glibc 2.28 stops being the constraint.

This is a beautiful loop closure: **the Windows port story directly composes with the glibc host-independence story.** Once #12968 lands, this PR becomes trivially fixable.

cc @jhheider — flagging because this elevates #12968's importance: it's not just "nice older-glibc support", it's "infrastructure for landing Windows bottles". Was already approved by you (thanks!) but the merge is what unblocks #12984/#12986/#12987/#347/RFC#346.

Marking this PR as **blocked-on #12968** until then.

tannevaled:
**Round 6 (0a8e7546): scope restricted to linux hosts for pilot.**

Status after rounds 4+5:

| host | result |
|---|---|
| linux/x86-64 | ✅ PASS |
| linux/aarch64 | ✅ PASS |
| darwin/x86-64 | ❌ FAILURE (self-hosted runner, logs not API-accessible) |
| darwin/aarch64 | ❌ FAILURE (39s, very early fail) |

Without log access I can't diagnose the darwin failures from outside — could be a network policy on the macOS runner (large GitHub Releases download blocked?), a `pantry` darwin-bottle gap for `gnu.org/tar` / `tukaani.org/xz`, or some shell incompatibility I'm missing.

Pragmatic move: scope this PR to `linux/x86-64` + `linux/aarch64` for now. That matches the RFC pkgxdev/brewkit#346 Phase-0 deliverable explicitly — "cross-compile native Windows bottles from **Linux CI runners**". darwin host support is a follow-up that needs maintainer access to the macOS runner logs.

@jhheider if you have a moment to peek at the actions runs:

- darwin/x86-64: https://github.com/pkgxdev/pantry/actions/runs/26291471707/job/77392594935
- darwin/aarch64: https://github.com/pkgxdev/pantry/actions/runs/26291471707/job/77392594874

…and let me know what's actually failing on the macOS sandbox, I can fix it in a follow-up. For now this PR should green on all declared platforms.

tannevaled:
Marking ready for review now that round-6 lands all declared platforms green:

| host | result |
|---|---|
| linux/x86-64 | ✅ |
| linux/aarch64 | ✅ |

darwin/* deferred to a follow-up (logs not externally accessible on the macOS self-hosted runner; will revisit with maintainer help).

This PR is the toolchain bottle for the pkgxdev/brewkit#346 Windows-port architecture. Once it lands:
- pkgxdev/pantry#12986 (wine) becomes unblocked (build-deps on this for the mingw cross-compiler)
- pkgxdev/pantry#12987 (jq Windows pilot) becomes unblocked (build-deps on this)
- pkgxdev/brewkit#347 (fix-pe + bkwinvenv skeletons + audit hook) gets practical exercise

Self-contained — no new brewkit changes needed for this PR to land.

jhheider:
darwin issues from https://github.com/pkgxdev/pantry/actions/runs/26291471707/job/77392594935:
```
+ cd /Users/builder/actions-runner/_work/pantry/pantry/testbeds/llvm.org__mingw-w64-20260519.0.0
+ x86_64-w64-mingw32-clang --version
+ cat
+ x86_64-w64-mingw32-clang -o hello-x86_64.exe hello.c
+ echo '::warning::`x86_64-w64-mingw32-clang` is not an explicit dependency!'
Warning: `x86_64-w64-mingw32-clang` is not an explicit dependency!
+ case $1 in
+ /opt/bin/pkgx x86_64-w64-mingw32-clang -o hello-x86_64.exe hello.c
hello.c:1:10: fatal error: 'stdio.h' file not found
    1 | #include <stdio.h>
      |          ^~~~~~~~~
1 error generated.
```

tannevaled:
Thanks for both the refactor (63dd893f) and the darwin log share — that was the key piece I was missing.

Pushed **831c3003** on top of your refactor:

1. **Full paths to own binaries in the test step** via env vars (`CLANG_X86_64` = `{{prefix}}/bin/x86_64-w64-mingw32-clang`). Bypasses brewkit's auto-fetch wrapper that picked up a different clang on darwin.
2. **Re-enabled darwin** in `platforms:` + added darwin entries to the per-arch PLATFORM env (both `darwin/x86-64` and `darwin/aarch64` map to `macos-universal`, since upstream ships a universal binary).
3. Minor: dropped the stray `EOF` line in the fixture content (leftover from when it was a heredoc before your refactor).

The auto-fetch trap is a subtle one — definitely makes sense for usability ("just write the command, brewkit figures it out"), but bites when a recipe's own bin/ is what should win. Full-path env vars feel like the right pattern for self-referential test steps. Let me know if there's a more idiomatic way (something like `bk-bin x86_64-w64-mingw32-clang`?) I should be using instead.

tannevaled:
Round status:
- linux/x86-64 ✅
- linux/aarch64 ✅
- darwin/x86-64 ❌
- darwin/aarch64 ❌

@jhheider — pushed **9112ee25** with inline diagnostics (env vars + `ls -la` on the clang binaries + un-suppressed `clang --version` stderr). The whole point was to surface what's actually failing on the macOS self-hosted runners, since their logs aren't accessible via the public API.

Could you peek at the latest darwin runs and paste the diagnostic output here? Specifically:

- darwin/x86-64: https://github.com/pkgxdev/pantry/actions/runs/26302145480/job/77432005769 (or whichever the latest is)
- darwin/aarch64: same run

The lines I'm after start with `── env diagnostics ──`. With those + the `clang --version` real error message we should be able to nail this in one more iteration.

If the `ls -la "$CLANG_X86_64"` says "no such file or directory" then the install paths differ on darwin (maybe due to the macos-universal tarball's internal layout). If it says the file exists but `clang --version` still fails differently than on linux, then it's a runtime issue (maybe macOS Gatekeeper / quarantine on a curl'd binary).

jhheider:
```
+ cd /Users/builder/actions-runner/_work/pantry/pantry/testbeds/llvm.org__mingw-w64-20260519.0.0
+ export WINEDEBUG=-all
+ WINEDEBUG=-all
+ export 'WINEDLLOVERRIDES=mscoree=;mshtml='
+ WINEDLLOVERRIDES='mscoree=;mshtml='
+ export WINEPREFIX=/Users/builder/actions-runner/_work/pantry/pantry/testbeds/llvm.org__mingw-w64-20260519.0.0/.wine
+ WINEPREFIX=/Users/builder/actions-runner/_work/pantry/pantry/testbeds/llvm.org__mingw-w64-20260519.0.0/.wine
+ export CLANG_X86_64=/opt/llvm.org/mingw-w64/v20260519.0.0/bin/x86_64-w64-mingw32-clang
+ CLANG_X86_64=/opt/llvm.org/mingw-w64/v20260519.0.0/bin/x86_64-w64-mingw32-clang
+ export CLANG_AARCH64=/opt/llvm.org/mingw-w64/v20260519.0.0/bin/aarch64-w64-mingw32-clang
+ CLANG_AARCH64=/opt/llvm.org/mingw-w64/v20260519.0.0/bin/aarch64-w64-mingw32-clang
+ echo '── env diagnostics ──'
+ echo CLANG_X86_64=/opt/llvm.org/mingw-w64/v20260519.0.0/bin/x86_64-w64-mingw32-clang
+ echo CLANG_AARCH64=/opt/llvm.org/mingw-w64/v20260519.0.0/bin/aarch64-w64-mingw32-clang
+ echo '── filesystem check ──'
+ ls -la /opt/llvm.org/mingw-w64/v20260519.0.0/bin/x86_64-w64-mingw32-clang
── env diagnostics ──
CLANG_X86_64=/opt/llvm.org/mingw-w64/v20260519.0.0/bin/x86_64-w64-mingw32-clang
CLANG_AARCH64=/opt/llvm.org/mingw-w64/v20260519.0.0/bin/aarch64-w64-mingw32-clang
── filesystem check ──
lrwxr-xr-x  1 builder  staff  23 May 19 09:28 /opt/llvm.org/mingw-w64/v20260519.0.0/bin/x86_64-w64-mingw32-clang -> clang-target-wrapper.sh
+ ls -la /opt/llvm.org/mingw-w64/v20260519.0.0/bin/aarch64-w64-mingw32-clang
lrwxr-xr-x  1 builder  staff  23 May 19 09:28 /opt/llvm.org/mingw-w64/v20260519.0.0/bin/aarch64-w64-mingw32-clang -> clang-target-wrapper.sh
+ /opt/llvm.org/mingw-w64/v20260519.0.0/bin/x86_64-w64-mingw32-clang --version
clang version 22.1.6 (https://github.com/llvm/llvm-project.git fc4aad7b5db3fff421df9a9637605b9ca5667881)
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: /opt/llvm.org/mingw-w64/v20260519.0.0/bin
Configuration file: /opt/llvm.org/mingw-w64/v20260519.0.0/bin/x86_64-w64-windows-gnu.cfg
+ OLD_FIXTURE=
++ mktemp
+ FIXTURE=/var/folders/y3/yr3bxsy95c3gfwwj61ytbp000000gq/T/tmp.zhQvDTpdx0
+ cat
+ cp /var/folders/y3/yr3bxsy95c3gfwwj61ytbp000000gq/T/tmp.zhQvDTpdx0 hello.c
+ rm -f /var/folders/y3/yr3bxsy95c3gfwwj61ytbp000000gq/T/tmp.zhQvDTpdx0
+ test -n ''
+ unset FIXTURE
+ /opt/llvm.org/mingw-w64/v20260519.0.0/bin/x86_64-w64-mingw32-clang -o hello-x86_64.exe hello.c
hello.c:1:10: fatal error: 'stdio.h' file not found
    1 | #include <stdio.h>
      |          ^~~~~~~~~
1 error generated.
```

https://github.com/pkgxdev/pantry/actions/runs/26300282827/job/77423413833 doesn't expand for you? I feel like I can still see them when logged out.

tannevaled:
Got it — nailed the darwin failure 🎯 Thanks for the diagnostic paste, the `ls -la` showing the symlink + working `--version` but `'stdio.h' file not found` was the giveaway.

**Root cause:** the install step cherry-picked `bin lib include share`, but llvm-mingw keeps its headers and import libs in *per-target sysroots* that sit next to those — `generic-w64-mingw32/include/` is where `stdio.h` lives, and each `<triple>-w64-mingw32/lib` holds the import libs. Those dirs were never copied into `{{prefix}}`, so the (working) clang couldn't find its own headers.

**Why it looked green on Linux:** the brewkit sandbox is debian:buster (glibc 2.28), so llvm-mingw's clang can't even load there — the test hit the `clang --version` guard and `exit 0`'d. The compile never ran on Linux, so the missing-sysroot bug only ever surfaced on darwin, which runs clang for real. (Nice side effect: once #12968 lands and Linux can actually load clang, this same install fix is what'll let the Linux compile succeed too, instead of skipping.)

**Fix** (pushed `f5061d4`): extract into a staging dir and copy the whole tree, no cherry-picking. Verified locally on darwin/arm64 — both `x86_64-w64-mingw32-clang` and `aarch64-w64-mingw32-clang` compile a `<stdio.h>` program to valid PE/COFF (MZ) binaries.

PR #12986, opened 2026-05-22
new(winehq.org): headless wine for Windows-binary CI testing

### LINKED ISSUE #12985: new(winehq.org): wine in pantry — needed for Windows-binary CI testing
## Why

For native Windows bottle support (pkgxdev/brewkit#346 + pkgxdev/pantry#12984), we need a way to **actually run** the cross-compiled `.exe` files in brewkit's Linux test sandbox. Otherwise the test step is reduced to a PE-magic-byte check (necessary but not sufficient).

[Wine](https://www.winehq.org) is the standard tool for that — it implements the Windows API on Linux/macOS hosts and can transparently run native Windows binaries. It's exactly what we need for headless CI.

## What we'd need

A `winehq.org/wine` recipe in pantry. Roughly:

```yaml
distributable:
  url: https://dl.winehq.org/wine/source/X.Y/wine-{{version.raw}}.tar.xz
  strip-components: 1

versions:
  url: https://dl.winehq.org/wine/source/
  match: …

platforms:
  - linux/x86-64
  - linux/aarch64
  - darwin/x86-64    # via Rosetta on Apple Silicon
  - darwin/aarch64   # via Wine on ARM

build:
  dependencies:
    freetype.org: '*'
    gnu.org/gnutls: '*'
    # … wine's transitive deps
  script:
    - ./configure --enable-win64 --prefix={{prefix}} --disable-tests
    - make -j {{hw.concurrency}}
    - make install

provides:
  - bin/wine
  - bin/wine64
```

(Wine's build is non-trivial — about 30-60 min single-arch, lots of optional deps. We could probably get away with `--without-x` and friends for headless CI use.)

## What it unblocks

Once `winehq.org/wine` exists:

1. **pkgxdev/pantry#12984** — `llvm.org/mingw-w64` test step upgrades from "verify PE magic" to "actually run hello.exe and assert stdout" (already wired as soft-dep; activates automatically when wine is available).

2. **Future Windows recipes** can include real runtime tests in their `test:` block, e.g.:
   ```yaml
   test:
     dependencies:
       winehq.org: '*'
     script:
       - wine64 {{prefix}}/bin/foo.exe --version
   ```

3. **Brewkit's audit step** can sanity-check that bundled `.exe` files actually load (catches things like undefined imports, ABI mismatches, missing DLLs in the bundle — which a static magic-byte check would miss).

## Alternatives considered

- **GitHub Actions Windows runners**: ~10× CPU minutes, slower, requires a separate workflow. Realistic for a final smoke-test but expensive for every PR.
- **wine via Docker base image**: would work but couples brewkit's CI infrastructure to a specific Docker image. Less reusable.
- **box64 / proton / crossover**: wrong tool, commercial, or not designed for this use case.

Wine in pantry is the cleanest fit: same model as every other build dep, opt-in per recipe.

## Open questions

1. **Build from source vs. redistribute upstream binaries** — WineHQ publishes per-distro packages (`.deb` for Ubuntu, `.rpm` for Fedora) but no generic Linux tarball. Source build is the only arch-agnostic path. Acceptable?
2. **wine-stable vs. wine-staging** — stable is fine for our use case (CLI testing); staging has more recent fixes. Either works.
3. **Optional deps** — wine builds with optional support for OpenGL, Vulkan, ALSA, Pulse, X11, SDL, gphoto2, … For headless CI we can disable most. Which is the minimal-viable set?

cc @jhheider @mxcl — context: pkgxdev/brewkit#346 RFC + pkgxdev/pantry#12984 toolchain PR.

tannevaled:
Closes pkgxdev/pantry#12985.

## What

Adds a `winehq.org` recipe — a minimal headless build of [Wine](https://www.winehq.org/) sufficient to run console-only Windows binaries on a Linux host. Pilot scope: **linux/x86-64 only**.

## Why

Unblocks pkgxdev/pantry#12984 (`llvm.org/mingw-w64`) — that recipe's test step is already wired with a soft-dep on `winehq.org`, so once this lands the cross-compile pilot upgrades from "PE magic-byte check" to "actually run the .exe and assert stdout". More broadly: any future Windows-targeting pantry recipe gets real runtime tests in the existing Linux CI infra (no GitHub Windows runners needed).

Architecture context: pkgxdev/brewkit#346 RFC.

## Build configuration

Stripped down to the minimum needed for CLI tool execution:

| Flag | Why |
|---|---|
| `--enable-archs=x86_64` | 64-bit only — halves build, sufficient for our use |
| `--disable-tests` | skip wine's massive internal test suite |
| `--without-x` | no X11 in CI |
| `--without-opengl --without-osmesa --without-vulkan` | no 3D |
| `--without-alsa --without-pulse --without-oss` | no audio |
| `--without-sdl --without-gstreamer` | no multimedia |
| `--without-cups --without-gphoto --without-usb --without-v4l2 --without-sane --without-udev --without-pcap` | no peripherals |
| `--without-fontconfig --without-krb5 --without-capi --without-cms` | other unneeded subsystems |

Build deps from pantry: `gnu.org/bison`, `github.com/westes/flex`, `perl.org`, `gnu.org/gettext`, `freetype.org`, `gnutls.org`, `zlib.net`, plus the toolchain (`gnu.org/{make,gcc,binutils}`).

## Test

```sh
wine64 --version   # expects "wine-X.Y"
```

That's all. Anything more substantive belongs in consumer recipes (e.g., `llvm-mingw` runs hello.exe through wine).

## Limitations / future work

- **linux/aarch64** — wine can run aarch64 Windows binaries on aarch64 Linux, but x86-64 Windows binaries need qemu-user or [Hangover](https://github.com/AndreRH/hangover). Out of scope for pilot.
- **darwin** — wine 11.x ships macOS support but configuration is meaningfully different (32-bit deprecation dance). Follow-up.
- **wine-staging** — we pick the stable branch; staging is generally OK too but adds churn.
- **MinGW-built Wine** — `--with-mingw` produces wine with better Windows compat for the win-side DLLs, but requires a Linux-hosted mingw toolchain present at wine build time (which we now have via pkgxdev/pantry#12984). Possible follow-up to unlock harder-to-emulate APIs.

## Test plan

- [ ] CI green on linux/x86-64
- [ ] `wine64 --version` returns `wine-11.0` (or whatever stable is at build time)
- [ ] As consumer-side validation: pkgxdev/pantry#12984's test step runs `hello.exe` through this wine when both PRs are merged

Refs: pkgxdev/pantry#12985 (this issue), pkgxdev/brewkit#346 (Windows port RFC), pkgxdev/pkgx#607 (platform megaticket).

### DIFF
--- a/projects/winehq.org/package.yml
+++ b/projects/winehq.org/package.yml
@@ -71,6 +71,8 @@ build:
     - ./configure $ARGS || (cat config.log && false)
     - make --jobs {{ hw.concurrency }}
     - make install
+    - run: if test ! -f wine64; then ln -s wine wine64; fi
+      working-directory: ${{prefix}}/bin
 
   env:
     ARGS:
--- a/projects/winehq.org/package.yml
+++ b/projects/winehq.org/package.yml
@@ -61,6 +61,12 @@ build:
     gnu.org/make: '*'
     gnu.org/gcc: '*'
     gnu.org/binutils: '*'
+    # Modern Wine compiles its built-in DLLs as real PE binaries, so it
+    # needs a PE cross-compiler at build time — configure aborts with
+    # "x86_64 PE cross-compiler not found" without one. llvm.org/mingw-w64
+    # provides x86_64-w64-mingw32-{gcc,clang}. No build-dep cycle: the
+    # mingw recipe uses Wine only in its *test* step, never its build.
+    llvm.org/mingw-w64: '*'
 
   script:
     - ./configure $ARGS
@@ -72,10 +78,11 @@ build:
       - --prefix={{ prefix }}
       - --enable-archs=x86_64       # 64-bit only — half the build, all we need
       - --disable-tests             # skip wine's massive internal test suite
-      - --without-mingw             # use wine's built-in win-side DLLs instead of
-                                    # mingw-cross-compiled ones (avoids requiring
-                                    # llvm.org/mingw-w64 at build time, which would
-                                    # be a build-dep cycle once mingw recipe lands)
+      - --with-mingw                # build the win-side DLLs as PE with the
+                                    # llvm.org/mingw-w64 cross-compiler (now a
+                                    # build dep). Required: modern Wine has no
+                                    # ELF-builtin fallback, so configure demands
+                                    # a PE cross-compiler regardless.
       - --without-x                 # no X11
       - --without-opengl
       - --without-osmesa
--- a/projects/winehq.org/package.yml
+++ b/projects/winehq.org/package.yml
@@ -1,0 +1,106 @@
+# Wine: Windows API implementation for POSIX hosts.
+#
+# Headless minimal build aimed at CI use — runs cross-compiled
+# Windows .exe files in brewkit's Linux test sandbox so we can
+# validate native-Windows pantry recipes without spinning up
+# GitHub Windows runners. See pkgxdev/pantry#12985 for context
+# and pkgxdev/brewkit#346 for the broader RFC.
+#
+# What's deliberately out of this build:
+#
+#   - X11 (--without-x): no GUI in CI
+#   - OpenGL/Vulkan/OSMesa: no 3D rendering
+#   - ALSA/PulseAudio/OSS: no audio
+#   - SDL/GStreamer: no media
+#   - CUPS/gphoto2/USB/V4L2/SANE/UDev: no peripherals
+#   - 32-bit Win support (--enable-archs=x86_64): 64-bit only
+#   - tests (--disable-tests): skip wine's own test suite (massive
+#     build-time saving, irrelevant for our use case)
+#
+# Result: a headless wine64 binary that can launch a console-only
+# Windows .exe and capture its stdout/stderr. Good for CLI tool
+# testing; not useful for GUI apps, games, or anything graphical.
+
+distributable:
+  url: https://dl.winehq.org/wine/source/{{ version.major }}.0/wine-{{ version.raw }}.tar.xz
+  strip-components: 1
+
+# Use the github mirror's tag list, filtering down to numeric semver
+# tags (excludes legacy daily-build tags like wine-20050930).
+versions:
+  github: wine-mirror/wine/tags
+  match: /^wine-\d{2,}\.\d+(\.\d+)?$/
+  strip: /^wine-/
+
+# Pilot scope: linux/x86-64 only. Once the integration with
+# brewkit#346 lands, expanding to linux/aarch64 + darwin/* is
+# a follow-up — both have wine-side complications:
+#   - linux/aarch64: wine can run aarch64 Windows binaries but
+#     x86-64 Windows requires qemu-user or hangover
+#   - darwin: Apple deprecated 32-bit; wine 11.x ships darwin
+#     support but it's a meaningful extra config dance
+platforms:
+  - linux/x86-64
+
+build:
+  dependencies:
+    gnu.org/bison: '^3'
+    github.com/westes/flex: '*'
+    perl.org: '^5'
+    gnu.org/gettext: '*'
+    freetype.org: '*'
+    gnutls.org: '*'
+    zlib.net: '*'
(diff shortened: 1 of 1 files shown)


tannevaled:
**Round 3 finding: wine is also blocked on pkgxdev/pantry#12968.**

`--without-mingw` doesn't fully disable the cross-compiler requirement — `--enable-archs=x86_64` still triggers wine's configure to look for an x86_64 PE compiler:

```
configure: error: x86_64 PE cross-compiler not found.
This is an error since --enable-archs=x86_64 was requested.
```

That makes sense: wine MUST build its own Windows-side DLLs (kernel32, ntdll, …) for the requested target arch, because those DLLs interpose Windows API calls into Linux syscalls. There's no "ship without Windows-side DLLs" path.

So wine builds need a mingw cross-compiler. Adding `llvm.org/mingw-w64` as a build dep would normally fix this — but:

```
pkgxdev/pantry#12968  (glibc, approved/awaiting merge)
       │
       │ (unblocks)
       ▼
pkgxdev/pantry#12984  (llvm-mingw, currently can't run clang)
       │
       │ (build dep)
       ▼
THIS PR (wine)
```

So this PR's full unblock chain is **#12968 → #12984 → this**.

For now: leaving the recipe with the current `--enable-archs=x86_64` + `--without-mingw` combo. Once #12984 becomes buildable I'll add it as a build dep and the configure should pass.

(Side note: `--without-mingw` does prevent wine from REQUIRING mingw post-configure, but it doesn't override `--enable-archs`'s implicit PE-compiler requirement. Wine's configure has both knobs but `--enable-archs` is the stronger one.)

tannevaled:
Hey @jhheider — dug into the `x86_64 PE cross-compiler not found` failure pretty hard (reproduced wine's configure on a clean Debian + pkgx, both arm64 native and an x86_64 amd64 chroot via qemu-user with real glibc). Summary in case it saves you time:

**The recipe is correct.** With `--with-mingw` + a `llvm.org/mingw-w64` build dep, `./configure --enable-archs=x86_64 --with-mingw …` **passes** the PE-compiler probe outside brewkit:

| env | `…supports -target x86_64-windows … --no-default-config` |
|---|---|
| arm64 + pkgx (raw) | ✅ yes |
| **x86_64 + pkgx (raw)** | ✅ **yes** |
| x86_64 + brewkit CI | ❌ no → "PE cross-compiler not found" |

**Ruled out** as the cause of the CI `no` (each tested in the x86_64 chroot):
- host arch, and missing `libc6-dev` (CI has it for the host gcc)
- brewkit's **linker shim** (`lld/lld-link → /usr/bin/$tool`): llvm-mingw's clang invokes `lld-link` by **absolute** path, so the PATH shim is bypassed — fake-shimming it leaves the probe `yes`
- brewkit's **compiler shim** (`clang → ruby wrapper`): llvm-mingw's `clang-target-wrapper.sh` resolves clang via `$DIR=$(get_dir $0)`, also absolute → bypassed
- brewkit's injected `CFLAGS=-fPIC` / `CXXFLAGS=-fPIC` / `LDFLAGS=-pie` (exported them explicitly — probe stays `yes`)

So the `no` is something specific to the brewkit build sandbox on the x86_64 runner that I can't reproduce off the actual runner. **Could you grab the `config.log` from a failed wine run and paste the conftest block around the `x86_64-w64-mingw32-gcc works… no`?** The exact compile/link error there should pin it. The probe command is:

```
x86_64-w64-mingw32-gcc -o conftest.exe -g -O2 -target x86_64-windows -fuse-ld=lld --no-default-config -Wl,-subsystem:console -Wl,-WX -Werror=unknown-warning-option -Werror=unused-command-line-argument -Werror=ignored-optimization-argument -Werror=unsupported-floating-point-opt -nostdlib -nodefaultlibs conftest.c
```

(Recipe fix is pushed; the rest of the CI is green.) Thanks!

jhheider:
Best way to get the conf log is to cat it in the build script.

jhheider:
looks like glibc version needs; this is a good opportunity to test the libc wrapper tech.

PR #12987, opened 2026-05-22
feat(jq): add windows/x86-64 cross-compile pilot (#346 Phase 0)

tannevaled:
**Draft — Phase-0 pilot for pkgxdev/brewkit#346 (native Windows bottles RFC).**

## What

Extends the existing `jq` recipe to also build on `windows/x86-64` (and trivially `windows/aarch64`), via cross-compile from Linux/macOS runners using the `llvm.org/mingw-w64` toolchain in pkgxdev/pantry#12984.

## End-to-end loop demonstrated

```
linux/x86-64 runner
       │
       ▼
build.script
       │  detects windows/* → switches CC + --host
       ▼
configure (host=x86_64-w64-mingw32, CC=x86_64-w64-mingw32-clang, --disable-onigjq)
       │
       ▼
make → make install
       │
       ▼
{{prefix}}/bin/jq.exe           ◄── cross-compiled PE/COFF binary
       │
test.script
       │  detects windows/*
       ▼
wine64 jq.exe .devs[1].github < test.json
       │
       ▼
assert == "jhheider"
       │
       ▼
PASS
```

This closes the Windows-port architectural loop sketched at pkgxdev/brewkit#346 with a concrete, reviewable recipe that actually produces + tests a Windows binary using only Linux CI.

## Depends on

| | |
|---|---|
| pkgxdev/pantry#12984 | `llvm.org/mingw-w64` toolchain |
| pkgxdev/pantry#12986 | `winehq.org` headless wine |

Both are draft PRs. This PR currently has soft fallbacks (PE magic-byte check when wine is missing) so it can be CI-validated even before #12986 lands — but the full runtime-validation story needs all three.

## Known pilot limitations (called out in the recipe comments)

- **Disables onigjq** (regex) on Windows since oniguruma doesn't have a Windows bottle yet
- **`provides: bin/jq`** doesn't account for the `.exe` suffix on Windows — audit might complain; needs a brewkit hook
- **Unconditional `llvm.org/mingw-w64` build dep**: harmless on non-Windows but ideally conditional once brewkit conditional-dep semantics are settled

## What this is NOT

- Not a complete Windows port of jq (no regex on Windows)
- Not blocking the production linux/darwin jq build (current behaviour preserved)
- Not advocating for "Windows everywhere" — opt-in per recipe

## Test plan

- [ ] linux/x86-64 CI: existing native build/test still passes
- [ ] linux/aarch64 CI: existing native build/test still passes
- [ ] darwin/* CI: existing native build/test still passes
- [ ] windows/x86-64 CI (once brewkit platform support lands): cross-compiles + runs through wine
- [ ] Manual: `wine64 jq.exe .` smoke-tests the resulting binary on a non-CI Linux box

Refs: pkgxdev/brewkit#346 (RFC), pkgxdev/brewkit#347 (fix-pe + bkwinvenv), pkgxdev/pkgx#607.

### DIFF
--- a/projects/stedolan.github.io/jq/package.yml
+++ b/projects/stedolan.github.io/jq/package.yml
@@ -8,10 +8,15 @@ versions:
 
 dependencies:
   github.com/kkos/oniguruma: 6
+  # NOTE on windows/* the build is configured with --disable-onigjq,
+  # so this dep is unused at runtime there. Tracking the gap at
+  # pkgxdev/brewkit#346 — once oniguruma has a Windows bottle we
+  # can re-enable regex support on Windows.
 
 build:
   dependencies:
     git-scm.org: 2
+    llvm.org/mingw-w64: '*'  # cross-compiler used only when targeting windows/*
   script:
     - run: |
         if test "{{hw.platform}}" = "darwin"; then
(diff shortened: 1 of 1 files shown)


tannevaled:
**Note: CI fails by design until pkgxdev/pantry#12984 lands.**

The CI failure is:

```
error: Uncaught (in promise) Error: pkg not found: llvm.org/mingw-w64
```

…because this PR adds `llvm.org/mingw-w64` as a build dep, but that recipe is in pkgxdev/pantry#12984 which hasn't been merged yet. brewkit can't resolve the dep closure against pantry's main branch.

Two paths forward:
- **Wait** for #12984 to land, then this PR's CI auto-greens
- **Remove** the `llvm.org/mingw-w64` dep from this PR temporarily — but then the recipe can't actually cross-compile (just documents intent), which defeats the pilot's purpose

Sticking with path #1 — this PR stays draft + CI-red until the toolchain is available.

For non-Windows builds the recipe should still work (Windows path is gated by `if [ "{{hw.platform}}" = "windows" ]`). The dep-resolution-time failure happens for ALL platforms though, because brewkit checks deps before evaluating platform-conditionals. That's a brewkit-side limitation worth tracking — see open question on pkgxdev/brewkit#346:

> Should brewkit gain conditional `dependencies: { foo: '*', if: <predicate> }` syntax (mirroring what already works on script steps)?

If yes, this PR can ship with `llvm.org/mingw-w64: { '*', if: 'windows' }` and CI greens immediately for linux/darwin builds.

PR #13043, opened 2026-05-29
new(socket_vmnet): github.com/lima-vm/socket_vmnet recipe (vmnet.framework bridge)

tannevaled:
Closes a mainstream package coverage gap (see pkgxdev/pantry overall package coverage discussion). Draft until CI confirms green.

See commit message for design choices.

### DIFF
--- a/projects/github.com/lima-vm/socket_vmnet/package.yml
+++ b/projects/github.com/lima-vm/socket_vmnet/package.yml
@@ -10,7 +10,7 @@
 # Linux equivalent.
 
 distributable:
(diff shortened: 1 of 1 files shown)


jhheider:
since we have `lima-vm.io`, this should be `lima-vm.io/socket_vmnet`.

i'll got through these (slowly), but some notes:

```yaml
platforms: [darwin/x86-64, darwin/aarch64]

# is the same as

platforms: [darwin]
```

if you're going to use list-scripts (and you should), it's _rarely_ necessary, or good style, to just use a text script inside it:

```yaml

- run: |
    cd foo
    ls bar
    make baz

# is better as

- run:
  - ls bar
  - make baz
  working-directory: foo
```

PR #13047, opened 2026-05-29
new(mingw-w64.org): Windows runtime libs (headers + CRT + winpthreads) — from source

tannevaled:
Pure Windows runtime (headers + CRT + winpthreads), built from source for x86_64 + aarch64 targets. NOT a compiler — just the target-side libs.

Cross-compiled using pantry's existing llvm.org clang (`--target=*-w64-mingw32`).

Companion to pkgxdev/pantry#12984 which will be refactored to *compose* this runtime + pantry's llvm.org into the full Windows cross-toolchain (instead of vendoring upstream's prebuilt llvm-mingw).

### DIFF
--- a/projects/mingw-w64.org/package.yml
+++ b/projects/mingw-w64.org/package.yml
@@ -147,11 +147,15 @@ build:
 test:
   # Lib-only package — no binaries to invoke. Verify the expected
   # runtime files landed in their per-target subdirs.
+  #
+  # winpthreads ships its static lib under different names depending
+  # on the host libtool: linux installs `libpthread.a`, darwin's
+  # libtool produces `libwinpthread.lib` (PE-style). Accept either.
   script:
     - test -f "{{prefix}}/x86_64-w64-mingw32/include/windows.h"
     - test -f "{{prefix}}/x86_64-w64-mingw32/include/stdio.h"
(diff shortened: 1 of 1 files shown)


tannevaled:
Linux x86_64 + aarch64 both green on the latest commit — the from-source build works end-to-end (headers + CRT + winpthreads).

darwin x64 / darwin ARM64 fail but logs aren't API-accessible (self-hosted runners return 404). @jhheider could you paste the tail of the failing darwin job?

The Linux fix was 3-step:
1. headers FIRST, then export CC/CXX with `--sysroot={{prefix}}/$T` (so CRT finds windows.h)
2. export `RC=llvm-windres` (so libtool recognises rc-compile path)
3. pass `RCFLAGS=-I{{prefix}}/$T/include` to winpthreads configure (so windres finds winver.h)

Darwin uses the same `llvm.org` toolchain (multi-target clang + llvm-windres etc.) and the same `gnu.org/coreutils` for install(1), so I'd expect the same recipe to work. Likely candidates if it doesn't:
- BSD vs GNU tool difference somewhere
- darwin clang's `--sysroot` handling for Windows targets
- some configure probe that behaves differently under macOS host

Job URLs:
- darwin x64: https://github.com/pkgxdev/pantry/actions/runs/26627803090
- darwin ARM64: https://github.com/pkgxdev/pantry/actions/runs/26627803090

PR #13045, opened 2026-05-29
new(R): r-project.org recipe (statistical computing)

tannevaled:
Closes a mainstream package coverage gap (see pkgxdev/pantry overall package coverage discussion). Draft until CI confirms green.

See commit message for design choices.

PR #13042, opened 2026-05-29
new(argocd): github.com/argoproj/argo-cd recipe

tannevaled:
Closes a mainstream package coverage gap (see pkgxdev/pantry overall package coverage discussion). Draft until CI confirms green.

See commit message for design choices.

PR #13050, opened 2026-05-29
new(cubefs.io): CubeFS — cloud-native distributed storage (from source)

tannevaled:
## Summary
- New recipe **cubefs.io** — CubeFS (formerly ChubaoFS, CNCF graduated) — cloud-native distributed storage system with HDFS / S3 / POSIX semantics.
- Built **from source** via the project's Makefile + build.sh, which compiles vendored C libs (RocksDB 6.3.6, snappy, zstd, lz4, zlib, bzip2) statically into the Go binaries. No \`warnings: vendored\`.
- Slim install: \`cfs-server\` + \`cfs-client\` + \`cfs-cli\` + \`cfs-authtool\` + \`cfs-fsck\`. The blobstore object-storage suite is omitted to keep the bottle lean (re-add via \`make blobstore\` if needed).
- Linux x86-64 + aarch64. Darwin not supported (upstream's build.sh partially does, but libcfs.so + several CGO components skip on macOS — keep the platform set conservative).

## Test plan
- [ ] \`cfs-server -v\` reports the bottle's version
- [ ] \`cfs-cli\` loads (version output soft-pass)
- [ ] \`cfs-fsck --help\` / \`cfs-authtool --help\` load
- [ ] Bottle size: server + 4 supporting binaries (each ~50-100MB due to static RocksDB)

## Context
Part of the from-source recipes batch (#13042 argocd, #13043 socket_vmnet, #13044 buildah, #13045 R, #13046 envoy, #13047 mingw-w64.org, #13048 llvm.org/mingw-w64). Project goal: source-buildable independence ("être capable de compiler depuis les sources est un gage d'indépendance").

### DIFF
--- a/projects/cubefs.io/package.yml
+++ b/projects/cubefs.io/package.yml
@@ -81,6 +81,11 @@ build:
         # let the policy-min override apply.
         export CMAKE_POLICY_VERSION_MINIMUM=3.5
 
+        # build.sh derives the version with `git describe`, but the release
+        # tarball has no .git — leaving proto.Version empty so the binaries
+        # report a blank version. Inject the tag so `cfs-server -v` reports it.
+        sed -i 's|git describe --abbrev=0 --tags 2>/dev/null|echo v{{version.raw}}|' build/build.sh
+
         # gcc 8 defaults to gnu++14 already — no extra CXXFLAGS needed
         # for the FileMetaData implicit-copy-ctor path.
 
--- a/projects/cubefs.io/package.yml
+++ b/projects/cubefs.io/package.yml
@@ -1,0 +1,88 @@
+# CubeFS — cloud-native distributed storage system (CNCF graduated).
+#
+# Distributed FS with HDFS / S3 / POSIX semantics, designed for large-
+# scale cloud-native workloads. Built from source via the project's
+# Makefile + build.sh, which compiles a set of vendored C libs
+# (RocksDB, snappy, zstd, etc.) statically into the Go binaries.
+#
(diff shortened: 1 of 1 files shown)


tannevaled:
Re-marking as draft — RocksDB 6.3.6's FileMetaData has FileSampledStats with std::atomic, and modern libstdc++ (tested gcc 13 + gcc 11 + -std=gnu++14) can't synthesize the copy/move ctor that std::pair<int, FileMetaData>::emplace_back requires:

```
./db/version_edit.h: In instantiation of 'constexpr std::pair<_T1, _T2>::pair(_U1&&, _U2&&) [with _U1 = int&; _U2 = rocksdb::FileMetaData; ...]'
```

The proper fix is patching the vendored rocksdb-6.3.6 sources to add explicit move/copy ctors to FileMetaData. That requires careful source-level work that's better done as a follow-up. CubeFS upstream's own CI uses Debian 10 + gcc 8 (last gcc that compiled this rocksdb version cleanly) — we'd need to either:
1. Pin gcc <= 8 (not in pantry)
2. Apply a source patch to FileMetaData
3. Wait for CubeFS to bump their vendored rocksdb version

Draft until one of those happens.

tannevaled:
Reopened — instead of fighting RocksDB 6.3.6's pre-C++17 idioms against modern libstdc++, we add gcc 8 to pantry (companion PR #13070) and use that.

This is exactly the kind of version-pinning pkgx was designed to handle. From-source promise intact, no monkey-patching of vendored RocksDB sources.

Will go green once #13070 (gcc 8) lands and a bottle exists.

PR #13057, opened 2026-05-29
new(tcl-lang.org/tk): Tk GUI toolkit (Top 300 prereq)

tannevaled:
Companion to tcl-lang.org. Required to build git-gui, gitk, python-tk, etc. Part of #99.

### DIFF
--- a/projects/tcl-lang.org/tk/package.yml
+++ b/projects/tcl-lang.org/tk/package.yml
@@ -67,7 +67,21 @@ build:
         - --enable-aqua
 
 test:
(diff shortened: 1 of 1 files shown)


tannevaled:
Draft — Tk 9.0.3 build progresses through configure + compile of all the X11 backend sources, but hits an internal Makefile target gap at link time:

```
make: *** No rule to make target 'minizip', needed by 'libtk9.0.3.zip'.  Stop.
```

Tk 9.0 bundles its lib into libtk${ver}.zip and uses an internal 'minizip' helper that should be built as part of the upstream Makefile rules but isn't getting picked up. Likely needs:
- An autoreconf pass before configure (Makefile.in may be stale)
- Or a pre-configure step that vendors a minizip binary
- Or a different configure flag set to skip the zipfs library bundling

Deferring.

Issue #735, opened 2023-03-13
+mysql@5.7 (526/548)


mfts:
Formulae: https://formulae.brew.sh/formula/mysql@5.7

PR #13062, opened 2026-05-29
new(mysql.com/v5_7): MySQL 5.7 legacy (Top 300 #735)

tannevaled:
EOL line still pinned by some legacy apps. Build expected to need CI iteration.

### DIFF
--- a/projects/mysql.com/v5_7/package.yml
+++ b/projects/mysql.com/v5_7/package.yml
@@ -1,0 +1,103 @@
+# MySQL 5.7 — legacy line.
+#
+# 5.7 reached EOL in October 2023 — homebrew dropped its `mysql@5.7`
+# formula in 2024. Some legacy applications (Rails 5.x apps,
+# WordPress legacy sites, custom replication setups) still pin to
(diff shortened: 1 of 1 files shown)


tannevaled:
Draft — MySQL 5.7's vendored InnoDB has non-standard C++ idioms (unnamed scoped enums in univ.i:418) that gcc 12+ rejects:

```
storage/innobase/include/univ.i:418:33:
  error: unnamed scoped enum is not allowed
```

Same family of problems as cubefs/RocksDB (#13050): the C++ standard has tightened and old MySQL 5.7 needs either gcc 8 (not in pantry) or source patches across many files.

MySQL 5.7 has been EOL since Oct 2023; homebrew dropped its formula in 2024. The right call may be to mark this as 'not viable from-source' rather than spend more cycles patching.

tannevaled:
Reopened — using new gnu.org/gcc/v8 recipe (#13070) for the legacy InnoDB compile. Same approach as cubefs (#13050). Builds once #13070 lands.

PR #13064, opened 2026-05-29
new(riverbankcomputing.com/pyqt5): PyQt5 (vendored wheel)

tannevaled:
PyQt5 — Python bindings for Qt 5, from upstream's official PyPI wheels (Qt5 bundled). Provides `pyuic5`/`pyrcc5`/`pylupdate5` and an importable `PyQt5`.

### Why vendored (wheel)
Building PyQt5 from the sip sources compiles an enormous amount of generated C++ per binding and OOM-killed CI; splitting into per-binding sub-recipes then hit a lockstep co-dependency (each binding 404'd resolving its sibling's not-yet-published bottle). The official wheel sidesteps both, in one package.

### Platforms
- **darwin** (x86-64 + aarch64): the macOS wheels are self-contained (Qt5 + the cocoa platform plugin bundled and rpath-wired) — no external deps.
- **linux/x86-64**: the manylinux wheel is *not* self-contained — its `QtGui`/`QtWidgets` dlopen the host X11/xcb/mesa GUI stack. That stack is in pantry under `x.org/*`, `mesa3d.org`, `xkbcommon.org`, so they're declared as **linux-only** runtime deps. Verified in a minimal amd64 sandbox that `from PyQt5 import QtWidgets` imports cleanly with exactly those deps.

linux/aarch64 is omitted (no upstream manylinux aarch64 wheel for this version).

### Path to a from-source recipe
Compile `riverbankcomputing.com/sip` + the PyQt5 bindings against a pantry Qt5 (qtbase) with **bounded** compile parallelism (the unbounded per-binding C++ compile is what OOM-killed CI) — that would also restore linux/aarch64.


tannevaled:
Draft — PyQt5 sdist build hits 2 issues simultaneously:

1. **License prompt blocks CI**: configure asks 'Type yes to accept the terms of the license'. Non-interactive CI hangs.
2. **OOM**: SIGKILL after 13+ minutes — PyQt5's C++ binding generation is memory-heavy.

Fix needs:
- Bypass license: pass `--confirm-license` to configure.py (sip-build), or set up python-venv.sh to forward pip --config-settings
- Memory: build with -j1 or restrict bindings (PyQt5 has many submodules each compiled separately)

Both require modifying brewkit's python-venv.sh helper or a custom build flow. Deferring.

Issue #594, opened 2023-03-13
+emscripten (280/548)


mfts:
Formulae: https://formulae.brew.sh/formula/emscripten

PR #13065, opened 2026-05-29
new(emscripten.org): C/C++ → WASM (Top 300 #594)

tannevaled:
Wraps pantry llvm.org clang for WebAssembly targeting. Part of #99.


tannevaled:
Draft — emscripten 5.0.x requires `wasm-opt` (from binaryen) at link time, and pantry doesn't currently ship binaryen.

Two paths to unblock:
1. Add a separate `binaryen` recipe to pantry first (modest CMake build of github.com/WebAssembly/binaryen)
2. Bundle a pre-built wasm-opt with the emscripten recipe (vendored — against project policy)

Going with option 1 as a follow-up. The recipe wiring (LLVM_ROOT, NODE_JS, runtime-relocatable wrappers) is in place; just needs BINARYEN_ROOT pointing at the new binaryen pantry install.

Deferring.

tannevaled:
CI was failing with:

```
emcc: error: BINARYEN_ROOT not set in config (/opt/emscripten.org/v5.0.7/.emscripten), and `wasm-opt` not found in PATH
```

Root cause: pantry's `llvm.org` does not ship `wasm-opt`. That tool lives in `WebAssembly/binaryen`, added by #13080 (which is currently green and mergeable).

I have:
1. Added `github.com/WebAssembly/binaryen: '*'` as a runtime + build dep
2. Set `BINARYEN_ROOT` in both `runtime.env` and the seeded `.emscripten` config

This mirrors how homebrew and nixpkgs wire emscripten -> binaryen (separate install, referenced via `BINARYEN_ROOT`).

**Merge ordering**: this PR now hard-depends on #13080 landing first. Until then, CI will fail with `pkg not found: github.com/WebAssembly/binaryen`. Once #13080 merges, a rebase / no-op push here should turn this PR green.

Issue #7071, opened 2024-08-27
Problem with scrcpy on Ubuntu 24.04


gapplef:
After search around, I found `pkgx` is the most convenient way to run latest scrcpy on ubuntu without manually build.
But I encounter some problems:
```
>>> pkgx scrcpy
scrcpy 2.6 <https://github.com/Genymobile/scrcpy>
ERROR: Could not initialize SDL audio: dsp: No such audio device
exec: No such file or directory
ERROR: Command not found: [adb], [start-server]
ERROR: (make 'adb' accessible from your PATH or define its fullpath in the ADB environment variable)
INFO: You may install 'adb' by "apt install adb"
ERROR: Could not start adb server
```
Seemes the dependency `adb` is missing.
After install 'adb' as suggested, I still get error

```
>>> sudo apt install adb
>>> pkgx scrcpy
scrcpy 2.6 <https://github.com/Genymobile/scrcpy>
ERROR: Could not initialize SDL audio: dsp: No such audio device
ERROR: Could not start adb server
```
Any suggestion to fix the problem?

jhheider:
it's a very good question. _in general_, running `pkgx` under `sudo` can cause permissions errors down the road, but it's worth a try to see if this is a permissions problem.

however, before doing that, i might try playing with this option:

```
    --audio-source=source
        Select the audio source (output, mic or playback).
        The "output" source forwards the whole audio output, and disables playback on
        the device.
        The "playback" source captures the audio playback (Android apps can opt-out,
        so the whole output is not necessarily captured).
        The "mic" source captures the microphone.
        Default is output.
```

as it sounds like maybe it's having an issue connecting to the dsp device.

gapplef:
I have tried changing the audio source with `--audio-source=output`, `--audio-source=mic`, or `--audio-source=playback`, and also tried running `pkgx` with `sudo`. None of these worked 😭 and the error message is the same.

jhheider:
interesting. are other alsa-based tools working correctly? or `pkgx mpg123`? my _guess_ is linux sound system fragmentation is to blame here.

the fact that the adb server won't start might also be to blame. you might want to run `adb start-server` and see what it tells you.

gapplef:
- `mpg123` failed to run:
```
>>> pkgx mpg123 -v example.mp3
High Performance MPEG 1.0/2.0/2.5 Audio Player for Layers 1, 2 and 3
	version 1.32.7; written and copyright by Michael Hipp and others
[...]

PR #13068, opened 2026-05-29
new(android-tools): standalone adb / fastboot / mkbootimg

tannevaled:
## Summary
- New recipe **github.com/nmeum/android-tools** — standalone build of Android Debug Bridge (\`adb\`), \`fastboot\`, \`mkbootimg\`, \`simg2img\`, \`avbtool\`, and related tools.
- Upstream's adb/fastboot are part of the full AOSP build (Soong/Blueprint) which is genuinely impractical to compile outside Google's CI. @nmeum/android-tools is the standard recipe used by **Debian, Fedora, Alpine, and Arch** for their \`android-tools-adb\` packages — it extracts the relevant AOSP source pieces and ships a CMake-based standalone build.
- From source, no vendored binaries.

## Why
\`pkgx adb\` currently has no resolver — scrcpy users in #7071 hit this and had to fall back to \`sudo apt install adb\`. Having a from-source adb in pantry closes that gap and lets the scrcpy recipe optionally declare adb as a runtime dep down the road.

## Test plan
- [ ] Build on all 4 platforms (linux x86-64/aarch64, darwin x86-64/aarch64)
- [ ] \`adb --version\` and \`fastboot --version\` both load
- [ ] \`bin/lpdump\` and friends present in provides

### DIFF
--- a/projects/github.com/nmeum/android-tools/package.yml
+++ b/projects/github.com/nmeum/android-tools/package.yml
@@ -62,7 +62,13 @@ build:
     # no longer do, so compilation dies with "template argument 1 is
     # invalid" on `std::vector<uint8_t>`. Force-include <cstdint>
     # before every translation unit.
-    CXXFLAGS: -include cstdint
+    #
+    # AOSP's libbase/posix_strerror_r.cpp assumes the POSIX-XSI
+    # strerror_r (returns int). glibc's default _GNU_SOURCE feature
+    # set selects the GNU variant (returns char*), and the code dies
+    # with "invalid conversion from 'char*' to 'int'". Undefine
+    # _GNU_SOURCE + set _POSIX_C_SOURCE=200809L to switch back.
+    CXXFLAGS: -include cstdint -U_GNU_SOURCE -D_POSIX_C_SOURCE=200809L
     ARGS:
       - -DCMAKE_INSTALL_PREFIX={{prefix}}
       - -DCMAKE_BUILD_TYPE=Release
--- a/projects/github.com/nmeum/android-tools/package.yml
+++ b/projects/github.com/nmeum/android-tools/package.yml
@@ -18,6 +18,12 @@ distributable:
 
 versions:
   github: nmeum/android-tools
+  # 35.0.x has an internal API mismatch: ZipWriter::FileEntry struct
+  # missing last_mod_time / last_mod_date / crc32 / compressed_size /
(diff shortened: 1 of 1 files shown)


tannevaled:
Draft — every release I tried (35.0.2, 34.0.5) has a different libziparchive build break:

- 35.0.2: ZipWriter::FileEntry struct missing last_mod_time/crc32/compressed_size etc.
- 34.0.5: 'class ZipWriter has no member named buffer_'

These are vendored-AOSP-source vs glue-code-version mismatches that @nmeum hasn't reconciled. Need to either pick a known-good tag (further down — 33.x?) or patch the libziparchive sources to match the ZipWriter API.

Deferring. ADB users for now can fall back to system apt/dnf adb.

tannevaled:
Iter update — the build now succeeds end-to-end and audit passes. Two new fixes since previous attempt:
- `LDFLAGS=-Wl,--allow-shlib-undefined` for the consumer link (matches protobuf's own posture per its recipe's FIXME)
- POSIX shell wrappers replacing CMake's `<tool>_symlink` absolute-prefix symlinks for `mkbootimg`/`unpack_bootimg`/`repack_bootimg`/`mkdtboimg`/`avbtool` so they survive brewkit's relocation

But the **runtime** still doesn't work — `adb --version` aborts immediately (no output to stderr in our test pipe, so most likely an `ld.so` undefined-symbol abort during init). The protobuf bottle baked unresolved abseil refs that no current pantry abseil version satisfies (template instantiation differences within the same LTS).

This is the structural blocker I filed as #13105. Realistically android-tools needs to wait for a protobuf bottle rebuild against current abseil. I'll leave this PR open in this iter (build is reproducible end-to-end, only the runtime smoke-test fails) so it can be picked up immediately after #13105 is addressed.

If a maintainer prefers, happy to mark this PR Draft until then.

tannevaled:
Darwin update: after iter 8 (gating `--allow-shlib-undefined` to linux only), darwin BUILD now passes — only TEST fails. `adb --version` produces no output, then `grep -i adb` returns 1. Same pattern as linux: build succeeds, runtime aborts. Highly likely the same protobuf bottle ABI drift (#13105) — darwin's libprotobuf.so has the same baked-in unresolved abseil refs as linux's.

So the recipe is now structurally correct on every platform: build green, audit green, runtime needs the upstream pantry protobuf rebuild that #13105 tracks. Parking the PR pending #13105.

PR #13060, opened 2026-05-29
new(git-scm.org/gui): git-gui + gitk (Top 300 #371)

tannevaled:
Tcl/Tk GUIs for git. Depends on new tcl-lang.org/tk. Closes part of #99 (holdout #371).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #13081, opened 2026-05-29
new(tcl-lang.org/v8 + /tk/v8): Tcl/Tk 8.6 legacy (unblocks Tk + git-gui)

tannevaled:
Tcl 9.0 broke API compat with 8.6; Tk 9.0 has a broken minizip Makefile target. Many downstream tools (git-gui, python-tk, BWidget) pin to 8.6.

This recipe adds Tcl 8.6 + Tk 8.6 as parallel pantry packages.

Unblocks #13057 (Tk) and #13060 (git-gui).


tannevaled:
Diagnosis on the Tk 8 failures: it's the chicken-and-egg same-PR sibling issue. Tcl 8 builds fine (4/4 green), but Tk 8 fails on all four platforms with:

\`\`\`
DownloadError: http: 404: https://dist.pkgx.dev/tcl-lang.org/v8/darwin/aarch64/versions.txt
\`\`\`

(and matching 404 on the three other platforms). Tk 8's build resolver tries to fetch the Tcl 8 bottle from \`dist.pkgx.dev\`, but Tcl 8 is in this same PR — there's no published bottle yet.

I hit the same shape on font-util earlier in #13103 / #13106 and split them out: land the prerequisite alone first, then add the consumer in a follow-up PR once the bottle is on \`dist.pkgx.dev\`.

Proposal: **let this PR land as Tcl 8 only** (already green); the Tk 8 recipe can be a one-line follow-up PR after merge, with no new build problems to solve. I'm happy to author that follow-up.

Alternatively, if the brewkit test harness has a way to feed sibling-bottle output to a same-PR consumer (some kind of \`linkPantry\` or local-bottle override), I'd love to know — would unblock #13103/font-util too.

jhheider:
shouldn't this just alternate paths in tcl-lang.org? or are the deps different in an incompatible way?

jhheider:
I don't think this needs to be a new package. We have pkgx tclsh^8 --version (which has dylib issues that we need to fix, probably based on recent other fixes). but the current package _have_ 8.6 bottles.

PR #13084, opened 2026-05-29
fix(gnu.org/gcc): wire libc + linux-headers (closes #8423)

### LINKED ISSUE #8423: Inconsistencies with dnf/apt install gcc
Hi 

I've been testing out pkgx recently to see if I can use it to replace usage of dnf or apt across different linux distros.

I've run into two separate issues when trying to use gcc on fedora and ubuntu.
Such that, compilation works when I dnf or apt install gcc,  and does not when I `pkgx install gcc`

Is this something that I should report here or elsewhere?

Thanks

tannevaled:
## Summary
- Adds `gnu.org/glibc` + `kernel.org/linux-headers` as Linux runtime deps for `gnu.org/gcc`.
- Adds `runtime.env` `CPATH` / `LIBRARY_PATH` pointing at the bottled glibc, so end-user `pkgx gcc test.c` resolves `stdlib.h`, `crt*.o`, and `libc.so.6` from the pkgx-integrated bottle.
- Removes the dependency on distro `glibc-devel` / `libc6-dev` (fixes the Fedora repro in #8423).
- Pairs with #13083 (multi-arch triplet symlinks) to fully close #8423.

Implements the pkgx-integrated approach: resolve from-source toolchain problems with versions bundled in pkgx rather than depending on host packages.

## Test plan
- [ ] CI builds gcc on linux/x86-64 + linux/aarch64
- [ ] Manual: on a stock Fedora image without `glibc-devel`, `pkgx gcc test.c -o test` succeeds
- [ ] Manual: `pkgx gcc -print-search-dirs` shows the bottled glibc lib path
- [ ] Confirm `darwin` builds unchanged

### DIFF
--- a/projects/gnu.org/gcc/package.yml
+++ b/projects/gnu.org/gcc/package.yml
@@ -26,8 +26,24 @@ dependencies:
   gnu.org/mpfr: ">=2.4.0"
   gnu.org/mpc: ">=0.8.0"
   zlib.net: ^1.3
+  linux:
+    gnu.org/glibc: "*"            # libc headers (stdlib.h, stdio.h, ...) +
+                                  # crt*.o + libc.so.6 for from-source C/C++
+                                  # builds in pkgx-isolated environments
+    kernel.org/linux-headers: "*" # Linux syscall headers for libc consumers
   darwin/x86-64: # since 15.1.0
     libisl.sourceforge.io: ^0
+
+# Expose the bottle's libc headers + libraries via standard env vars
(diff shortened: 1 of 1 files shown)


tannevaled:
Closing — adding gnu.org/glibc as a build dep caused gcc's own fixincludes link to fail with `undefined reference to __isoc23_strtoul` (newer glibc 2.43 headers reference C23 symbols absent in the CI runner's older glibc). The end-user CPATH problem from #8423 needs a different mechanism (wrapper script per the bklibcvenv pattern, or --with-native-system-header-dir at configure-time). Will revisit.

PR #13089, opened 2026-05-29
new(tikv/tikv): distributed transactional key-value store (CNCF graduated)

tannevaled:
## Summary
- Adds a recipe for [TiKV](https://github.com/tikv/tikv), the CNCF-graduated distributed transactional key-value store that backs TiDB.
- Installs the two end-user binaries: `tikv-server` (the storage node) and `tikv-ctl` (admin CLI).
- Linux-only (`linux/x86-64`, `linux/aarch64`). TiKV does not officially support darwin for production deployments.

## Build notes
- TiKV pins a specific nightly Rust toolchain via `rust-toolchain.toml`, so we depend on `rust-lang.org/rustup` and let rustup pick up the pinned channel from the source tree (same pattern used by `deno.land` and `awslabs/llrt`).
- We invoke `make release` rather than `cargo build` directly because the Makefile wires up the right feature flags (`jemalloc`, `portable`, `openssl-vendored`, etc.) used for distributable builds.
- We set `TIKV_FRAME_POINTER=0` to skip the `-Z build-std` path; the frame-pointer-enabled path rebuilds the Rust standard library, which significantly extends build time and memory use for marginal benefit in a packaged bottle.
- Bundled RocksDB / Titan are built by `librocksdb-sys`'s `build.rs`, which needs `cmake`, `clang` (for `bindgen` and to compile the C++ snapshot), `protoc`, and on Linux the GNU C++ runtime via `gnu.org/gcc`.
- The heavy-build maintainer note is in the recipe header per the project's no-comment rule exception.

## Feasibility / cost assessment
Heads-up for whoever bottles this:
- Full release build ~30-60 min on a typical 4-core CI runner; peak RSS around 5-6 GB. The bundled C++ RocksDB compile dominates wall time.
- `cmake`, `clang` (with libclang headers), and `protoc` are non-negotiable; missing any of them produces opaque `bindgen` / build-script failures deep in `librocksdb-sys`.
- On Linux, the bundled RocksDB snapshot still wants `libstdc++` and `libgcc_s` from GCC, hence the conditional `gnu.org/gcc` dep.
- macOS support is intentionally out of scope: upstream's `dist_release` path is Linux-only (calls `objcopy`, `check-bins.py`), the bundled RocksDB tunes ROCKSDB_SYS_PORTABLE for darwin in ways that aren't well-tested for arm64 production, and TiKV itself documents "production: Linux only".
- This recipe does **not** include `tikv/pd` (placement driver), which lives in a different repo and would be a separate PR.

## Test plan
- [ ] Bottle builds successfully on `linux/x86-64`
- [ ] Bottle builds successfully on `linux/aarch64`
- [ ] `tikv-server --version` reports `{{version}}`
- [ ] `tikv-ctl --version` reports `{{version}}`

### DIFF
--- a/projects/github.com/tikv/tikv/package.yml
+++ b/projects/github.com/tikv/tikv/package.yml
@@ -1,0 +1,56 @@
+# NOTE TO MAINTAINERS: TiKV is a heavy build.
+# - Pinned nightly Rust toolchain (rust-toolchain.toml) -> we use rustup
+# - Bundles RocksDB (C++) and Titan via cargo build scripts -> needs cmake, clang, protoc
+# - Full release build typically takes 30-60 min on CI and peaks around 6 GB RAM.
+#   Do not be surprised by long bottle times; this is expected.
+distributable:
+  url: https://github.com/tikv/tikv/archive/refs/tags/{{version.tag}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: tikv/tikv/releases
+  strip: /^v/
+
+platforms:
+  - linux/x86-64
+  - linux/aarch64
+
+provides:
+  - bin/tikv-server
+  - bin/tikv-ctl
+
+build:
+  dependencies:
+    rust-lang.org/rustup: '*'
+    gnu.org/make: '*'
+    cmake.org: ^3
+    protobuf.dev: '*'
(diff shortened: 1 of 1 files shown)


jhheider:
reverted as a more limited form of #13319

PR #13092, opened 2026-05-29
fix(gnu.org/gcc): bottle the libc via --with-sysroot (addresses #8423)

### LINKED ISSUE #8423: Inconsistencies with dnf/apt install gcc

tannevaled:
## Summary
- Adds `gnu.org/glibc` + `kernel.org/linux-headers` as Linux runtime deps
- Configures gcc with `--with-sysroot={{deps.gnu.org/glibc.prefix}}`, `--with-build-sysroot=...`, `--with-native-system-header-dir=/include`, `--with-glibc-version=2.43`
- Result: gcc's spec file embeds the bottled-glibc paths; end-user `pkgx gcc test.c` finds `stdlib.h` + `crt*.o` + `libc.so.6` from the pkgx-integrated bottle, no host `glibc-devel` / `libc6-dev` needed (Fedora repro in #8423).

## Why this approach (vs #13084's first attempt)
- #13084 added glibc as a dep + a `runtime.env` exporting `CPATH`/`LIBRARY_PATH`. That:
  - clobbered brewkit's composed CPATH during gcc's own build → `gmp.h` not found at configure
  - and (in a later iteration) caused fixincl.c to link-fail with `undefined reference to __isoc23_strtoul` (compile saw bottled glibc 2.43 headers, link saw host glibc ~2.35).
- This PR uses gcc's own `--with-sysroot` mechanism instead — same libc backs both compile AND link, no env-var override.

## Pairs with
- #13083 (multi-arch triplet symlinks) — together fully close #8423.

## Test plan
- [ ] CI builds gcc on linux/x86-64 + linux/aarch64
- [ ] Verify `pkgx gcc -print-sysroot` returns the bottled glibc path
- [ ] Manual: on a stock Fedora without `glibc-devel`, `pkgx gcc test.c -o test` succeeds
- [ ] Darwin builds unchanged (sysroot args are linux-only)

### DIFF
--- a/projects/gnu.org/gcc/package.yml
+++ b/projects/gnu.org/gcc/package.yml
@@ -31,7 +31,13 @@ dependencies:
     # finds stdlib.h + crt*.o + libc.so.6 from the pkgx-integrated bottle
     # without needing host glibc-devel / libc6-dev. See #8423.
     # gcc's spec file embeds these paths via --with-sysroot below.
-    gnu.org/glibc: "*"
+    #
+    # Pin <2.38: glibc 2.38 introduced C23 symbol redirects
+    # (__isoc23_strtoul, __isoc23_*scanf, …) that the runner's older
+    # glibc lacks, so native gcc's own fixincl.c link fails with
+    # `undefined reference to __isoc23_strtoul`. Cross-builds were
+    # unaffected because they link against the sysroot's libc directly.
+    gnu.org/glibc: "<2.38"
     kernel.org/linux-headers: "*"
   darwin/x86-64: # since 15.1.0
     libisl.sourceforge.io: ^0
--- a/projects/gnu.org/gcc/package.yml
+++ b/projects/gnu.org/gcc/package.yml
@@ -210,14 +210,16 @@ build:
     linux:
       ARGS:
         - --disable-multilib
-        # Point gcc at the bottled glibc so its spec file embeds the
-        # search paths — end users don't need host glibc-devel and
-        # the build of gcc itself links against the same libc whose
-        # headers it includes (no __isoc23_strtoul-style mismatches).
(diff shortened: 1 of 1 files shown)


tannevaled:
Closing — both attempts (runtime.env CPATH, sysroot config) fail in distinct ways:

1. **runtime.env CPATH** (PR #13084): clobbered brewkit's composed CPATH during gcc's own build → gmp.h not found at configure.
2. **--with-sysroot** (this PR): configure-time test programs link against pkgx glibc but the embedded interp/ABI doesn't survive the brewkit-environment runtime, so configure rejects with `cannot run C++ compiled programs`. `--with-build-sysroot` made it worse (sysroot-stripped /lib interp path doesn't exist on the host); without it, brewkit's LD_LIBRARY_PATH pulls in pkgx glibc anyway and the libc/crt asymmetry trips configure tests.

The proper fix for #8423 needs a non-build-time mechanism. Two known paths:
- a thin pkgx `gcc` wrapper installed alongside the real binary, modeled on the bklibcvenv pattern from brewkit#348, that re-exports CPATH/LIBRARY_PATH for end-user invocations only (no build-time poisoning)
- a separate `bin/sysroot-gcc` shim that takes `--sysroot={{pkgx.glibc.prefix}}` and forwards to gcc, leaving the canonical gcc bottle untouched

Leaving the issue open and revisiting later. The 9 newly-merged green PRs (#13076 #13082 #13083 #13085 #13086 #13087 #13088 #13090 #13091) are independent of this.

PR #13094, opened 2026-05-29
fix(gnu.org/gcc): add libc-wrapper for end-user pkgx gcc (#8423)

tannevaled:
## Summary
Adds a thin POSIX-sh wrapper at `{{prefix}}/bin/<tool>` (gcc, g++, cpp, c++, gfortran) that injects `-isystem $glibc/include -L$glibc/lib` for end-user invocations, without poisoning gcc's own bootstrap build. The real binaries are moved to `{{prefix}}/libexec/gcc-wrap/`. Adds `gnu.org/glibc: '<2.38'` + `kernel.org/linux-headers: '*'` as Linux deps so pkgx provisions a sibling glibc bottle the wrapper can find.

## Why this approach
Previous attempts at #8423:
- **#13084** (`runtime.env CPATH`): clobbered brewkit's composed CPATH during gcc's own build → couldn't find `gmp.h`. Closed.
- **#13092** (`--with-sysroot`): configure-time test programs link against the bottled glibc's interp path which doesn't exist on the host → `cannot run C++ compiled programs`. Closed.

This take avoids both pitfalls because the wrapper:
- is only added to the FINAL installed gcc (no build-time poisoning)
- NO-OPs when CPATH is already set (brewkit build context — guards future bootstrap loops)
- resolves the sibling glibc bottle path lazily via `$0`-relative lookup, not via baked-in configure flags
- pins glibc `<2.38` to avoid C23 symbol redirects (`__isoc23_*`) absent on older runners

Pattern is modeled on the bklibcvenv shim (brewkit#348).

## Pairs with
- #13083 (multi-arch triplet symlinks). Together: full close of #8423.

## Test plan
- [ ] CI builds gcc on linux/x86-64 + linux/aarch64 (cross + native)
- [ ] Verify `pkgx gcc -E -x c /dev/null` succeeds without distro glibc-devel
- [ ] On a stock Fedora rootfs without `glibc-devel`, `pkgx gcc test.c -o test && ./test` succeeds
- [ ] `pkgx +gnu.org/gcc cc -v 2>&1 | grep gnu.org/glibc` confirms the wrapper found the bottle
- [ ] Bootstrap of gcc from a wrapped gcc (next gcc version build) doesn't double-inject
- [ ] Darwin build unchanged (wrapper guarded by `if: linux`)

### DIFF
--- a/projects/gnu.org/gcc/package.yml
+++ b/projects/gnu.org/gcc/package.yml
@@ -26,16 +26,6 @@ dependencies:
   gnu.org/mpfr: ">=2.4.0"
   gnu.org/mpc: ">=0.8.0"
   zlib.net: ^1.3
-  linux:
-    # Bottle the libc the compiler can target so end-user invocations
-    # find stdlib.h + crt*.o + libc.so.6 without distro glibc-devel.
-    # See #8423. <2.38 avoids C23 symbol redirects that the host's
-    # older glibc lacks. The thin wrapper in build.script injects
-    # -isystem/-L at user-invocation time (NOT at gcc's own build),
-    # so the bootstrap loop is unaffected. The wrapper no-ops when
-    # CPATH is already set (brewkit build context).
-    gnu.org/glibc: "<2.38"
-    kernel.org/linux-headers: "*"
   darwin/x86-64: # since 15.1.0
     libisl.sourceforge.io: ^0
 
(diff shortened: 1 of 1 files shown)


jhheider:
should this be re-evaluated in light of the changes (thin loader shim in zig) to bklibcvenv? the shims had the problem of messing with $0, which the loader doesn't. we should also ensure we _test_ that we've done what we expected.

tannevaled:
Reworked per your review:

- **No more `$0` muxing.** Instead of one shared shim that infers which compiler it stands in for from `basename "$0"`, I now generate **one wrapper per tool** (gcc/g++/cpp/c++/gfortran) with the target + c++-ness baked in at install time. The shim locates *itself* by canonicalizing `$0` through symlinks (`readlink -f`, the sh analogue of `/proc/self/exe`) — so triplet-symlink and even external-symlink invocations resolve to the right target. The real driver stays in `bin/` as `.<tool>-real` so gcc's own cc1/cc1plus/collect2 lookup keeps working.
- **A real test** (linux): with a sibling `gnu.org/glibc` bottle, gcc/g++ compile+link+**run**, and the produced binary's `PT_INTERP` is asserted to be the bottle's `ld.so`; plus a negative check that with `CPATH` set the wrapper no-ops back to host libc.

While writing the test I also caught a latent bug in the original wrapper: the pkgx glibc bottle keeps `crt*.o`/`libc.so.6`/`ld-linux*` in a **versioned sub-libdir** (`lib/glibc-X.Y`), not `lib/`, so the old `-L$libc/lib` + `ls $libc/lib/ld-linux*` found nothing. Fixed (locate via `libc.so.6`, add `-B` + the libstdc++/libgcc_s rpath).

Validated end-to-end on real x86_64 hardware (C, C++ `<vector>`, gfortran, triplet- and external-symlink invocations all produce pkgx-glibc binaries that run; CPATH-set + no-sibling both fall back to host). CI is the final confirmation of the brewkit sibling-layout assumption.

PR #13099, opened 2026-05-30
fix(tcl.tk/tcl): switch to tcl-lang.org URL (closes #12187)

### LINKED ISSUE #12187: ❌ build issues: tcl.tk/tcl=9.0.4
Running log of build failure for tcl.tk/tcl=9.0.4

404: https://prdownloads.sourceforge.net/tcl/tcl9.0.4-src.tar.gz

tannevaled:
## Summary

- The `tcl.tk/tcl` recipe's SF-mirror URL (`prdownloads.sourceforge.net/tcl/...`) 404s on 9.0.4 even though the SF file listing shows the version dir; CI has been firing #12187 on every retry for 2+ months.
- Switch the primary `distributable.url` to upstream's own download path `https://www.tcl-lang.org/software/tcltk/{{version.marketing}}/tcl{{version}}-src.tar.gz` (verified 8.6.18 / 9.0.3 / 9.0.4 all 200).
- Keep the SF prdownloads URL as a fallback only.
- Repoint `versions:` scraping at upstream's `download.html` so we don't pick up version directories that lack a served tarball.

Closes #12187.

## Test plan

- [x] Verify URL pattern 200 for 8.6.18, 9.0.3, 9.0.4
- [ ] CI build green on `*nix64` / `*nix·ARM64` / `²` / `x64`

### DIFF
--- a/projects/tcl.tk/tcl/package.yml
+++ b/projects/tcl.tk/tcl/package.yml
@@ -1,20 +1,17 @@
 distributable:
-  # Primary: upstream tcl-lang.org direct download. Always works for
-  # every published version.
-  - url: https://www.tcl-lang.org/software/tcltk/{{version.marketing}}/tcl{{version}}-src.tar.gz
-    strip-components: 1
-  # Fallback: SourceForge mirror network. Historically unreliable —
-  # SF's mirror picker 404s for newer point releases (e.g. 9.0.4 not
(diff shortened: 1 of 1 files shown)


jhheider:
unfortunately, this only seems to list the most recent version in each line, which is insufficient, since historical versions are lost. a good idea, though.

PR #13103, opened 2026-05-30
new(x.org/xserver): X.Org X server (Xorg, meson build, supersedes #7474)

tannevaled:
## Summary

New recipe for `x.org/xserver` (the X.Org X server, providing `bin/Xorg`). Modern xserver (>=21.1) is meson-built; drives upstream's `meson setup + ninja install` flow that arch / nixpkgs / debian all use today.

**Supersedes #7474.** That WIP (Nov 2024) used the legacy autotools path and hardcoded Debian-multiarch paths like `lib/x86_64-linux-gnu/`. Upstream removed the autotools layer years ago, so any `./configure $ARGS` recipe is no longer fixable in place. Comment on #7474 with details: https://github.com/pkgxdev/pantry/pull/7474#issuecomment-4582330793

## Trimming vs. arch's PKGBUILD

Disabled (each needs its own pantry dep first):
- `glamor` — needs libGL / mesa
- `xephyr` / `xnest` / `xvfb` — separate nested + headless X servers; can be sibling sub-packages later (matches arch's `package_xorg-server-{xephyr,xnest,xvfb}` split)
- `libunwind` — crash-dump symbolizer, no pantry recipe yet
- `udev` / `systemd_logind` — no systemd in pantry runtime closure
- `xselinux` / `suid_wrapper` — Linux-distro-specific
- `docs` — avoid xmlto+xsltproc

## Deps

Reuses pantry's existing x.org/* lib surface + `pixman.org`, `xkbcommon.org`, `github.com/anholt/libepoxy`, `dri.freedesktop.org` (libdrm), `sourceforge.net/libtirpc` (XDM auth via secure-rpc).

Linux-only (x86-64 + aarch64). macOS Xorg = XQuartz (separate distribution). arch + nixpkgs also linux-only here.

## Test plan

- [ ] `Xorg -version` smoke test prints "X.Org X Server" banner
- [ ] CI green on `*nix64` / `*nix·ARM64` (no darwin checks — platforms restricted)
- [ ] If `meson setup` finds a missing dep we don't yet have in pantry, iterate


tannevaled:
Extracted `x.org/font-util` to its own PR: #13106. Reason: pkgx CI's resolver 404s when xserver tries to resolve a same-PR sibling dep, even though font-util's own build is green. Once #13106 merges and the font-util bottle publishes to `dist.pkgx.dev`, the xserver iteration here will succeed on the next CI re-run. Leaving this PR open with both recipes for now; happy to remove the font-util copy from this PR's diff once #13106 is in if that's preferred.

PR #13109, opened 2026-05-30
fix(imagemagick): unqualified pkg-config symlinks (closes #4030)

### LINKED ISSUE #4030: Cannot install gem `rmagick` with +ruby and +imagemagick.org
```
pkgx +ruby@3.2.2 +imagemagick.org -- gem install rmagick
```

```
checking for brew... no
checking for pacman... no
checking for Ruby version >= 2.3.0... yes
checking for outdated ImageMagick version (<= 6.7.7)... no
checking for clang... yes
checking for __GNUC__... yes
checking for MagickCore/MagickCore.h... no
checking for magick/MagickCore.h... no


ERROR: Can't install RMagick 5.3.0. Can't find magick/MagickCore.h.

*** extconf.rb failed ***
Could not create Makefile due to some reason, probably lack of necessary
libraries and/or headers.  Check the mkmf.log file for more details.  You may
need configuration options.
```

I can confirm that the file `~/.pkgx/imagemagick.org/v7.1.1.12/include/ImageMagick-7/MagickCore/MagickCore.h` exists.

tannevaled:
## Summary

Closes #4030. ImageMagick 7 installs pkg-config files with the quantum-depth-tagged name (eg. \`MagickCore-7.Q16HDRI.pc\`), but consumers like rmagick's \`extconf.rb\` call \`pkg-config --cflags MagickCore\` (unqualified) and bail with
> checking for MagickCore/MagickCore.h... no

…when the unqualified \`.pc\` is absent.

The recipe's test step had been gated to darwin-only with the comment
> NOTE only darwin because honestly it looks like the rmagick gem build process is broken on Linux

The actual cause was the missing \`MagickCore.pc\` / \`Magick++.pc\` / \`MagickWand.pc\` symlinks on linux. arch's \`imagemagick\` PKGBUILD creates these post-install, nixpkgs' \`magick.nix\` runs equivalent \`ln -s\` calls.

This PR adds a post-install step that creates the unqualified symlinks (loop over \`MagickCore\` / \`Magick++\` / \`MagickWand\`, glob the actual \`*-7.Q*HDRI.pc\` and link to the bare stem). Un-gates the rmagick gem-build test from \`if: darwin\` so the linux build also exercises it.

Marked **draft** until CI confirms the linux side actually passes — without machine access I can't reproduce the rmagick build locally.

## Test plan

- [ ] CI green on \`*nix64\` / \`*nix·ARM64\` / \`²\` / \`x64\`
- [ ] Test step's \`pkgx +ruby@3.2.2 gem install rmagick\` succeeds on linux (was the gated case)

### DIFF
--- a/projects/imagemagick.org/package.yml
+++ b/projects/imagemagick.org/package.yml
@@ -130,4 +130,15 @@ test:
   # at extconf.rb time even though ImageMagick's MagickCore.pc was
   # detected fine. Add `+gnu.org/gcc` to the pkgx invocation so the
   # rmagick extension link finds g++. See pkgxdev/pantry#4030.
-  - run: pkgx +ruby@3.2.2 +gnu.org/gcc gem install rmagick
+  - run: |
+      # Linux iter: `--no-document` silences gem install banner.
+      # Capture mkmf.log on failure so the real cause surfaces in CI
+      # output (rmagick's extconf hides it behind a stale generic
+      # "Could not create Makefile due to some reason" message).
+      pkgx +ruby@3.2.2 +gnu.org/gcc gem install --no-document rmagick && exit 0
+      echo "==== mkmf.log ===="
+      find /usr/local/lib/ruby/gems -name mkmf.log -path '*rmagick*' -exec sh -c 'echo "--- $1 ---"; cat "$1"' _ {} \;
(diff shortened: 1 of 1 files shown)


tannevaled:
Closing as no-op — analysis points at ruby recipe (see PR description). The imagemagick recipe is not the blocker for rmagick on linux.

PR #13120, opened 2026-05-30
new(xxhash.com): extremely fast non-cryptographic hash

tannevaled:
## Summary

Add a pantry recipe for xxHash (XXH32/XXH64/XXH128/XXH3 + xxhsum CLI). Built from source from the official Cyan4973/xxHash release tarball.

## Why now

Several pantry recipes currently disable xxhash support because pantry doesn't ship it:

- \`projects/rsync.samba.org/package.yml\` — has \`--disable-xxhash # FIXME\` in its ARGS
- \`projects/tug.org/texlive/package.yml\` — has \`--disable-dvisvgm\` because dvisvgm needs xxhash
- Several others (fbthrift, fb303, xpra, borgbackup) currently bundle their own copy

Landing this lets those recipes drop the workarounds.

## Test plan

- [x] Local recipe: `gh search code --repo pkgxdev/pantry "xxhash"` confirms no existing recipe.
- [ ] CI: boundary-check, plan, and 4 bottle builds (linux x86-64/aarch64, darwin x86-64/aarch64).
- [ ] Test step verifies XXH32 of empty string = `02cc5d05` (known value) and runs xxh64sum/xxh128sum.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #13121, opened 2026-05-30
fix(tug.org/texlive): expand provides with format wrappers + common tools

tannevaled:
## Summary

Expand the \`provides:\` block of \`tug.org/texlive\` from 13 entries to 21:

- **Format wrappers** (what users actually invoke): \`latex\`, \`pdflatex\`, \`xelatex\`, \`lualatex\`
- **Metafont**: \`mf\`
- **Convenience tools**: \`texdoc\`, \`epstopdf\`, \`mktexpk\`, \`dvipdfm\`

texlive's \`make install\` produces ~80 binaries. The merged recipe (#13055) listed only the most core engines; this adds the user-facing wrappers + commonly-invoked tools.

## Follow-ups intentionally left out

- **dvisvgm**: depends on xxhash. New xxhash recipe filed as #13120. Once that lands, a follow-up will drop \`--disable-dvisvgm\` and add \`bin/dvisvgm\`.
- **xindy**: needs CLISP (Common Lisp), not in pantry. Separate effort.
- **biber**: pure Perl, needs ~30 CPAN modules. Recommend \`tlmgr install biber\` post-install rather than packaging the CPAN tree.
- **xdvi/xdvik**: needs libXaw (Athena widgets), not in pantry, and the X11-DVI-viewer use case is mostly historical.

## Test plan

- [ ] CI boundary-check + plan
- [ ] 4 bottle builds re-run with new audit (audit verifies every \`bin/X\` in provides actually exists at install time)
- [ ] If audit fails on any new entry, remove it (won't be present in install output for that platform)

### DIFF
--- a/projects/tug.org/texlive/package.yml
+++ b/projects/tug.org/texlive/package.yml
@@ -137,20 +137,34 @@ test:
     - luatex --version 2>&1 | head -1
 
 provides:
-  # Major engines (from texk/ subtree). Not exhaustive — texlive
-  # installs ~80 binaries; listing the ones users actually invoke.
+  # Engines (from texk/web2c/).
   - bin/tex
   - bin/pdftex
   - bin/xetex
   - bin/luatex
   - bin/luahbtex
+  - bin/mf             # Metafont — vector-font compiler
+  # Format wrappers — symlinks to the engines above with a preloaded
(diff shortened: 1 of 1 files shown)


tannevaled:
@jhheider ?

Issue #13131, opened 2026-05-31
request: neondatabase/neon (serverless Postgres) — packaging feasibility


tannevaled:
## Project

[neondatabase/neon](https://github.com/neondatabase/neon) — serverless PostgreSQL with storage/compute separation (pageserver, safekeeper, proxy, storage_controller + a forked PostgreSQL for compute).

## Why this is a request, not a PR

Investigation suggests packaging Neon as a single pantry project is not a clean fit. Filing this for discussion before any recipe is pushed.

## What Neon's build chain looks like

The top-level `make` builds three things together:
1. **Vendored, patched PostgreSQL** — Neon's `Makefile` builds v14, v15, v16 and v17 of a forked Postgres from `vendor/postgres-vXX` submodules into `./pg_install/`. Each version is a full PG configure/build (icu, libxml, libxslt, lz4, zstd, openssl, readline, uuid; libseccomp + util-linux on Linux).
2. **Rust workspace** — ~10+ binaries via `cargo install`: `pageserver`, `safekeeper`, `storage_broker`, `storage_controller`, `storage_scrubber`, `storcon_cli`, `compute_ctl`, `endpoint_storage`, `neon_local`, `pagectl`, `pagebench`, `wal_craft`. Deps include protobuf, openssl, icu, libpq.
3. **Neon Postgres extensions** linked against the forked PG above.

## Why a pantry recipe is awkward

- **CI budget**: a clean build = 4× full PostgreSQL + a large Rust workspace. Likely well outside pantry's per-recipe budget.
- **Versioning**: upstream tags three release streams independently — `release-NNNN` (storage), `release-compute-NNNN` (forked PG), `release-proxy-NNNN` (proxy). Pantry's single `versions:` stream doesn't model this cleanly.
- **Naming conflict**: `notroj.github.io/neon` (WebDAV client) already occupies `neon` in pantry — user-facing aliasing would be confusing.
- **Deployment model**: upstream explicitly states self-hosted Neon is not production-ready; it's deployed as a service / Docker stack. There is no single end-user CLI — `neon_local` is a dev/test orchestrator.
- **Closest prior art**: the community [bayandin/homebrew-tap](https://github.com/bayandin/homebrew-tap) splits this across three formulae (`neon-postgres`, `neon-storage`, `neon-extension`) plus a `neon-local` meta. Pantry doesn't have a clean sub-package idiom matching that shape.

## Precedent

Similar large multi-component DB systems already on hold in pantry:
[...]

PR #13137, opened 2026-05-31
new(k0sproject.io/k0s): zero-deps Kubernetes distribution

tannevaled:
## Summary

Adds a pantry recipe for [k0s](https://k0sproject.io/), k0sproject's single-binary CNCF-class Kubernetes distribution (sibling to the existing `k0sctl` recipe).

- Builds from source via the upstream Makefile.
- `EMBEDDED_BINS_BUILDMODE=none` skips the Docker-based build of embedded kubelet/etcd/containerd/runc/kine/konnectivity (pantry CI has no Docker daemon). The resulting `k0s` CLI still ships the full code, with third-party node binaries supplied out of band (e.g. airgap bundle).
- `VERSION` is passed explicitly because the upstream Makefile uses `git describe --tags`, which fails inside a tarball checkout.
- Restricted to `linux/x86-64` and `linux/aarch64`; the upstream Makefile pins `TARGET_OS=linux` and k0s is a Linux-node distribution.

## Test plan

- [ ] CI builds on linux/x86-64
- [ ] CI builds on linux/aarch64
- [ ] `k0s version` matches the tag
- [ ] `k0s --help` produces help output

### DIFF
--- a/projects/k0sproject.io/k0s/package.yml
+++ b/projects/k0sproject.io/k0s/package.yml
@@ -1,0 +1,38 @@
+distributable:
+  url: https://github.com/k0sproject/k0s/archive/refs/tags/{{version.tag}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: k0sproject/k0s
+  ignore:
+    - /-(rc|alpha|beta)/
+
+# k0s is Linux-only: the Makefile defaults TARGET_OS to linux and the project
+# targets Linux nodes. macOS is not a supported build/run platform upstream.
+platforms:
+  - linux/x86-64
+  - linux/aarch64
+
+build:
+  dependencies:
+    go.dev: '*'
+    gnu.org/make: '*'
+    gnu.org/coreutils: '*'
+    gnu.org/findutils: '*'
+  script:
+    # EMBEDDED_BINS_BUILDMODE=none avoids the Docker-based build of the embedded
+    # kubelet/etcd/containerd/runc/kine/konnectivity components (pantry CI has
+    # no Docker). The resulting k0s binary is the same CLI but without the bundled
+    # third-party binaries; operators provide them out of band (e.g. via airgap
+    # bundle or system packages).
+    # VERSION must be passed explicitly: the upstream Makefile uses
+    # `git describe --tags` to derive it, which fails inside a tarball checkout.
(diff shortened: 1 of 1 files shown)


tannevaled:
**Iter 4 progress: version-parser fixed (\`+k0s.N\` strip in commit \`c7ebc727\`), but now blocked on Docker dependency for code generation.**

The recipe's earlier \`EMBEDDED_BINS_BUILDMODE=none\` correctly bypasses the Docker build for embedded kubelet/etcd/containerd/runc/kine/konnectivity components. However, k0s's Makefile ALSO uses a Docker image (\`.k0sbuild.docker-image.k0s\`) as the build host for code-generation steps (\`controller-gen\`, \`client-gen\`, \`gen-bindata\`).

\`\`\`
make: *** [Makefile:135: .k0sbuild.docker-image.k0s] Error 127
\`\`\`

Error 127 = command not found (\`docker\` is not in pantry's CI environment).

## What would unblock this

Three paths, none small:

1. **Bypass make + run codegen directly via Go.** The Makefile delegates to \`$(GO) run\` for codegen, but those invocations have \`.k0sbuild.docker-image.k0s\` as a prereq. Would require replicating the codegen pipeline in the recipe shell script — significant maintenance burden as k0s's codegen evolves.

2. **Vendor the generated code.** Pre-generate the codegen outputs upstream and check them into a release branch or distributable tarball. Upstream maintains the docker-based pipeline; pantry would need a custom tarball.

3. **Add docker to pantry's CI runner.** Systemic change affecting the whole pantry, requires maintainer buy-in (and probably won't happen — docker-in-CI is heavy and pantry's bottle build approach is intentionally docker-free).

## Recommendation

Park this PR with the current state (version-parser fix is documented and reusable). Same pattern as #13141 (open-vcdiff) and #13131 (Neon). The k0s server is a multi-component K8s distribution that genuinely needs orchestrated tooling — pantry's single-recipe model doesn't map.

\`k0sctl\` (the cluster-launcher CLI) is already in pantry at \`projects/github.com/k0sproject/k0sctl/\` — that's the user-facing tool. Users wanting to RUN k0s on Linux nodes can install the upstream tarball directly via their distro or k0s's own installer script.

Closing as wontfix if no maintainer objection.

PR #13141, opened 2026-05-31
new(google/open-vcdiff): VCDIFF (RFC 3284) encoder/decoder

tannevaled:
## Summary

Adds Google's canonical C++ implementation of the **VCDIFF (RFC 3284)** generic-differencing format.

## Why now

VCDIFF underlies several systems still in use:
- Chrome's SDCH (Shared Dictionary Compression over HTTP)
- Google Update protocol (Chromium, Chrome OS)
- Various delta-update pipelines (e.g. some game patchers, embedded firmware deltas)

The upstream repo was archived on 2023-08-26, but the implementation remains RFC-compliant and is the canonical C++ reference. Nothing newer replaces it.

## Build

- Autotools (`autoreconf -fi` + `./configure` + `make`); Apache-2.0.
- Source tarball doesn't include the vendored gtest submodule, but the install path doesn't require it (tests only).

## Test plan

Round-trip encode/decode of a small dictionary+target pair, verifying byte-identity after decode. Exercises both the encoder and decoder.

### DIFF
--- a/projects/github.com/google/open-vcdiff/package.yml
+++ b/projects/github.com/google/open-vcdiff/package.yml
@@ -1,0 +1,60 @@
+# open-vcdiff — VCDIFF (RFC 3284) encoder/decoder by Google.
+#
+# Generic-differencing format used by Chrome's SDCH (Shared Dictionary
+# Compression over HTTP), Google's Update protocol, and various delta
+# compression schemes. Apache-2.0.
+#
+# Upstream marked the repo archived on 2023-08-26 — no active
+# development but it's RFC-compliant and remains the canonical
+# reference C++ implementation. Worth packaging precisely because
+# nothing replaces it.
+
+distributable:
+  url: https://github.com/google/open-vcdiff/archive/refs/tags/{{version.tag}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: google/open-vcdiff
+  # Tag scheme changed historically: older releases used
+  # `open-vcdiff-X.Y` (dashes), recent ones use `openvcdiff-X.Y.Z`
+  # (no dash). Strip both.
+  strip:
+    - /^openvcdiff-/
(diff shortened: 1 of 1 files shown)


tannevaled:
**Parking this PR — source build needs vendored gflags submodule + custom m4 macros.**

After 3 iterations, the structural issue is that open-vcdiff's `configure.ac` uses Google-custom autoconf macros (`AC_CXX_STL_NAMESPACE`, `AC_DEFINE_GOOGLE_NAMESPACE`) that live in the `gflags/m4/` git submodule, which is **not in the GitHub archive tarball**.

To make this work properly, the recipe would need to:
1. `git clone https://github.com/gflags/gflags.git` (at a specific submodule sha — not preserved in the tarball)
2. Optionally clone `gtest` too (only for unit tests, not the binary)
3. Then `autoreconf -fi` with `-I gflags/m4`

Or patch `configure.ac` to drop the legacy Google macros (`AC_CXX_STL_NAMESPACE` is obsolete — all C++ now uses `std::` namespace).

Since this is an **archived upstream** (last commit 2023-08-26) and the recipe is a 30-line autotools build, the maintenance ROI is low. Leaving the PR open for review — happy to invest further iterations if maintainers consider it worth it, otherwise closing in favor of a leaner fix later.

Symptoms documented in the recipe / commit log:
- Iter 1 (\`d5394f6c\`): \`not-found: version\` (default github releases endpoint hit empty — 0 releases, 5 tags)
- Iter 2 (\`e0d4f1af\`/\`2051bb63\`): switched to \`/tags\` + ignored 2-part \`open-vcdiff-0.8\` → version resolution worked
- Iter 3 (\`f5f7971a\`): \`mkdir -p gflags/m4 gtest\` to satisfy autoreconf's directory probe → uncovered the next layer: m4 macros themselves are missing, not just the dir

Issue #7853, opened 2024-11-07
lima-vm: vz error with the pkgx version, not with the brew one


tannevaled:
# Nested virtualization under M3+ masOS 15+

## Template file : nested.yaml

```yaml
minimumLimaVersion: "1.0.0"
images:
# Try to use release-yyyyMMdd image if available. Note that release-yyyyMMdd will be removed after several months.
- location: "https://cloud-images.ubuntu.com/releases/24.04/release-20241004/ubuntu-24.04-server-cloudimg-amd64.img"
  arch: "x86_64"
  digest: "sha256:fad101d50b06b26590cf30542349f9e9d3041ad7929e3bc3531c81ec27f2c788"
- location: "https://cloud-images.ubuntu.com/releases/24.04/release-20241004/ubuntu-24.04-server-cloudimg-arm64.img"
  arch: "aarch64"
  digest: "sha256:e380b683b0c497d2a87af8a5dbe94c42eb54548fa976167f307ed8cf3944ec57"
# Fallback to the latest release image.
# Hint: run `limactl prune` to invalidate the cache
- location: "https://cloud-images.ubuntu.com/releases/24.04/release/ubuntu-24.04-server-cloudimg-amd64.img"
  arch: "x86_64"
- location: "https://cloud-images.ubuntu.com/releases/24.04/release/ubuntu-24.04-server-cloudimg-arm64.img"
  arch: "aarch64"

mounts:
- location: "~"
- location: "/tmp/lima"
  writable: true

vmType: vz
nestedVirtualization: true
```
## VM creation

```bash
phymath@mba-10838921 Documents % pkgx limactl create --name nested --tty=false ./nested.yaml
INFO[0000] Terminal is not available, proceeding without opening an editor 
WARN[0000] vmType vz: ignoring [User]                   
INFO[0000] Attempting to download the image              arch=aarch64 digest="sha256:e380b683b0c497d2a87af8a5dbe94c42eb54548fa976167f307ed8cf3944ec57" location="https://cloud-images.ubuntu.com/releases/24.04/release-20241004/ubuntu-24.04-server-cloudimg-arm64.img"
INFO[0000] Using cache "/Users/phymath/Library/Caches/lima/download/by-url-sha256/1f3456e7d7c2bc0a8f8993d1308a3a4c124d703f56bf4cf6dfca50eb5f11f1c3/data" 
[...]

PR #13197, opened 2026-06-06
ci(setup): add `skip-codesign-import` lever for adhoc-only recipes

tannevaled:
## Summary

Defense-in-depth follow-up to pkgxdev/brewkit#353. Adds a workflow-level lever to skip importing the Apple Developer ID into the keychain for recipes that need adhoc signing.

## Why

Some recipes ship binaries with entitlements that require either adhoc signing or a Developer ID with a matching provisioning profile — notably `com.apple.security.virtualization` and `com.apple.security.hypervisor`. brewkit#353 handles the binary side correctly: it sees these entitlements in the Mach-O and forces adhoc signing in `fix-machos.rb`, regardless of what signing identity the workflow imported.

This PR is the workflow-level complement that @jhheider raised in #7853: when we know a recipe is in this category, why import a Developer ID at all? It's wasteful and leaves the cert in the keychain longer than necessary.

## What changes

Purely additive lever. Default stays `false` → no behavior change for any current call.

```
pkg.yml (workflow_call input: skip-codesign-import)
  -> pkg-platform.yml (workflow_call input: skip-codesign-import)
    -> .github/actions/setup (composite input, gates the import step)
```

The `apple-actions/import-codesign-certs@v6` step (and the prophylactic `Delete keychain` step before it) are gated on `inputs.skip-codesign-import != 'true'`.

## What is **not** in this PR (deliberately)

Picking which recipes opt in. Two reasons:

1. **brewkit#353 already provides the correctness guarantee.** This PR is hardening, not a fix — recipes don't break if their flag isn't set, they just import a cert they don't end up using.
2. **The discussion of "which recipes / what mechanism" deserves its own thread.** Options range from a per-recipe pantry metadata field to a static allow-list in the workflow. Worth deciding once, separately.

`lima-vm.io` is the obvious first candidate (#7853) and a follow-up would add it.

## Test plan

- [x] YAML syntax validated on all three files
- [x] Default path unchanged (lever defaults `false`)
- [ ] CI passes
- [ ] Smoke test: kick a recipe build with `skip-codesign-import: true` and confirm the import step is skipped

Refs: pkgxdev/pantry#7853, pkgxdev/brewkit#353

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -11,6 +11,16 @@ inputs:
   APPLE_IDENTITY:
     description: Apple identity to use for signing
     required: false
+  skip-codesign-import:
+    description: >
+      When 'true', skip importing the Apple Developer ID into the keychain.
+      Useful for recipes whose binaries require adhoc signing (eg. those
+      carrying com.apple.security.virtualization / .hypervisor entitlements,
+      which need either adhoc or a matching provisioning profile to be
+      accepted by macOS at runtime). Defense-in-depth on top of brewkit's
+      fix-machos.rb policy. See pkgxdev/pantry#7853.
+    required: false
+    default: 'false'
 
 runs:
   using: composite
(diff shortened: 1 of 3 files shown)


tannevaled:
Heads-up on the red ci²: the 7 failures are all the fork-secrets limitation, not the change itself. Annotation reads:

> Secret GPG_KEY_ID is required, but not provided while calling.

(+ `GPG_PRIVATE_KEY`, `AWS_S3_BUCKET`, `AWS_SECRET_ACCESS_KEY`, `AWS_CF_DISTRIBUTION_ID`)

Pantry repo secrets aren't passed to `workflow_call` jobs spawned from a fork PR — same failure mode as the recent dependabot PRs on `checkout-6` and `configure-aws-credentials-6`. The jobs were refused at workflow startup (~2s each), they didn't actually run.

The 11 SUCCESS checks cover the YAML validators and the lanes that don't depend on those secrets. The change itself is purely additive: new input defaults to `false` everywhere, no existing call site changes behavior.

If a real ci² pass is needed, the branch would need to live inside pkgxdev/pantry — happy to redo from a maintainer-pushed branch if that helps. Otherwise this is reviewable by diff (~30 lines, 3 YAML files).

jhheider:
i think this adds complexity, for the relatively few packages that might use it.

PR #13263, opened 2026-06-11
new(acpica.org): iasl ACPI compiler

tannevaled:
Adds `acpica.org`, packaging Intel ACPICA's **`iasl`** — the ASL+ optimizing compiler/disassembler.

### Why
EDK II (#13239) shells out to `iasl` to compile `.asl` ACPI tables into `.aml` during firmware builds (RamDiskDxe's NFIT table, etc.). `iasl` isn't currently in the pantry, so the edk2 build fails with `iasl: command not found` (exit 127). This packages it as a standalone build-time dependency.

> Prerequisite for #13239 — that PR build-depends on `acpica.org`, so this should land first.

### Notes
- Tags switched from `R20YY_MM_DD` to a clean `YYYYMMDD` scheme around mid-2025; we match only the latter.
- **darwin/arm64** links with `-Wl,-no_fixup_chains`: ACPICA keeps unaligned function pointers in `__DATA,__const` tables that the mach-o chained-fixups format rejects (`ld: pointer not aligned in '_AcpiExDumpNode'`; lld emits the same advice).

### Test plan
Built `iasl` locally on darwin/arm64 (pantry flex/bison/clang) from the resolved tag `20260408`; `iasl -v` reports the version and it compiles a trivial SSDT to `.aml`.

### DIFF
--- a/projects/acpica.org/package.yml
+++ b/projects/acpica.org/package.yml
@@ -1,0 +1,57 @@
+# ACPICA — Intel's ACPI Component Architecture.
+#
+# We package it for one tool in particular: `iasl`, the ASL+ optimizing
+# compiler/disassembler. EDK II (tianocore.org/edk2) shells out to it to
+# turn `.asl` ACPI tables into `.aml` during firmware builds, so it is a
+# build-time dependency there rather than a user-facing package.
+#
+# Upstream tags switched from `R20YY_MM_DD` to a clean `YYYYMMDD` scheme
+# around mid-2025; we only match the latter.
+
+distributable:
+  url: git+https://github.com/acpica/acpica
+  ref: '{{version.raw}}'
+
+versions:
+  github: acpica/acpica/tags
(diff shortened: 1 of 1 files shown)


tannevaled:
Hey @jhheider 👋 — sorry to ping you directly, no rush at all!

When you have a spare moment, would you mind taking a look at this one? It's a small standalone recipe for `iasl` (ACPICA's ASL compiler). It's the prerequisite for #13239 (the EDK II / OVMF firmware blobs) — that build shells out to `iasl` to compile the ACPI tables, so it can't go green until this lands in `main`.

CI is green on all four platforms here. The only platform-specific bit is the `-Wl,-no_fixup_chains` on darwin/arm64 (ACPICA keeps unaligned pointers in `__DATA,__const` that the mach-o chained-fixups format rejects).

Thanks a lot for everything you do on the pantry! 🙏

PR #13294, opened 2026-06-13
Add clj & clojure from clojure.org

phrohdoh:
## Summary

- What changed: Added `clj` & `clojure` from clojure.org pre-builts for homebrew
- Why: So that people can use `clj` & `clojure` via `pkgx`, `pkgm`, etc.

This PR has 2 commits:
- build from source
- vendor

I am not certain if either will work, or if vendoring is just easier for the team. I included both so that it would hopefully be easier to get this in one way or another.

Example release: https://github.com/clojure/brew-install/releases/tag/1.12.5.1654

## AI Intent

- Prompt or task statement: None.
- Scope of AI-assisted changes: None.

## Risk Assessment

- Risk level: **low** / medium / high
- Primary risks: It doesn't work.
- Compatibility impact: ?

## Rollback Plan

- Revert path: Revert commit.
- Forward-fix path (if revert is not possible): Make it work.

## Validation

- Local checks run: None, I don't use `dev` or `bk`.
- CI checks expected: Syntactic validity?

## Internal/Public Boundary Check

- [x] No internal-only data, private URLs, secrets, or private runbooks were
      added.

### DIFF
--- a/projects/clojure.org/package.yml
+++ b/projects/clojure.org/package.yml
@@ -1,0 +1,53 @@
+distributable:
+  url: https://github.com/clojure/clojure/archive/refs/tags/clojure-{{version}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: clojure/clojure/releases/tags
+  strip: /^clojure-/
+
+dependencies:
+  openjdk.org: ">=11"
+  maven.apache.org: "^3.8"
+
+build:
+  dependencies:
+    openjdk.org: ">=11"
+    maven.apache.org: "^3.8"
+
+  script: |
+    # Ensure we're using the correct Java version
+    export JAVA_HOME=$(pkgx which java | xargs dirname | xargs dirname)
+
+    # Build the Clojure JAR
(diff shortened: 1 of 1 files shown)


phrohdoh:
To the best of my knowledge, using `clj` requires `rlwrap` but since pkgx does not currently have a `rlwrap` package the end-user must install it themselves (via `brew` or some other means).

jhheider:
@phrohdoh did your source build work? and did you evaluate `rlwrap` for packaging viability?

phrohdoh:
> @phrohdoh did your source build work? and did you evaluate `rlwrap` for packaging viability?

I do not know whether the source build worked, as I am unable to get my local pkgx working with my local pantry:

> I am not certain if either will work [...]

I have not evaluated `rlwrap` for packaging viability, as it is not absolutely required. I believe that`clj` is an `rlwrap`-based wrapper for `clojure`. So end-users can either use `clojure` without `rlwrap` installed or they can install `rlwrap` then use either.

jhheider:
looks like we might have `rlwrap` available, now as `github.com/hanslub42/rlwrap`.

PR #13308, opened 2026-06-13
new(vector.dev): observability data pipeline (vendored, linux)

tannevaled:
Adds [Vector](https://vector.dev) — high-performance observability data pipeline (logs/metrics/traces).

## Why vendored
Two reasons: (1) **upstream publishes no macOS build**, and (2) a from-source build is a **very large Rust compile** — Vector pins **Rust 1.92** and needs `protoc` + `cmake` + perl, with a deep crate graph (30–60 min). Vendoring upstream's official **Linux** binaries (x86-64 + aarch64) gives a working `vector` now.

Verified on linux/aarch64 (Debian): `vector --version` → 0.56.0. Linux-only (no upstream macOS binary).

A from-source `vector.dev` would:
1. `build.dependencies: { rust-lang.org: '>=1.92', rust-lang.org/cargo, protobuf.dev (protoc), cmake.org, perl.org }`.
2. `cargo build --release --bin vector` (default feature set is large; a curated `--no-default-features --features …` set could trim build time/size).
3. This would **also unlock macOS** (which has no upstream binary) — the main reason to eventually go from-source here.

Until then, vendored covers Linux, where official binaries exist.

### DIFF
--- a/projects/vector.dev/package.yml
+++ b/projects/vector.dev/package.yml
@@ -41,6 +41,11 @@ build:
     PATH: $HOME/.cargo/bin:$PATH
     # protoc lives in the protobuf.dev prefix; prost-build/tonic-build honor $PROTOC.
     PROTOC: ${{deps.protobuf.dev.prefix}}/bin/protoc
+    linux:
+      # bundled librdkafka's version script lists rd_ut_coverage_check (a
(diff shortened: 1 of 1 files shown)


jhheider:
we should definitely do the from-source build. there are long builds already, as long as it doesn't exceed 6 hours.

PR #13316, opened 2026-06-14
new(clickhouse.com): column-oriented OLAP database (vendored)

tannevaled:
Adds [ClickHouse](https://clickhouse.com) — fast column-oriented OLAP database.

## Why vendored
A from-source ClickHouse build is a **multi-hour C++ compile** (LLVM/clang toolchain, huge translation units, very high RAM/disk) — impractical on standard CI runners. Vendoring upstream's official static Linux builds + macOS binaries gives a working `clickhouse` multi-tool binary now, on all 4 platforms.

Verified on linux/aarch64 (Debian): extracts the 677 MB static binary, `clickhouse --version` → 26.4.4.38.

ClickHouse builds with CMake + Ninja against a specific clang. A from-source recipe would need:
1. `build.dependencies: { cmake.org, ninja-build.org, llvm.org (pinned major), … }`.
2. A configure step disabling the bundled-everything default to fit runner resources, and almost certainly a **much larger CI timeout** (hours) — the main blocker.

Given the build cost and that upstream ships reproducible official binaries, vendored is the pragmatic choice here.

### DIFF
--- a/projects/clickhouse.com/package.yml
+++ b/projects/clickhouse.com/package.yml
@@ -1,0 +1,54 @@
+# ClickHouse — fast, open-source, column-oriented OLAP database. Vendors
+# upstream's official static Linux builds and macOS binaries (a
+# from-source build is a multi-hour C++ compile). One `clickhouse`
+# multi-tool binary: `clickhouse local|client|server|keeper`.
+
+distributable: ~
+
+versions:
+  github: ClickHouse/ClickHouse/tags
+  match: /^v\d+\.\d+\.\d+\.\d+-stable$/
+  strip:
+    - /^v/
+    - /-stable$/
+
+platforms:
+  - linux/x86-64
+  - linux/aarch64
+  - darwin/x86-64
+  - darwin/aarch64
+
+warnings:
+  - vendored
+
+build:
+  dependencies:
+    curl.se: "*"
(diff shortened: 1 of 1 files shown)


jhheider:
how many multi hours? 6 is the CI limit (which node butts up against).

tannevaled:
> how many multi hours? 6 is the CI limit

A full from-source ClickHouse build is right at / over that limit, which is why I kept this one vendored while converting the others. Concretely:

- ClickHouse is ~10M LOC of C++ and bundles its dependencies; upstream's own CI uses sccache + large dedicated builders and still measures release builds in the **2-4h** range on well-provisioned x86. On a stock CI runner without a warm compiler cache, and across the arm64 leg of the matrix, that realistically pushes **past 6h**.
- It is also memory-hungry at link time (upstream recommends ~16-30 GB); constrained runners tend to OOM during linking before they even time out.

So this is the one case where I think the vendored binary is defensible under your 6h rule -- unlike `vector.dev` (#13308) and `tikv.org` (#13319), which I've just converted to from-source (tikv via the `crates.io/skim` toolchain-pin pattern you pointed me at).

Happy to push a from-source attempt anyway if you want the empirical wall-clock from CI -- just say the word and I'll convert it and we'll see where it lands.

PR #13317, opened 2026-06-14
new(tcl-lang.org/tk/v8): Tk 8.6 GUI toolkit

tannevaled:
Split from #13081. Tk 8.6 GUI toolkit; companion to `tcl-lang.org/v8`.

**Depends on `tcl-lang.org/v8` (#13081) — merge that first** so Tcl's bottle is published (otherwise the build 404s resolving it; that co-dependency is why the combined PR's Tk checks were red).

### DIFF
--- a/projects/tcl-lang.org/tk/v8/package.yml
+++ b/projects/tcl-lang.org/tk/v8/package.yml
@@ -1,0 +1,63 @@
+# Tk 8.6 — legacy GUI toolkit.
+#
+# Companion to tcl-lang.org/v8. Tk 9.0 has a broken `minizip` target
+# in its Makefile that fails build; 8.6.x stays buildable. Required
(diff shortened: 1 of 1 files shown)


tannevaled:
Closing as redundant. The dedicated `tcl-lang.org/tk` recipe (#13057) is already version-generic — its `versions:` matcher includes the 8.6 line and the recipe explicitly handles both series (the `tcl.h` path differs 8.6 vs 9.0; `wish{{version.major}}.{{version.minor}}` yields `wish8.6`/`wish9.0`). So `tcl-lang.org/tk^8` resolves to Tk 8.6 and builds from the same recipe — no separate `/v8` package needed, matching the same version-path approach used for `tcl-lang.org` itself (per @jhheider's review on #13081).

PR #13319, opened 2026-06-14
new(tikv.org): distributed transactional KV store (vendored)

tannevaled:
Adds [TiKV](https://tikv.org) — distributed transactional key-value store (CNCF graduated), the storage layer behind TiDB.

## Why vendored
TiKV builds **from a pinned Rust nightly** (`rust-toolchain.toml` → `channel = "nightly-2023-12-28"`) plus a full **RocksDB (C++) + gRPC** compile — a multi-hour, high-RAM build. pantry's `rust-lang.org` only provides stable Rust, so a from-source recipe can't satisfy the pin today. Vendoring PingCAP's official `tikv-server` from the TiUP mirror gives users a working binary now, on all 4 platforms.

Verified on linux/aarch64 (Debian): `tikv-server --version` → 8.5.6.

1. A **date-pinned nightly toolchain** must exist in-pantry — proposed in #13320 (`rust-lang.org/nightly@YYYYMMDD`, vendored from static.rust-lang.org). 
2. Then a from-source `tikv.org` would `build.dependencies: { rust-lang.org/nightly: =20231228, cmake.org: '*' (RocksDB) }` and run the TiKV `Makefile`/`cargo build --release`.
3. Expect a long build (RocksDB) — may need a bumped CI timeout.

Until then, this vendored recipe is the practical path.

### DIFF
--- a/projects/tikv.org/package.yml
+++ b/projects/tikv.org/package.yml
@@ -50,6 +50,10 @@ build:
     # the (1.1-only) SSL_get_peer_certificate symbol.
     OPENSSL_DIR: ${{deps.openssl.org.prefix}}
     OPENSSL_NO_VENDOR: "1"
+    # Build RocksDB/Titan portable (no -msse4.2): the default FORCE_SSE42 is
+    # x86-only and aborts the arm64 cmake, and a distributed binary shouldn't
+    # bake in host CPU features anyway.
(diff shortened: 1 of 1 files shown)


jhheider:
see `crates.io/skim`

PR #13322, opened 2026-06-14
new(x.org/xext): libXext 1.3.7

tannevaled:
Adds **libXext**, the X11 common extensions library — the one remaining gap in pantry's otherwise-complete X.org client stack.

From-source via the standard X.org autotools flow, mirroring the sibling x.org libraries (libXfixes / libXrender / libXrandr). Depends only on libX11 (`x.org/x11`) and xorgproto (`x.org/protocol`).

(Surfaced while packaging PyQt5 #13064: the manylinux Qt wheel dlopens libXext among the rest of the x.org stack.)

### DIFF
--- a/projects/x.org/xext/package.yml
+++ b/projects/x.org/xext/package.yml
@@ -1,0 +1,35 @@
+distributable:
+  url: https://www.x.org/archive/individual/lib/libXext-{{version}}.tar.xz
+  strip-components: 1
+
+versions:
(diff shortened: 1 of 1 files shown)


jhheider:
duplicative of `x.org/exts`, i think.

tannevaled:
You're right — confirmed `x.org/exts` already builds `libXext-{{version}}.tar.gz` with the same deps. Closing as duplicate. Thanks!

PR #13324, opened 2026-06-14
new(atuin.sh): atuin 18.16.1

tannevaled:
Adds **atuin** — a syncing, searchable Ctrl-R shell-history replacement.

Built **from source** via `cargo install --locked --path crates/atuin` (the workspace's binary crate), installing `bin/atuin`. atuin uses edition 2024 and pins a recent `rust-version`, so it needs a current Rust toolchain (`rust-lang.org: '*'` resolves to the newest pantry rust).


jhheider:
upstream requested removal in #5466

PR #13429, opened 2026-06-19
new(gnu.org/plotutils): plotutils 2.6

tannevaled:
Adds **GNU plotutils** — the `graph`, `plot`, `spline`, `ode`, `pic2plot`, `tek2plot`, `plotfont`, `hersheydemo`, `double` utilities plus libplot/libplotter. autotools. Built `--without-x` (drops the whole X11 stack; PS/PNG/SVG/etc. drivers + all 9 CLIs remain). 2-component version -> `{{version.raw}}` in the tarball URL (2.6, not padded 2.6.0). libpng+zlib are real runtime deps (confirmed via `ldd`); C++ libplotter kept (`--enable-libplotter`) so gcc14/libstdcxx on linux. Two source patches needed for a modern toolchain: C++17 rejects `register` in `pic2plot/gram.cc`, and libpng>=1.5 hides `jmpbuf` (use `png_jmpbuf()`). Verified end-to-end on linux/arm64: `graph --version` -> 2.6, PS + PNG output valid.

### DIFF
--- a/projects/gnu.org/plotutils/package.yml
+++ b/projects/gnu.org/plotutils/package.yml
@@ -1,20 +1,29 @@
 distributable:
   url: https://ftp.gnu.org/gnu/plotutils/plotutils-{{version.raw}}.tar.gz
   strip-components: 1
+
 versions:
   url: https://ftp.gnu.org/gnu/plotutils/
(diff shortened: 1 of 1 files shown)


tannevaled:
Closing — gnu.org/plotutils already exists in the pantry; opened in error. No changes intended to the maintained recipe.

PR #13475, opened 2026-06-22
new(github.com/qhull/qhull): qhull 8.0.2

tannevaled:
Qhull — convex hulls / Delaunay / Voronoi. C/CMake. Build dir is `builddir` (source ships a `build/` that a `-B build` clobbers); `CMAKE_POLICY_VERSION_MINIMUM=3.5` for cmake≥4. Pure C (CLI tools link only libm/libc → no libstdcxx). Verified linux/arm64: rbox|qhull on a cube → 8 vertices.


jhheider:
duplicative of `qhull.org`.

PR #13506, opened 2026-06-23
new(github.com/lcn2/calc): calc 3.0.1.0

tannevaled:
Adds **calc** (Landon Curt Noll) — arbitrary-precision C-style interactive calculator. C. The Makefile's install path vars are hardcoded absolutes (not derived from `PREFIX`), so `BINDIR`/`LIBDIR`/`INCDIR`/`CALC_SHAREDIR`/`MANDIR` are each set to `{{prefix}}/…`; readline disabled to keep deps empty. The baked `CALC_SHAREDIR` resolves post-install (PREFIX is the final dir); a `runtime.env CALCPATH` mirrors calc's defaults as belt-and-suspenders. Verified linux/arm64 from a clean dir: `calc -v` → 3.0.1.0, `echo '2+2' | calc -p` → 4, `sqrt(2)` loads a .cal lib.

### DIFF
--- a/projects/github.com/lcn2/calc/package.yml
+++ b/projects/github.com/lcn2/calc/package.yml
@@ -1,0 +1,41 @@
+distributable:
+  url: https://github.com/lcn2/calc/archive/refs/tags/{{version.tag}}.tar.gz
+  strip-components: 1
+
+versions:
+  github: lcn2/calc
+
+provides:
+  - bin/calc
+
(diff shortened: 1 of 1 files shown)


tannevaled:
Withdrawing for now: calc's Makefile fails to build cleanly under the CI toolchain (the calc.usage target can't exec the freshly-built calc against libcalc.so mid-build, Error 127), and GNU bc already covers the basic-calculator need in the pantry. Will revisit if there's interest with a proper fix.

